Issue #76, opened 2022-02-26
TEMPLATE Default Event Template


mauteri:
On new event in editor, setup a default template with common event blocks in displayed ready for configuration.

MervinHernandez:
I created a custom FSE template for `events` which is alive on both wpNashville and in gatherpress site.
Can you leverage this by baking it into the plugin, and if the person installing gatherpresess uses `twenty-twenty-two` then it will be auto poulated.



![Image](https://user-images.githubusercontent.com/8400371/188206342-b10109a0-6e01-439b-b05c-c0ccc508cac3.png)

mauteri:
@MervinHernandez will need to keep it to just GatherPress blocks as we can't assume any other blocks will be registered.

MervinHernandez:
That sounds fine - my thought is just to bake in a template for default rendering of a single-event.

MervinHernandez:
ACTIVE CONVERSATION = this should be aligned between admin side (writing experience and view) and front-end rendering. Paul has thoughts on this.

mauteri:
MVP of this is completed. We can revisit block layout in the future, but wanted to have our basic blocks populating in a new event.

pbrocks:
@MervinHernandez @mauteri Let's close

PR #126, opened 2022-08-20
starter template and move category to top

pbrocks:
Move Categories to top
image

Add starter template
image

### DIFF
--- a/core/classes/class-setup.php
+++ b/core/classes/class-setup.php
@@ -87,15 +87,12 @@ class Setup {
 	 * @return array
 	 */
 	public function block_category( $categories ) {
-		return array_merge(
-			$categories,
-			array(
-				array(
-					'slug'  => 'gatherpress',
-					'title' => __( 'GatherPress', 'gatherpress' ),
-				),
-			)
+		$gatherpress_blocks = array(
+			'slug'  => 'gatherpress',
+			'title' => __( 'GatherPress', 'gatherpress' ),
 		);
+		array_unshift( $categories, $gatherpress_blocks );
+		return $categories;
 	}
 
 	/**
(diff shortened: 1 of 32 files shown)


mauteri:
@pbrocks this is similar to https://github.com/GatherPress/gatherpress/issues/76 but the idea was not to have a starter template block, but to set `template` in the event post type when it is registered: https://developer.wordpress.org/block-editor/reference-guides/block-api/block-templates/ This saves a click from setting the default block layout, by presenting an event with the most common blocks to get started. We don't have all the blocks we need yet, so this hasn't been done, but if you want to get started on this that'd be great.

core/classes/class-setup.php
```diff
@@ -147,6 +144,11 @@ public function register_post_types() {
 					'revisions',
 				),
 				'menu_icon'     => 'dashicons-nametag',
+				'template'  => array(
+					array(
+						'gatherpress/event-starter',
```
mauteri, line 149: Okay, going back to my comment before I see you are adding it here. Why would we make this a block too in `event-starter`? I'd rather us not clutter up blocks UI and just call the blocks in template.

pbrocks:
> this is similar to #76 but the idea was not to have a starter template block, but to set `template` in the event post type when it is registered:

That is what I did. I registered the template in the post type, but they can be considered placeholders.

mauteri:
> > this is similar to #76 but the idea was not to have a starter template block, but to set `template` in the event post type when it is registered:
> 
> That is what I did. I registered the template in the post type, but they can be considered placeholders.

Not sure we need this listed in with blocks, I think it just adds clutter. I would just keep it simple and set the choice blocks to the template. I'm thinking within settings at some point we can provide a few layout options and control it from there.

PR #129, opened 2022-08-27
Bug fix. Default event in event list block was object vs array so map…

mauteri:
… function through an error.


pbrocks (commented):
OMFG!!

Issue #155, opened 2022-09-11
ADD Venue to Events List Block


mauteri:
Events List Block needs to display venue name that is linked to Venue page.

MervinHernandez:
PLUS - debug why this content (venue) does not render in the front end 

![Image](https://user-images.githubusercontent.com/8400371/192869406-090570ff-4b95-46bc-a610-d28f9992d0cb.png)

jmarx:
@MervinHernandez You can assign this to me.

jmarx:
@MervinHernandez I believe the issue in your screenshot(The venue not displaying on the event template).
Is solved by this code which was merged last week:
https://github.com/GatherPress/gatherpress/pull/161/files#diff-2289f56eecb10faaa9f12edf27aa910fe9e69ab4da1bee92de6b013e74035402R147-R153
From this PR:
https://github.com/GatherPress/gatherpress/pull/161/files#diff-2289f56eecb10faaa9f12edf27aa910fe9e69ab4da1bee92de6b013e74035402R147-R153
It was referencing a very old issue:
https://github.com/GatherPress/gatherpress/issues/76

Upon creation of a new event, you should see a set of blocks available by default on the editor screen.
![Screen Shot 2022-10-10 at 2 07 30 PM](https://user-images.githubusercontent.com/223919/194928417-8e0448ba-d4a7-4e96-a471-5970153e3d01.png)

Once you add details and save/publish the event, you should see all these details(Including the venue) on the front end of the single event template.

![Screen Shot 2022-10-10 at 2 13 02 PM](https://user-images.githubusercontent.com/223919/194928856-8a5451a3-6dc7-43ec-9211-898cf8b125a5.png)

Can you do me a favor and make sure that I am understanding the issue correctly? And please try and create an event and make sure you see the venue as we would expect? If not, let's schedule a zoom session so I can see what you are seeing.

Then @mauteri this Pr should be ready to be reviewed assuming @MervinHernandez 's concerns are satisfied.
https://github.com/GatherPress/gatherpress/pull/166

MervinHernandez:
Hi @jmarx . After updating WP Nashville to the latest 0.1, this issu persists.
Populated data for event venue, is not displayed in the front-end to visitors, despite being added via block in the FSE editor.


![Image](https://user-images.githubusercontent.com/8400371/195996683-0e34b9cc-01e1-4b1f-8de7-6e05c8ba1658.png)

pbrocks:
[...]

PR #166, opened 2022-10-06
Feature/155 display venue name

jmarx:
#155 

Display the venue name in the event list block.


core/classes/class-rest-api.php
```diff
@@ -316,6 +316,7 @@ function( $slug ) {
 		if ( $query->have_posts() ) {
 			foreach ( $query->posts as $post_id ) {
 				$event   = new Event( $post_id );
+				$venue_information = $event->get_venue_information();
```
jmarx, line 319: @jmarx Maybe set a default in case this function fails for whatever reason?

pbrocks:
@jmarx Can you share what you are seeing with some screenshots?

jmarx:
@pbrocks I have to wait until later(when I turn off my work VPN) But I can tell you that I just get the post title, but don't see any events when I save a post with the events list block. Like I said in slack, It's probably some silly issue like permalinks or something. I'll follow up later.

core/classes/class-rest-api.php
jmarx: As a follow-up, I realize that this is not needed as that `get_venue_information` method returns and empty venue by default.

core/classes/class-rest-api.php
```diff
@@ -327,6 +328,7 @@ function( $slug ) {
 					'featured_image_thumbnail' => get_the_post_thumbnail( $post_id, 'thumbnail' ),
 					'attendees'                => ( $event->attendee ) ? $event->attendee->attendees() : array(),
 					'current_user'             => ( $event->attendee && $event->attendee->get( get_current_user_id() ) ) ? $event->attendee->get( get_current_user_id() ) : '',
+					'venue'                    => $venue_information
```
mauteri, line 331: `'venue'                    => $event->get_venue_information(),`
only using once, so no need to set variable; add trailing comma per standards.

assets/src/components/EventItem.js
```diff
@@ -19,6 +19,8 @@ const EventItem = ( props ) => {
 
 	const eventClass = `gp-events-list`;
```
mauteri, line 21: remove linebreak

assets/src/components/EventItem.js
```diff
@@ -19,6 +19,8 @@ const EventItem = ( props ) => {
...
 
+	const venue = event.venue;
```
mauteri, line 22: what do you think of something like...
``const venue = <a href="`${ venue.permalink }`">`${ venue.name }`</a>``

assets/src/components/EventItem.js
```diff
@@ -35,6 +37,12 @@ const EventItem = ( props ) => {
 					>
 						<strong>{ event.datetime_start }</strong>
 					</div>
+					{ venue.name && (
+						<div>
+							Venue: { HtmlReactParser( '<a href=' + venue.permalink + '>' + venue.name + '</a>' ) }
```
mauteri, line 42: `Venue: { HtmlReactParser( venue ) }`

core/classes/class-rest-api.php
mauteri: I would write a unit test for `get_venue_information` to ensure it doesn't fail in a way we don't expect.

mauteri (changes requested):

mauteri (approved):

Issue #44, opened 2022-01-07
ADD Action Flush Permalinks on activation


MervinHernandez:
Upon activating plugin - flush permalinks.

MervinHernandez:
@pbrocks is this still on your radar my friend?

MervinHernandez:
@pbrocks is this just `add_action('init','rewrite_flush()')` ... or something else is needed?

pbrocks:
Mike did this

mauteri:
I don't think I did... I was leaving this one for you.

pbrocks:
@mauteri @MervinHernandez https://github.com/GatherPress/gatherpress/pull/136

pbrocks:
Looks like [this](https://github.com/GatherPress/gatherpress/pull/167) takes care of it. I suggest closing this Issue, @MervinHernandez

pbrocks:
@MervinHernandez @mauteri Let's close

PR #168, opened 2022-10-08
Flush is needed on Fresh Install

pbrocks:
This PR fixes the flush that is needed on a fresh install. Screencast illustrates activation on main branch rendering a 404 for a new event, then switching to this branch, deactivating and reactivating, some event renders properly.

https://github.com/GatherPress/gatherpress/issues/44

flush-needed


jmarx (approved):

gatherpress.php
```diff
@@ -30,3 +30,40 @@
...
+/**
+ * Activate GatherPress plugin.
+ *
+ * @return void
+ */
+function deactivate_gatherpress_plugin() {
+	flush_rewrite_rules();
+}
+
+add_action( 'init', 'maybe_flush_gatherpress_rewrite_rules', 20 );
```
mauteri, line 59: Wouldn't use just random priorities. If you want it to run right after `10` make it `11`. If you want it to run last make it `PHP_INT_MAX`.

pbrocks: In the end I was able to remove it altogether

mauteri (approved):

PR #173, opened 2022-10-23
Move dot files from assets/ to root.

mauteri:


pbrocks (approved):
Let's merge to main so we can move forward from the root fixing the time issues, etc

pbrocks:
Glad to see that we can get to a Green Check for CircleCI
image

mauteri:
@pbrocks Yeah, I think CircleCI only works for my account... I'm thinking of moving things to Github Actions.

PR #180, opened 2022-10-29
No Ticket: Use dashicon for venue span

jmarx:
@mauteri 

Screen Shot 2022-10-29 at 2 48 13 PM


src/components/EventItem.js
```diff
@@ -38,7 +38,7 @@ const EventItem = ( props ) => {
 					</div>
 					{ venue && (
 						<div>
-							Venue: { venue }
+							<span class="dashicons dashicons-location">&nbsp;</span> { venue }
```
mauteri, line 41: Take a look at styling done for the dashicons on the editor so you don't need to use `&nbsp;`. I think they are flex layouts if I recall.

mauteri: Also, let's put this after the title rather than before. Thx!

mauteri (changes requested):

src/components/EventItem.js
```diff
@@ -50,7 +50,12 @@ const EventItem = (props) => {
 					>
 						<strong>{event.datetime_start}</strong>
 					</div>
-					{venue && <div>Venue: {venue}</div>}
+					{venue && (
+						<div>
+							{venue}
+							<span className="dashicons dashicons-location"></span>
+						</div>
+					)}
```
mauteri, line 58: Let's move this after the title (line 64) so it goes
DATE TIME
TITLE
VENUE

mauteri (changes requested):

src/components/EventItem.js
jmarx: @mauteri 
Done

mauteri (approved):

PR #200, opened 2022-11-28
Feature/199

jmarx:
Force localized gatherpress object script into the head.


includes/core/classes/class-assets.php
```diff
@@ -57,6 +57,19 @@ protected function setup_hooks() {
 		add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_scripts' ) );
 		add_action( 'admin_enqueue_scripts', array( $this, 'admin_enqueue_scripts' ), 10, 1 );
 		add_action( 'enqueue_block_editor_assets', array( $this, 'block_enqueue_scripts' ) );
+		add_action( 'enqueue_block_assets', array( $this, 'add_global_object' ) );
```
mauteri, line 60: should this be priority 9?

mauteri: and do you need to add `enqueue_block_editor_assets` too?

PR #203, opened 2022-12-20
Dev 16dec22

pbrocks:
Removes webpack from root, adjusts css classes, builds canonical blocks, except for Attendance list and selector


pbrocks:
In the attendance-list css file, I added a green border to show when the blocks are output with the typical `wp-block-*` wrapper 
image

This shows output on frontend wrapped correctly for 3 blocks 

image

This shows the block properly wrapped on backend 

image

Issue #150, opened 2022-09-05
BUG Dates on Event Details


MervinHernandez:
The date in the date-picker, is not reflected in the event details displayed, even after saving. 

![Image](https://user-images.githubusercontent.com/8400371/188478452-b1ece94b-e701-4b13-8fe3-2b7690df201e.png)

mauteri:
You're going to have to walk me through this one. The time updates on blur. So when you change it, as soon as you leave the field, it updates. It functions exactly as changing publish date. At least for me. If you have a workflow that is different from mine, I'd like to see it.

MervinHernandez:
1. I select a date-time for the event
2. I save the event 
3. This "Start" field, does not reflect the date I selected.

pbrocks:
more discussion in [Slack](https://gatherpress.slack.com/archives/C010M3FV65V/p1665072935496199)

pbrocks:
![image](https://user-images.githubusercontent.com/5721313/209586557-709ebad4-c9a6-4e24-88a5-63c5b98d835c.png)

pbrocks:
@MervinHernandez This should be fixed/closed

PR #214, opened 2022-12-26
Add notice if timezone not set

pbrocks:
This PR addresses [this issue](https://github.com/GatherPress/gatherpress/issues/150) addressing a JS error that persists when the Timezone is not set in WP Settings.

timezone-not-set

Error message shows on admin pages, such as plugins.php, just after activation:

image

### DIFF
--- a/includes/core/classes/class-setup.php
+++ b/includes/core/classes/class-setup.php
@@ -69,6 +69,7 @@ class Setup {
 		add_filter( 'the_time', array( $this, 'get_the_event_date' ) );
 		add_filter( 'body_class', array( $this, 'body_class' ) );
 		add_filter( 'display_post_states', array( $this, 'set_event_archive_labels' ), 10, 2 );
+		add_action( 'admin_notices', array( $this, 'timezone_check_admin_notice' ) );
 	}
 
 	/**
@@ -133,6 +134,31 @@ class Setup {
 		array_unshift( $block_categories, $gatherpress_category );
 
 		return $block_categories;
+	}
+
+	/**
+	 * Display custom admin notice function
+	 *
+	 * @return void
+	 */
+	public function timezone_check_admin_notice() {
+		$timezone = get_option( 'timezone_string' );
+		if ( $timezone ) {
+			return;
+		}
+		$class   = 'notice notice-error is-dismissible';
+		$string1 = __( 'Please set', 'gatherpress' );
+		$string2 = __( 'your timezone', 'gatherpress' );
+		$string3 = __( 'in order to ensure proper GatherPress settings!', 'gatherpress' );
+
+		printf(
+			'<div class="%1$s"><p>%2$s <a href="%3$s">%4$s</a> %5$s</p></div>',
+			esc_attr( $class ),
+			esc_html( $string1 ),
+			esc_url( get_admin_url() . 'options-general.php#timezone_string' ),
+			esc_html( $string2 ),
+			esc_html( $string3 )
+		);
 	}
 
 	/**


includes/core/classes/class-setup.php
```diff
@@ -135,6 +136,31 @@ public function block_category( $block_categories ) {
...
+		$class   = 'notice notice-error is-dismissible';
+		$string1 = __( 'Please set', 'gatherpress' );
+		$string2 = __( 'your timezone', 'gatherpress' );
+		$string3 = __( 'in order to ensure proper GatherPress settings!', 'gatherpress' );
+
+		printf(
+			'<div class="%1$s"><p>%2$s <a href="%3$s">%4$s</a> %5$s</p></div>',
+			esc_attr( $class ),
+			esc_html( $string1 ),
+			esc_url( get_admin_url() . 'options-general.php#timezone_string' ),
```
mauteri, line 158: `esc_url( get_admin_url( null,  'options-general.php#timezone_string' ) )`

includes/core/classes/class-setup.php
```diff
@@ -135,6 +136,31 @@ public function block_category( $block_categories ) {
...
+		if ( $timezone ) {
+			return;
+		}
+		$class   = 'notice notice-error is-dismissible';
+		$string1 = __( 'Please set', 'gatherpress' );
+		$string2 = __( 'your timezone', 'gatherpress' );
+		$string3 = __( 'in order to ensure proper GatherPress settings!', 'gatherpress' );
+
+		printf(
+			'<div class="%1$s"><p>%2$s <a href="%3$s">%4$s</a> %5$s</p></div>',
```
mauteri, line 155: These are always tricky... Looks like correct way to handle is with a simple `wp_kses` 
example: https://codex.wordpress.org/I18n_for_WordPress_Developers

on page example:
`<p>
<?php
$url = 'http://example.com';
$link = sprintf( wp_kses( __( 'Check out this link to my <a href="%s">website</a> made with WordPress.', 'my-text-domain' ), array(  'a' => array( 'href' => array() ) ) ), esc_url( $url ) );
echo $link;
?>
</p>`

mauteri (changes requested):
Left a couple small comments.

includes/core/classes/class-setup.php
```diff
@@ -135,6 +136,31 @@ public function block_category( $block_categories ) {
...
+	 */
+	public function timezone_check_admin_notice() {
+		$timezone = get_option( 'timezone_string' );
+		if ( $timezone ) {
+			return;
+		}
+		$class   = 'notice notice-error is-dismissible';
+		$string1 = __( 'Please set', 'gatherpress' );
+		$string2 = __( 'your timezone', 'gatherpress' );
+		$string3 = __( 'in order to ensure proper GatherPress settings!', 'gatherpress' );
```
mauteri, line 152: I think instead of `GatherPress` should say something like `in order to ensure proper event date and times` or something like that. People may forget the plugin is GatherPress, but know the feature is events.

includes/core/classes/class-setup.php
```diff
@@ -135,6 +136,36 @@ public function block_category( $block_categories ) {
...
+	 * Display custom admin notice function
+	 *
+	 * @return void
+	 */
+	public function timezone_check_admin_notice() {
+		$timezone = get_option( 'timezone_string' );
+		if ( $timezone ) {
+			return;
+		}
+		$class = 'notice notice-error is-dismissible';
```
mauteri, line 149: this doesn't need to be a variable. I'd just put it in the `class`

includes/core/classes/class-setup.php
```diff
@@ -135,6 +136,36 @@ public function block_category( $block_categories ) {
...
+		printf(
+			wp_kses(
+				// translators: %1$s classes for admin_notice, %2$s url link to setting.
+				__( '<div class="%1$s"><p>Please set <a href="%2$s">your timezone</a> in order to ensure proper GatherPress settings!</p></div>', 'gatherpress' ),
```
mauteri, line 153: don't need the `div` or `p` in there, sorry if that wasn't clear. just need the `a` since that's in with the translated string.

mauteri (approved):

PR #215, opened 2022-12-27
GP-150: Refactor datetime panel.

mauteri:


pbrocks (changes requested):
Let's check that the panel and the block get the same start/end times

image

In this situation, the times are off because the Timezone is not set.

mauteri:
> Let's check that the panel and the block get the same start/end times
> 
> image
> 
> In this situation, the times are off because the Timezone is not set.

Good catch! I'll fix that.

pbrocks (approved):
Looks like it's working properly. Also, thank you for consolidating the /helpers folders.

PR #218, opened 2022-12-28
Feature/select event date or post date gp 112

pbrocks:
Adds a checkbox
image


pbrocks:
image

includes/core/classes/class-settings.php
```diff
@@ -352,22 +459,17 @@ public function get_sub_pages(): array {
...
+									'name' => __( 'Show post date as Event date', 'gatherpress' ),
 								),
 								'field'  => array(
-									'type'    => 'autocomplete',
-									'options' => array(
-										'type'  => 'page',
-										'label' => __( 'Select Page', 'gatherpress' ),
-										'limit' => 1,
-									),
+									'type' => 'checkbox',
```
mauteri, line 467: Let's add this one option to check or not check by default. in this case, check.
```
'options' => array(
    'default' => true,
),
```

includes/core/classes/class-settings.php
```diff
@@ -342,8 +449,8 @@ public function get_name_field( string $sub_page, string $section, string $optio
 	 * @return array
 	 */
 	public function get_sub_pages(): array {
-		$sub_pages = array(
-			'general'    => array(
+		$gp_admin_tabs = array(
```
mauteri, line 452: maybe `$contents`? (This is more than just tabs). Also, don't need `$gp` here since local variable so I would drop it.

src/settings/style.scss
```diff
@@ -23,3 +23,7 @@
 		}
 	}
 }
+
+.components-form-token-field__input-container {
```
mauteri, line 27: This is a general override, can we get this in `.gp-settings` so it only applies to stuff we put in.?

mauteri (changes requested):

includes/core/classes/class-settings.php
mauteri: This should be hooked to too (so when you uncheck or check you see the event or publish date)

src/settings/style.scss
pbrocks: I thought that's where I put it

includes/core/classes/class-settings.php
pbrocks: It is hooked up. I put the screenshot in PR
image

pbrocks: ```if ( ! is_a( $post, '\WP_Post' ) || Event::POST_TYPE !== $post->post_type || 1 !== intval( $gp_settings['pages']['post_or_event_date'] ) ) {```

mauteri: Cool, missed that. `if` can be simplified to just 
```
if ( Event::POST_TYPE !== get_post_type( $post ) || 1 !== intval( $gp_settings['pages']['post_or_event_date'] ) ) {
```

mauteri (approved):

PR #224, opened 2023-01-08
Bug/gp 150 2

mauteri:
Fixes issues with timezones. Keeps timezones localized to Event. TODO, need to add dropdown on single event to manually change timezone. Currently, it inherits global timezone and saves it. If timezone changes globally, the timezone is already localized to a single event.


pbrocks (commented):
@mauteri Are these CSS changes manual or linted changes?

mauteri:
> @mauteri Are these CSS changes manual or linted changes?

Linted. I merged your changes and now it's same as your changes.

pbrocks (approved):
Looks good

PR #234, opened 2023-01-21
lint:css:src

pbrocks:
Lint css doesn't seem to do anything. It just hangs in the terminal. Seems to me that the only CSS we should be linting is what is in the /src folder anyway. This PR adds that action.

lint-css-hangs


mauteri:
@pbrocks can we close this one? Just merged your other PR.

PR #244, opened 2023-01-26
adjust the output

pbrocks:


pbrocks:
overt-address

src/blocks/venue/edit.js
```diff
@@ -52,6 +52,13 @@ const Edit = ({ attributes, setAttributes }) => {
 			venuePost?.title.rendered ??
 			__('No venue selected.', 'gatherpress');
 
+		useEffect(() => {
+			setAttributes({
+				venueName: name,
+				venueAddress: fullAddress,
```
mauteri, line 58: This and line `49` I think need to adjjust to `venueAddress`

src/blocks/venue/render.php
```diff
@@ -26,13 +26,76 @@
 
 // (WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase)
 // phpcs:ignore
-$gatherpress_full_address = $gatherpress_venue_information->fullAddress;
+$gatherpress_full_address = $gatherpress_attributes['venueAddress'];
```
mauteri, line 29: Change to `$gatherpress_venue_address` should change everywhere. Like this change for consistency with the other variables.

src/blocks/venue/edit.js
```diff
@@ -52,6 +52,13 @@ const Edit = ({ attributes, setAttributes }) => {
 			venuePost?.title.rendered ??
 			__('No venue selected.', 'gatherpress');
 
+		useEffect(() => {
+			setAttributes({
+				venueName: name,
```
mauteri, line 57: name should be `venueName: venueName` I think it can just be `venueName` in that case if I'm remembering things right... same with `venueAddress`, yeah?

mauteri (approved):

PR #246, opened 2023-01-27
GP-237: Fix datetime display: issue was how GMT was saving, required …

mauteri:
…method that allowed to pass localized time zone.


pbrocks (changes requested):
Would it be possible to prepend UTC to numerical renderings of TZ offset? So, something like this:
LA-PAZ--TZ

I love this for the US TZ. Does it work for elsewhere?

LOS-ANGELES--TZ

pbrocks:
Looks like Europe works too
image

pbrocks (approved):

PR #261, opened 2023-02-10
Adjustments to Venue blocks

mauteri:


jmarx (commented):
LGTM

pbrocks:
Looks like we lost the slick functionality that @jmarx put together on the Venue Info block when changing venue info, and there's a little quirkiness with the map rendering in Event Venue block in editor.

mauteri:
> Looks like we lost the slick functionality that @jmarx put together on the Venue Info block when changing venue info, and there's a little quirkiness with the map rendering in Event Venue block in editor.

I added my reasons for removing that as a bullet point in my post in #development channel. Biggest issue is that it was not accessible (you could not access it from keyboard, you had to click on it).

mauteri:
If there's quirkiness with the map we can address that in a follow up PR.

PR #264, opened 2023-02-16
Feature/fix venue post type

mauteri:


pbrocks (approved):
Nice job, but you really don't like attributes, huh? ;o)

mauteri:
> Nice job, but you really don't like attributes, huh? ;o)

I love attributes! Just want them to make sense ;-)

PR #339, opened 2023-09-02
Jmarx unit tests testing branch wip

jmarx:


tests/includes/core/classes/class-test-query.php
```diff
@@ -120,4 +120,31 @@ public function test_get_past_events(): void {
...
+	 * @covers ::adjust_admin_event_sorting
+	 *
+	 * @return void
+	 */
+	public function test_adjust_admin_event_sorting(): void {
+		$instance = Query::get_instance();
+
+		$this->mock->user( false, 'admin' );
+		$response = $instance->adjust_admin_event_sorting( array() );
+		$this->assertEmpty( $response, 'The array is not empty.' );
```
mauteri, line 135: I've been following a convention I picked up from PMC where we write message as `Failed to assert array is not empty`

PR #495, opened 2024-01-21
WIP: RSVP Response autocomplete

jmarx:


src/components/RsvpResponseContent.js
```diff
@@ -44,11 +75,30 @@ const RsvpResponseContent = ({ items, activeValue, limit = false }) => {
...
+										<input
+											onChange={onChange}
+											aria-controls={listBoxId}
+											aria-activedescendant={activeId}
+										/>
+
+										<ul id={listBoxId}></ul>
+									</div>
+								)}
+							</Autocomplete>
```
jmarx, line 99: This is how we are supposed to implement this component. However, I am stuck on a strange error.
(error happens both on the block editor and the front end).
Screenshot 2024-01-21 at 10 10 35 AM

mauteri: @jmarx this should only exist on the backend (not the frontend). There's examples of other things in GatherPress using Autocomplete, so would take a look at those as well. Thx!

Issue #507, opened 2024-01-24
Performance Issue: get_formatted_datetime() Called too many times, Duplicate Queries in Query Monitor


deshabhishek007:
### Describe the bug

I'm Using Query Monitor on my test site. On Event Page, GatherPress\Core\Event->get_formatted_datetime() function is being called too frequently, resulting in an excessive number of database queries. This will impact the site's performance

### Steps to Reproduce

1. Navigate to the event listing page.
2. Observe the database queries via the Query Monitor plugin.

### Screenshots, screen recording, code snippet

<img width="1423" alt="Too Many Calls to get_formatted_datetime" src="https://github.com/GatherPress/gatherpress/assets/428211/43c4edaf-e025-48a6-9a0e-d34c252197ec">

### WordPress information

- WordPress Version: 6.4.2
- PHP Version: 8.3.0

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #508, opened 2024-01-24
Fix for Issue #507: Optimize get_formatted_datetime Performance

deshabhishek007:
This PR provides a solution to the performance issue reported in Issue #507. It introduces caching to the get_formatted_datetime function, reducing database query load and improving efficiency.

- Implemented caching using WordPress Transients API.
- Unique cache keys are generated based on function parameters.
- Transient Expiry time set to 12 Hours

### DIFF
--- a/includes/core/classes/class-event.php
+++ b/includes/core/classes/class-event.php
@@ -399,6 +399,14 @@ class Event {
 		string $which = 'start',
 		bool $local = true
 	): string {
+
+		$cache_key = 'formatted_datetime_' . md5($format . $which . ($local ? 'local' : 'gmt'));
+
+		$cached_date = get_transient($cache_key);
+		if ($cached_date !== false) {
+			return $cached_date;
+		}
+
 		$dt             = $this->get_datetime();
 		$date           = $dt[ sprintf( 'datetime_%s_gmt', $which ) ];
 		$dt['timezone'] = static::maybe_convert_offset( $dt['timezone'] );
@@ -419,6 +427,7 @@ class Event {
 			$date = wp_date( $format, $ts, $tz );
 		}
 
+		set_transient($cache_key, $date, 43200);
 		return (string) $date;
 	}
 


includes/core/classes/class-event.php
```diff
@@ -399,6 +399,14 @@ protected function get_formatted_datetime(
 		string $which = 'start',
 		bool $local = true
 	): string {
+
+		$cache_key = 'formatted_datetime_' . md5($format . $which . ($local ? 'local' : 'gmt'));
+
```
mauteri, line 404: I would remove this line break and line up `=` also add more whitespace to meet WordPress standards.
```
$cache_key   = 'formatted_datetime_' . md5( $format . $which . ( $local ? 'local' : 'gmt' ) );
$cached_date = get_transient( $cache_key );
```

includes/core/classes/class-event.php
```diff
@@ -399,6 +399,14 @@ protected function get_formatted_datetime(
...
+		$cached_date = get_transient($cache_key);
```
mauteri, line 405: add line break here.

includes/core/classes/class-event.php
```diff
@@ -399,6 +399,14 @@ protected function get_formatted_datetime(
...
+		if ($cached_date !== false) {
```
mauteri, line 406: Yoda notation and a bit more whitespace.
```
if ( false !== $cached_date ) {
```

includes/core/classes/class-event.php
```diff
@@ -419,6 +427,7 @@ protected function get_formatted_datetime(
 			$date = wp_date( $format, $ts, $tz );
 		}
 
+		set_transient($cache_key, $date, 43200);
```
mauteri, line 430: line break here before `return` and add white space
```
set_transient( $cache_key, $date, HOUR_IN_SECONDS * 12 );

return (string) $date;
```

mauteri (changes requested):
Thanks for the PR, @deshabhishek007! Couple small clean ups and this will be good to merge. :-)

mauteri:
Looks good, merging! Thanks!

mauteri:
@deshabhishek007 Need to revert this change, it is caching and not changing the date when the date changes. We'll need to look at this deeper to ensure this makes fewer DB calls, but does not cache the date so it cannot be changed.

mauteri:
@deshabhishek007 I reverted this code here. https://github.com/GatherPress/gatherpress/pull/516

Steps to reproduce bug I saw is to go to an event, change the date, view the event and the date is the previous date as it is still cached. We need a way to invalidate the cache on save so this doesn't happen. In next attempt, please include some unit tests to ensure this works after caching. Thanks!

PR #522, opened 2024-01-29
Rsvp form token field

### LINKED ISSUE #485: Ability to change attendees in RSVP Response block
Image

In the RSVP Response block we need the ability to edit who is attending. That includes removing and/or adding people to attending or not attending. Some notes for updates to this block.

- Ability to toggle between the two views: View and Edit. View is always the default. This should look like the HTML blocks's HTML and Preview toggles
Image
- Edit screen should include a list of those Attending in an Autocomplete component. We use the Autocomplete component for Leadership. Don't use the specific version of Autocomplete we have for Settings, but use that core component.
Image
- When someone is added to either Attending or Not Attending, they should automatically be removed from the other if they are currently listed. Either that, or they shouldn't be able to be added to one until the are manually removed from the other.
- We do not need to add or remove people for the Waiting List. This happens programmatically (see class-rsvp.php). The UI of this should honor and reflect the Waiting List.
- A user with `edit_posts` capability should be able to make this change. So make sure to do that check.

jmarx:
### Description of the Change

Administrators can now add/remove attendees to events.

Closes #485

### Credits
Props @linusx @jmarx @mauteri 

### Checklist:
- [ ] I have added tests to cover my change(pending).
- [ ] All new and existing tests pass(pending).


jmarx:
@mauteri I am throwing this pr up here as this is as far as I got. It includes most of the heavy lifting we need to do here. I have a tight week and off to AZ for a week. So I figured maybe someone could work on it in case I can't get back to it until I return next week.

includes/core/classes/class-rest-api.php
```diff
@@ -251,7 +251,7 @@ protected function events_list_route(): array {
 	 * @return bool True if the parameter is a valid RSVP status, false otherwise.
 	 */
 	public function validate_rsvp_status( $param ): bool {
-		return ( 'attend' === $param || 'attending' === $param || 'not_attending' === $param );
+		return ( 'attend' === $param || 'attending' === $param || 'not_attending' === $param || 'remove' === $param );
```
mauteri, line 254: I was thinking of `remove` when I was working on anonymous feature. I ended up just going with `attend` so we don't have to add another status. This can be confusing though for removal. 

I think we can update things a bit and remove `attend` and `remove` and replace it with `unspecified`. I think that word can be used in both cases without confusion. Thoughts?

mauteri: PR to update this https://github.com/GatherPress/gatherpress/pull/525

mauteri (changes requested):

src/blocks/rsvp-response/edit.js
```diff
@@ -20,13 +23,39 @@ import EditCover from '../../components/EditCover';
...
+			{!editMode && (
+				<EditCover>
+					<RsvpResponse />
+				</EditCover>
+			)}
+			{isAdmin && (
+				<BlockControls>
+					<ToolbarGroup>
+						<ToolbarButton
+							label="Edit Attendees"
```
mauteri, line 48: Change to just `Edit`. Make sure to add to translation function. `{__('Edit', 'gatherpress')}`. Super important we do this. We already have the plugin translated to 3 languages and we'll need to update when this feature goes out to ensure text is translated.

src/blocks/rsvp-response/edit.js
```diff
@@ -20,13 +23,39 @@ import EditCover from '../../components/EditCover';
...
+							text={
+								editMode
+									? 'Preview Attendees'
```
mauteri, line 51: Change to `Preview`. Add translation function.

src/blocks/rsvp-response/edit.js
```diff
@@ -20,13 +23,39 @@ import EditCover from '../../components/EditCover';
...
+									: 'Edit Attendees'
```
mauteri, line 52: Change to `Edit`. Add translation function.

src/components/RsvpResponse.js
```diff
@@ -62,11 +43,6 @@ const RsvpResponse = () => {
 
 	Listener({ setRsvpStatus }, getFromGlobal('post_id'));
 
-	// Make sure rsvpStatus is a valid status, if not, set to default.
-	if (!items.some((item) => item.value === rsvpStatus)) {
-		setRsvpStatus(defaultStatus);
-	}
-
```
mauteri, line 69: Looks like this file got messed up. These changes should remain. Think something happened when you updated from `main` after I pushed code. Looks like the file can just be reverted to what's on `main`.

mauteri (changes requested):

includes/core/classes/class-rsvp.php
```diff
@@ -122,6 +122,34 @@ public function get( int $user_id ): array {
...
+			return $responses;
+		}
+		foreach($responses['all']['responses'] as $key => $value ) {
+				if ($value['id'] == $user_id) {
+					global $wpdb;
+					$table = sprintf( static::TABLE_FORMAT, $wpdb->prefix );
+					$result = $wpdb->delete( $table, array( 'user_id' => $value['id'] ) );
+				}
+		}
+		return $responses;
```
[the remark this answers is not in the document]
linusx: @mauteri wpdb->delete will return an int on success, not boolean. Might have to check the $result if false, otherwise return true.

mauteri (changes requested):

includes/core/classes/class-rsvp.php
```diff
@@ -122,6 +122,35 @@ public function get( int $user_id ): array {
...
+			return false;
+		}
+		foreach($responses['all']['responses'] as $key => $value ) {
+				if ($value['id'] == $user_id) {
+					global $wpdb;
+					$table = sprintf( static::TABLE_FORMAT, $wpdb->prefix );
+					$result = $wpdb->delete( $table, array( 'user_id' => $value['id'] ) );
+				}
+		}
+		return true;
```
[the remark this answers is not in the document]
jmarx: @mauteri I just removed the `remove` function and wrapped it all in with the `save` function.

mauteri (approved):

PR #525, opened 2024-02-03
Change attend to no_status

### LINKED ISSUE #524: Change attend to no_status
Current the word `attend` is used to indicate there is no record of an RSVP status. This is confusing and hard to understand in the codebase. To fix this, we will replace the word `attend` with `no_status`

mauteri:
### Description of the Change
We use the word `attend` in the code as a status (like "attending", "not_attending", "waiting_list"). This was from early on in the plugin and was not a good choice at that time either. This PR replaces the word `attend` with `no_status`, which will work for both people that have not responded yet to an event or those that have been removed from an event.

Closes #524 

### How to test the Change
RSVP flow should remain unchanged.

### Changelog Entry

N/A

### Credits

Props @mauteri 

### Checklist:


mauteri:
Other possible names for this status...
`no_record`, `none`, something else?

PR #529, opened 2024-02-04
Allow blocks to work in FSE.

### LINKED ISSUE #526: Update block registration to be smarter
Currently, blocks are registered/unregistered based on post type. With FSE we need to do this a bit smarter...
- Blocks should still belong to their correct post type when post type is set
- FSE templates should allow GatherPress blocks at template level for single templates, but not their settings that are specific to the post itself

I don't know a TON about FSE (I don't use it much), so need others to weigh in on this to make sure our approach is appropriate.

mauteri:
### Description of the Change
GatherPress blocks were only enabled in the post block editor. This change allows GP blocks to now work in FSE templates.

Closes #526 

### How to test the Change
- Add GatherPress blocks to FSE templates
- Should be able to add all GP blocks and see no errors in editor or console
- Blocks are for display and positioning purpose only in this context; there should be no configurations (except venue map).

### Credits

Props @mauteri 

### Checklist:


src/components/MapEmbed.js
```diff
@@ -16,20 +18,24 @@
...
+	let { location, height } = props;
 
 	if (!height) {
 		height = 300;
 	}
 
-	const style = { border: 0, height, width: '100%' };
-	const baseUrl = 'https://maps.google.com/maps';
+	if (!isSinglePostInEditor() && !location) {
+		location = '60 29th Street #343, San Francisco, CA 94110, USA';
```
jmarx, line 29: @mauteri Assuming this hard-coded address is a default location to use for the map?

jmarx (approved):

jmarx:
@mauteri Looks like we have a merge conflict

PR #556, opened 2024-02-20
Adding user formats and timezone

### LINKED ISSUE #446: Allow subscribers to define their own date/time format and timezone
Would be great to allow everyone to define their own timezone and date/time format for events to be displayed in their chosen format, and event times to be shown in their chosen timezone. Settings saved in DB for each user.

linusx:
### Description of the Change
Allow subscribers to define their own date/time format and timezone

Steps to verify:

1. Log in as an end user, and go to the edit profile page.
2. You should see **Date & Time Formatting** with some fields under it.
3. Add some custom formats ( you should see the preview functionality working).
4. Go to the event page, and you should see the date/time formats match what you entered.

Closes #446

### How to test the Change
Testing instructions are above with the verification.

### Changelog Entry

> Added - New feature
Allow subscribers to define their own date/time format and timezone

### Credits

Props @linusx

### Checklist:

- [X] I agree to follow this project's [**Code of Conduct**](https://github.com/GatherPress/gatherpress/blob/main/CODE_OF_CONDUCT.md).


includes/templates/admin/user/date-time.php
```diff
@@ -0,0 +1,84 @@
...
+
+$gatherpress_date_attrs = array(
+	'name'  => 'gp_date_format',
+	'value' => ! empty( $date_format ) ? $date_format : '',
+);
+
+$gatherpress_time_attrs = array(
+	'name'  => 'gp_time_format',
+	'value' => ! empty( $time_format ) ? $time_format : '',
+);
```
mauteri, line 25: Let's move these variables to the `profile_fields` method and out of the template. Then just pass them to the template.

includes/templates/admin/user/date-time.php
```diff
@@ -0,0 +1,84 @@
...
+					foreach ( $gatherpress_tz_choices as $gatherpress_location => $gatherpress_timezones ) {
+						echo esc_html( '<optgroup label="' . $gatherpress_location . '">' );
+						foreach ( $gatherpress_timezones as $gatherpress_tz => $gatherpress_name ) {
+							echo esc_html( '<option value="' . $gatherpress_tz . '"' );
+							if ( $timezone === $gatherpress_tz ) {
+								echo ' selected';
+							}
+							echo esc_html( '>' . $gatherpress_name . '</option>' );
+						}
+						echo esc_html( '</optgroup>' );
```
mauteri, line 77: use `selected` here https://developer.wordpress.org/reference/functions/selected/

includes/templates/admin/user/date-time.php
```diff
@@ -0,0 +1,84 @@
...
+	'name'  => 'gp_time_format',
+	'value' => ! empty( $time_format ) ? $time_format : '',
+);
+?>
+<div style="margin-top: 40px; margin-bottom: 30px;">
+	<h2 id="gp-user-date-time">
+		<?php esc_html_e( 'Date & Time Formatting', 'gatherpress' ); ?>
+	</h2>
+	<div>
+		<?php esc_html_e( 'For more information read the <a href="https://wordpress.org/documentation/article/customize-date-and-time-format/">Documentation on date and time formatting</a>.', 'gatherpress' ); ?>
```
mauteri, line 32: This is going to escape the markup. Maybe echo with __ wrapped in KSES function.

includes/core/classes/class-assets.php
```diff
@@ -176,7 +176,8 @@ function ( $key ) {
 			array_keys( $settings->get_sub_pages() )
 		);
 
-		if ( in_array( $hook, $setting_hooks, true ) ) {
+		// Had to add the profile page to the check since it uses some of the same assets.
+		if ( in_array( $hook, $setting_hooks, true ) || 'profile.php' === $hook ) {
```
mauteri, line 180: Let's not put in settings as that's the wrong place. I would put in it's own JS file or in admin.js for now.

mauteri (changes requested):

includes/core/classes/class-assets.php
linusx: Not sure what you mean by this.

linusx: Refactored both this and the settings one to use a common helper.

includes/templates/admin/user/date-time.php
```diff
@@ -0,0 +1,72 @@
...
+if ( ! isset( $timezone ) && ! isset( $date_format ) && ! isset( $time_format ) ) {
+	return;
+}
+?>
+<div style="margin-top: 40px; margin-bottom: 30px;">
+	<h2 id="gp-user-date-time">
+		<?php esc_html_e( 'Date & Time Formatting', 'gatherpress' ); ?>
+	</h2>
+	<div>
+		<?php echo __( wp_kses( 'For more information read the <a href="https://wordpress.org/documentation/article/customize-date-and-time-format/" target="_blank">Documentation on date and time formatting</a>.', array( 'a' => array( 'href' => array(), 'target' => array() ) ) ), 'gatherpress' ); ?>
```
mauteri, line 20: switch these, `wp_kses( __(`

src/profile/style.scss
```diff
@@ -0,0 +1,39 @@
...
+		img {
+			border-radius: 999rem;
+			margin-bottom: 0.5rem;
+		}
+
+		li {
+			text-align: center;
+			width: 10rem;
+		}
+	}
```
mauteri, line 34: This doesn't look right. Credits should be a setting thing, but not needed here.

mauteri (changes requested):

includes/templates/admin/user/date-time.php
linusx: yeah, I realized that after I pushed.

src/profile/style.scss
linusx: Yeah, I don't think any styles are needed. Im going to leave an empty file for now.

includes/core/classes/class-user.php
```diff
@@ -72,6 +72,10 @@ protected function setup_hooks(): void {
 	public function profile_fields( WP_User $user ): void {
 		$event_updates_opt_in = get_user_meta( $user->ID, 'gp-event-updates-opt-in', true );
 
+		$default_tz   = static::get_system_timezone();
```
mauteri, line 75: Shouldn't this be
`$timezone_default = get_option( 'timezone_string' );`?

(reversed in var to match the other two format below)

includes/core/classes/class-user.php
```diff
@@ -72,6 +72,10 @@ protected function setup_hooks(): void {
...
+		$date_default = get_option( 'date_format', 'l, F j, Y' );
+		$time_default = get_option( 'time_format', 'g:i A' );
```
mauteri, line 77: GatherPress has it's own settings API you can use to get this data from what is set in those settings.
```
$settings     = Settings::get_instance();
$data_default = $settings->get_value( 'general', 'formatting', 'date_format' );
$time_default = $settings->get_value( 'general', 'formatting', 'time_format' );
```

includes/templates/admin/user/date-time.php
```diff
@@ -0,0 +1,82 @@
+<?php
+/**
+ * User Date Time Settings Template.
+ *
+ * @package GatherPress\Core
+ * @since 0.29.0
```
mauteri, line 6: I'm just keeping everything `@since 1.0.0` as that will be the first official release. These little alpha releases I'm not too worried about.

mauteri (changes requested):

includes/core/classes/class-user.php
mauteri: Ok never mind, though I think we can move this to method to `Utility` class in another PR.

includes/core/classes/class-assets.php
```diff
@@ -176,6 +176,7 @@ function ( $key ) {
 			array_keys( $settings->get_sub_pages() )
 		);
 
+		// Had to add the profile page to the check since it uses some of the same assets.
```
mauteri, line 179: We can remove this comment.

includes/core/classes/class-user.php
```diff
@@ -110,5 +171,8 @@ public function save_profile_fields( int $user_id ): void {
 		}
 
 		update_user_meta( $user_id, 'gp-event-updates-opt-in', intval( filter_input( INPUT_POST, 'gp-event-updates-opt-in' ) ) );
```
mauteri, line 173: I need to update this meta key to use underscores to keep things consistent.

includes/templates/admin/user/date-time.php
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * User Date Time Settings Template.
+ *
+ * @package GatherPress\Core
+ * @since 1.0.0
+ */
+
+use GatherPress\Core\Utility;
+
+if ( ! isset( $timezone ) && ! isset( $date_format ) && ! isset( $time_format ) ) {
```
mauteri, line 11: This can be just
`if ( ! isset( $timezone, $date_format, $time_format ) ) {`

mauteri (approved):

PR #566, opened 2024-02-24
Integrated new feature which allows users to decline attending immediately

### LINKED ISSUE #534: Ability to set "Not Attending" initially
Need a feature that allows a user to set themselves as "Not Attending" right from the start. 

Make this setting a toggled feature that is both global as well as an event specific (to override global)

stephenerdelyi:
### Description of the Change
Need a feature that allows a user to set themselves as "Not Attending" right from the start.
Make this setting a toggled feature that is both global as well as an event specific (to override global)

Closes #534 

### How to test the Change
Create new event (or edit existing). Select "RSVP" block in backend and check the new box titled "Allow attendees to select 'not attending' immediately". Navigate to plugin settings and test the new global option as well.

### Credits

Props @stephenerdelyi 

### Checklist:


stephenerdelyi:
creating new branch/PR with proper node version + code standards

Issue #557, opened 2024-02-21
Purpose, bug oder room for enhancements


carstingaxion:
### Describe your question

Dear GatherPress Team,

during my translation and while testing the plugin for some days now, some findings came up. I’m unsure if things are done by purpose, by buggy accident or just not polished (enough), so I’d like to ask first and would go on opening issues, if your answers identify my findings as bugs or potential new features.

Let’s start:
- [ ] Necessity to load JS on the frontend to render an address.
- [ ] Uses fixed non-translatable rewrite slugs for post_types and taxonomy.
- [ ] Why are unclosed issues in the DONE column of the project board?
- [ ] Venue post_type does not take care about the WordPress Geodata Standard.
- [ ] Missing i18n tooling by wp-cli in package.json
- [ ] Uses non-hierachical venue taxonomy, why?
- [ ] Weird ´get_taxonomy_registration_args()´
- [ ] Need a way to rename the "Attending" block
- [ ] Dateformat of datetime block is not translated/translatable
- [ ] Does it allow hybrid events?
- [ ] uses Google Maps by default, which could be a problem in the EU
- [ ] No location set; Results in Map shows: 60 29th St, San Francisco, CA 94110, USA
- [ ] use block-hooks instead of content filters for the two defined special pages

An amazing piece of poetry! 

Waiting greedy for your answers.
Best Carsten

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mauteri:
Thanks for this @carstingaxion. Some of these issues may be from ignorance and some are intentional. The code base believe it or not has code that is up to 4 years old, and some areas I probably haven't re-reviewed in years. I'll answer some as best I can, and feel free to put in pull requests to improve areas you think need to be revisited.

 **Necessity to load JS on the frontend to render an address.**
This can definitely be changed. I'm not sure why that whole block is set to be dynamic. Only the map really requires it.

**Uses fixed non-translatable rewrite slugs for post_types and taxonomy.**
This should also be updated as this was an international miss on our exclusively American team at the time.

**Why are unclosed issues in the DONE column of the project board?**
@jeffpaul also pointed this out and I have since closed those issues. I didn't realize they remained open when I moved them to `Done` to be honest. I'll make sure things in `Done` are closed going forward.

**Venue post_type does not take care about the WordPress Geodata Standard.**
I wasn't familiar with this standard, but this is good to know and to fix. Thanks!

**Missing i18n tooling by wp-cli in package.json**
[...]

PR #585, opened 2024-03-06
Feature/wpcli i18n tooling

carstingaxion:
As a translation contributor it is super-handy to use the i18n-tooling, that is included in `wp-cli` by default. So I would suggest adding it, at least to the development-requirements for GatherPress.  

_**But**_ ... how to install `wp-cli` is heavily different depending on the system, you want to run it on. This is well documented at [Installing – WP-CLI – WordPress.org](https://make.wordpress.org/cli/handbook/guides/installing/). That's why we can not or should not just add it it to GatherPress' composer dependencies.

### Description of the Change

This PR only *suggests* the wp-cli package via composer.josn, but leaves the decision open on how and where to install it.

Assuming `wp-cli` is installed now per the suggestion within the dev environment, someone can now use three new scripts to create translation files for GatherPress.

```
npm run i18n:pot
npm run i18n:json
npm run i18n:php
```
The third script provides access to the new *performant translations* functionality, introduced in WP 6.5 and already mentioned by @patriciabt [over here](https://github.com/GatherPress/gatherpress/issues/557#issuecomment-1971181992). 

~~Closes #~~

### How to test the Change

...

### Changelog Entry

> Added - i18n tooling by wp-cli

### Credits

Props @carstingaxion 

### Checklist:


carstingaxion:
Seems, we need to exclude the newly generated *performant translation* files from the CS checks.

mauteri:
> Seems, we need to exclude the newly generated _performant translation_ files from the CS checks.

Just committed change for that for you.

mauteri (approved):

carstingaxion:
Incredible, how fast you were @mauteri. Thanks!

PR #591, opened 2024-03-08
changed default venue address; removed default address from displaying on the frontend

### LINKED ISSUE #568: No location set; Results in Map shows: 60 29th St, San Francisco, CA 94110, USA
This is showing up on Front End too and should be fixed ONLY to show up on backend when using FSE to add the block to a template.

stephenerdelyi:
### Description of the Change
Map with updated default address now displays on backend only (FSE or WP admin page).
Closes #568 

### How to test the Change
Download, create new venue without map address information; navigate to new event and select the new venue with no address. You will see the new default address of WP foundation (660 4th Street San Francisco CA 94107)

### Credits
@stephenerdelyi 

### Checklist:


stephenerdelyi:
Noting the test needs to be updated, can take care of this tomorrow morning!

src/components/MapEmbed.js
```diff
@@ -18,15 +21,16 @@ import { isSinglePostInEditor } from '../helpers/globals';
...
 	const { zoom, type, className } = props;
 	let { location, height } = props;
 
 	if (!height) {
 		height = 300;
 	}
 
-	if (!isSinglePostInEditor() && !location) {
-		location = '60 29th Street #343, San Francisco, CA 94110, USA';
+	if (isAdmin && !location) {
```
mauteri, line 32: @stephenerdelyi we want to make sure this only happens in FSE, widget areas, etc and not on a single post edit because in single post edit it's expected someone would set the actual address and not need a default. Thx!

mauteri (changes requested):

src/components/MapEmbed.js
stephenerdelyi: Hey @mauteri, just pushed an update to this whenever you have a moment to check it out. It now checks to see if it's on the post editor (as opposed to other locations, like FSE/widgets, etc).

Let me know if you're having any issues or have other questions/thoughts

mauteri: Hey @stephenerdelyi mind syncing this PR? Thanks!

stephenerdelyi: @mauteri syncing now!

src/components/MapEmbed.js
```diff
@@ -18,15 +21,17 @@ import { isSinglePostInEditor } from '../helpers/globals';
  * @return {JSX.Element} The rendered React component.
  */
 const MapEmbed = (props) => {
+	const isAdmin = select('core')?.canUser('create', 'posts');
+	const isPostEditor = select('core/edit-post') ? true : false;
```
mauteri, line 25: Non-blocking, but would do something like `const isPostEditor = Boolean(select('core/edit-post'));`

src/components/MapEmbed.js
```diff
@@ -18,15 +21,17 @@ import { isSinglePostInEditor } from '../helpers/globals';
...
 	let { location, height } = props;
 
 	if (!height) {
 		height = 300;
 	}
 
-	if (!isSinglePostInEditor() && !location) {
-		location = '60 29th Street #343, San Francisco, CA 94110, USA';
+	if (isAdmin && !isPostEditor && !location) {
+		location = '660 4th Street #119 San Francisco CA 94107';
```
mauteri, line 34: Need to update Jest test below with new address.

src/components/MapEmbed.js
mauteri, line 25: would do something like `const isPostEditor = Boolean(select('core/edit-post'));` so it looks a bit cleaner.

mauteri (changes requested):

src/components/MapEmbed.js
stephenerdelyi: Finally got that test updated. Since only admin should see the default address map, the test now checks for an empty container, as that is the now-expected result.

mauteri (approved):

PR #597, opened 2024-03-09
GP-596: Make online link available after user says they are attending.

### LINKED ISSUE #596: UPDATE Show Link to Attendees

mauteri:
### Description of the Change
This change reverts a change made earlier in the project that only made the event link available during the time of the event. Now the event link is available to whoever says they are attending.

Closes #596 

### Changelog Entry

### Credits

Props @mauteri, @nilovelez 

### Checklist:


mauteri:
@nilovelez @MervinHernandez if we want to add the online event to the calendar, we are going to need to make that block dynamic, which is fine, but will take a bit of work. Basically, we can make that a react component like some of the other blocks that listens for changes in RSVP and updates its link from API response. All the pieces for this exist int the Event class and just need to be included in the RSVP response.

PR #616, opened 2024-03-19
Introduce wp.org plugin-guidelines check action

carstingaxion:
With the set goal in mind to become a versatile community plugin within the WordPress sphere, GatherPress should at **every time** align with the .org/plugin review standards, at least with the automated part of the review process.

This is where the [Plugin Check (PCP)](https://wordpress.org/plugins/plugin-check/) WordPress plugin comes into play, because thankfully there is also a [corresponding github action](https://github.com/swissspidy/wp-plugin-check-action).
 

### Description of the Change

After the overwhelming feedback we got from the WordPress plugin review team we should introduce this action into our default `push` and `pull_request` workflows. This would also be a nice aspect to promote further, saying "We align with the guidelines not only once (like normal), but with every release".

~~Closes #~~

### How to test the Change

I hope to see the test running within this PR, to get a basic testing.

### Changelog Entry

> Added - Test against standards for the WordPress.org plugin directory

### Credits

Props @carstingaxion 

### Checklist:


carstingaxion:
The test is running and correctly complaining about some stuff. My last two commits, where not needed by the PR itself, but to make sure the test is recognizing changes. 

Now there is one thing left, I'm unsure about. Please, @mauteri could you have a look at this screenshot. What is the problem with the custom table ?

grafik

mauteri:
Sure, I'll take a look. Thx!

.github/workflows/wp-plugin-check-action.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build plugin
+        run: |
+          npm run build
+          wp dist-archive . ./gatherpress.zip
+          mkdir tmp-build
+          unzip gatherpress.zip -d tmp-build
```
mauteri, line 42: @carstingaxion the error looks like it's related to the plugin not being activated. on activation it creates the tables if they do not exist. maybe try `wp plugin activate gatherpress` here?

carstingaxion: You may be right to some point, but I think at this time, _we_ don't even have a WordPress running, where we could activate anything. This happens in the next lines. I [asked for help and suggested your idea](https://github.com/swissspidy/wp-plugin-check-action/issues/61); let's see and wait a little.

swissspidy: You don't need to activate anything.

Right now I don't see why this _warning_ with the database table happens, as I can't reproduce it locally. But you should be able to just ignore it, it's just a notice/warning.

The GitHub Action is still working fine.

The only _error_ is that you are using an `Update URI` header, which is disallowed on dotorg :-)

But it looks like you removed that already in 9a18e9f656cefbec3e475fa94eb0b2a74b2c86b4

carstingaxion: Thanks again for helping  @swissspidy!

With [1.0.4](https://github.com/swissspidy/wp-plugin-check-action/releases) in place, I'm happy to run this again.

PR #622, opened 2024-03-27
Fix SQL calls with %i placeholder

### LINKED ISSUE #614: WP.ORG: Unsafe SQL calls
When making database calls, it's highly important to protect your code from SQL injection vulnerabilities. You need to update your code to use wpdb calls and prepare() with your queries to protect them.

Please review the following:
https://developer.wordpress.org/reference/classes/wpdb/#protect-queries-against-sql-injection-attacks
https://codex.wordpress.org/Data_Validation#Database
https://make.wordpress.org/core/2012/12/12/php-warning-missing-argument-2-for-wpdb-prepare/
https://ottopress.com/2013/better-know-a-vulnerability-sql-injection/
Example(s) from your plugin:
21_21-26-52_gatherpress/includes/core/classes/class-rsvp.php:306 $data        = (array) $wpdb->get_results( $wpdb->prepare( 'SELECT user_id, timestamp, status, guests, anonymous FROM ' . esc_sql( $table ) . ' WHERE post_id = %d LIMIT %d', $event_id, $total_users ), ARRAY_A ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery
# There is a call to a wpdb::prepare() function, that's correct.
# You cannot add calls like "esc_sql($table)" directly to the SQL query.
# Using wpdb::prepare($query, $args) you will need to include placeholders for each variable within the query and include the variables in the second parameter.

21_21-26-52_gatherpress/includes/core/classes/class-rsvp.php:120 $data = $wpdb->get_row( $wpdb->prepare( 'SELECT id, timestamp, status, guests, anonymous FROM ' . esc_sql( $table ) . ' WHERE post_id = %d AND user_id = %d', $event_id, $user_id ), ARRAY_A ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
# There is a call to a wpdb::prepare() function, that's correct.
# You cannot add calls like "esc_sql($table)" directly to the SQL query.
# Using wpdb::prepare($query, $args) you will need to include placeholders for each variable within the query and include the variables in the second parameter.

... out of a total of 4 coincidences.

mauteri:
### Description of the Change
In WP 6.2 the %i placeholder was made available for table or field names in a prepare statement. This change utilizes that feature to improve SQL calls.

Closes #614

### How to test the Change
All queries should remain same.

### Changelog Entry

> Security - Vulnerability

### Credits

Props @mauteri 

### Checklist:

### DIFF
--- a/includes/core/classes/class-event-query.php
+++ b/includes/core/classes/class-event-query.php
@@ -337,9 +337,9 @@ class Event_Query {
 		$current = gmdate( Event::DATETIME_FORMAT, time() );
 
 		if ( 'upcoming' === $type ) {
-			$pieces['where'] .= $wpdb->prepare( ' AND %s.datetime_end_gmt >= %s', $table, $current );
+			$pieces['where'] .= $wpdb->prepare( ' AND %i.datetime_end_gmt >= %s', $table, $current );
 		} elseif ( 'past' === $type ) {
-			$pieces['where'] .= $wpdb->prepare( ' AND %s.datetime_end_gmt < %s', $table, $current );
+			$pieces['where'] .= $wpdb->prepare( ' AND %i.datetime_end_gmt < %s', $table, $current );
 		}
 
 		return $pieces;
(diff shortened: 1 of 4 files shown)


stephenerdelyi (approved):
lgtm!

PR #645, opened 2024-04-25
Update gatherpress prefix.

### LINKED ISSUE #612: WP.ORG: Generic function/class/define/namespace/option names
All plugins must have unique function names, namespaces, defines, class and option names. This prevents your plugin from conflicting with other plugins or themes. We need you to update your plugin to use more unique and distinct names.

A good way to do this is with a prefix. For example, if your plugin is called "Easy Custom Post Types" then you could use names like these:
function ecpt_save_post()
define( 'ECPT_LICENSE', true );
class ECPT_Admin{}
namespace ECPT;
update_option( 'ecpt_settings', $settings );

Don't try to use two (2) or three (3) letter prefixes anymore. We host nearly 100-thousand plugins on WordPress.org alone. There are tens of thousands more outside our servers. Believe us, you’re going to run into conflicts.

You also need to avoid the use of __ (double underscores), wp_ , or _ (single underscore) as a prefix. Those are reserved for WordPress itself. You can use them inside your classes, but not as stand-alone function.

Please remember, if you're using _n() or __() for translation, that's fine. We're only talking about functions you've created for your plugin, not the core functions from WordPress. In fact, those core features are why you need to not use those prefixes in your own plugin! You don't want to break WordPress for your users.

Related to this, using if (!function_exists('NAME')) { around all your functions and classes sounds like a great idea until you realize the fatal flaw. If something else has a function with the same name and their code loads first, your plugin will break. Using if-exists should be reserved for shared libraries only.

Remember: Good prefix names are unique and distinct to your plugin. This will help you and the next person in debugging, as well as prevent conflicts.

Analysis result:
# This plugin is using the prefix "gatherpress" for 12 element(s).

# Looks like there is an element not using common prefixes.
21_21-26-52_gatherpress/includes/core/classes/class-setup.php:665 update_option('gp_suppress_membership_notification', true);

mauteri:
### Description of the Change
Prepends all data and globals with `gatherpress` rather than `gp` or nothing.

Closes #612

### How to test the Change
Make sure no functionality has changed after checking out branch and updating data with [GatherPress Alpha](https://github.com/GatherPress/gatherpress-alpha). Also re-save permalinks before testing.

### Changelog Entry

### Credits

Props @mauteri 

### Checklist:


stephenerdelyi:
This looks great on my end! I was able to successfully run both plugins and see expected results. I would like to note: I tested on a relatively small environment instance, we might want to also test on a higher-volume site?

mauteri:
> This looks great on my end! I was able to successfully run both plugins and see expected results. I would like to note: I tested on a relatively small environment instance, we might want to also test on a higher-volume site?

Thanks Steven, we'll have a few others test this as well. @jmarx do you mind testing this new branch with the GatherPress Alpha plugin to ensure things work as expected? Thanks!

Issue #626, opened 2024-03-29
blocks in the big picture


carstingaxion:
### Is your enhancement related to a problem? Please describe.

_**(THIS IS A DRAFT .... I'm still writing this issue)**_

---

As GatherPress strives to stay close to WordPress core development, this issue should provide on overview of recent changes within GatherPress' overall **block architecture**.

Leading goal is to **have greatest flexibility while using the blocks, for the end-user as for developers** as well. This could also help lowering the maintenance time needed, to keep GatherPress' blocks in sync with actual WordPress core development.

## This involves two main aspects:

1. Allow for flexible styling of all (our) blocks, while still relying on any conventions given by the current theme.
2. Allow for flexible handling, insertion and removing of any blocks by extenders and themers.

## 1. Flexible block styling and full theme compatibility

As already pointed out extensively, Block Variations provide (actually) one of the best ways to extend WordPress. That's why GatherPress will convert as much of the existing custom blocks into core block variations with the needed extra functionality added.

While block variations seem to be the best option, there are stil some caveats to think about or discuss:

### 1.1 Caveats of block variations

- [ ] Naming things
- [ ] Icons
- [ ] the buggy isActive callback
- [ ] Visibility scope within the editor (simliar to ...)

### 1.2 Blocks under construction

- [ ] _Event List_
- [ ] _Venue_
   - [ ] NEW _Venue Website_
   - [ ] NEW _Venue Telephone_
   - [ ] NEW _Venue Map_
- [ ] _Add to calendar_
- [ ] _Event date_
- [ ] _RSVP_
- [ ] _RSVP Response_

### 2.1 Using the Block Hooks API

Since WordPress 6.5 the Block Hooks API allows to set attributes on injected blocks. Until 6.5 WordPress it was only allowed to hook in blocks by their name. This addition was a necessary need to use this feature in GatherPress, because since we (will) don't have custom blocks anymore. 

Currently GatherPress _decides_ what blocks will be loaded by default using the `template` property of the registered post types for venues and events respectively. 

Instead of loading _blocks_ directly into the editor, we are going to load _patterns_ now. Patterns allow to be filtered by the (upgraded since 6.5) Block Hooks API. Making use of this API brings at some advantages, which are at least:

- GatherPress' blocks can be easily moved, modified or removed by extenders via standardised core code
- GatherPress provides central entry points for plugin developers to hook in own blocks, to extend GatherPress
[...]

PR #654, opened 2024-05-04
Do not load admin-CSS and -JS on every wp-admin request

### LINKED ISSUE #636: Do not load admin-CSS and -JS on every wp-admin request

As pointed out by @shawfactor in #602 it is not necessary to load assets like CSS and JS files on every wp-admin request, if they are not needed. 

And because GatherPress does unfortunately load CSS and JS files when not needed, I think we could improve this.

We have two files that are loaded everywhere:

1. `build/admin.js`
2. `build/admin_style.css`

Both can be handled a little stricter, I think.

While the `admin.js` currently _only_ removes unwanted blocks, i would suggest loading this file only within the editor, probably using `enqueue_block_editor_assets`. This seems to be trivial and a low hanging fruit, because it would benefit a lot of requests, when not loading useless stuff.

Second, the `admin_style.css` is a little trickier, because this one is used to style the notification bar, that could be shown everywhere within wp-admin. Avoiding this _overall-approach_ could be the second good reason to only show this notification bar within our own admin-pages aka `screen_ID`s, [as I already suggested](https://github.com/GatherPress/gatherpress/issues/421#issuecomment-2027832429). Showing this notification-bar only on some admin-pages - GatherPress' own ones - would open the door, to also prevent loading this CSS on every request.

As a compromise for now, we could at least check against the `gp_suppress_membership_notification` option to decide, if this CSS should be loaded or not. 

If we would like to achieve semi-perfection level, we could also transform the current `wp_enqueue_script()` call in [class-assets.php](https://github.com/GatherPress/gatherpress/blob/main/includes/core/classes/class-assets.php#L209-L215) into an `wp_register_script()` call and move the `wp_enqueue_script()` somewhere into [`check_users_can_register()`](https://github.com/GatherPress/gatherpress/blob/main/includes/core/classes/class-setup.php#L400-L423)

### Designs

_No response_

_No response_

### Code of Conduct

carstingaxion:
### Description of the Change

GatherPress does unfortunately load CSS and JS files in `/wp-admin` when not needed.
I think we could improve this.

Detailed explanation can be found on the issue itself: #636 

Closes #636

#### Test changes to `admin.js`

1. Go to `wp-admin/post-new.php?post_type=gatherpress_event`
2. Open the developer tools of your browser
3. Open the network tab 
4. Find the `admin.js` in the list of requested files
5. Go to `wp-admin`
6. Make sure the `admin.js` is NOT in the list of requested files

#### Test changes to `admin-style.css`

1. Login to `wp-admin` as an admin
2. Go to `wp-admin/options.php`
3. Set `users_can_register` and `gatherpress_suppress_membership_notification` to `0`
4. Find the admin-notice on GatherPress-related admin pages like `wp-admin/edit.php?post_type=gatherpress_event`
5. But you shouldn't see the notice on the Dashboard or somewhere else.
6. Logout
7. Login to `wp-admin` as an editor
8. You shouldn't see the admin-notice at all 

### Changelog Entry

> Changed - Do not load admin-CSS and -JS on every wp-admin request

### Credits

Props @carstingaxion 

### Checklist:


carstingaxion:
This one is finished, I think. 

@mauteri could you have a look please?

includes/core/classes/class-assets.php
```diff
@@ -253,6 +241,18 @@ public function editor_enqueue_scripts(): void {
...
+
+		$asset = $this->get_asset_data( 'admin' );
+
+		wp_enqueue_script(
+			'gatherpress-admin',
+			$this->build . 'admin.js',
+			$asset['dependencies'],
+			$asset['version'],
+			true
+		);
```
mauteri, line 253: Might be better to just combine `gatherpress-admin` with `gatherpress-editor`. I think the separation of these files was an oversight early in the project.

includes/core/classes/class-setup.php
```diff
@@ -400,11 +400,14 @@ protected function create_tables(): void {
...
-			filter_var( get_option( 'gatherpress_suppress_membership_notification' ), FILTER_VALIDATE_BOOLEAN )
+			filter_var( get_option( 'gatherpress_suppress_membership_notification' ), FILTER_VALIDATE_BOOLEAN ) ||
+			filter_var( ! current_user_can( 'manage_options' ), FILTER_VALIDATE_BOOLEAN ) ||
+			false === strpos( get_current_screen()->id, 'gatherpress' )
+
 		) {
 			return;
 		}
-
+		wp_enqueue_style( 'gatherpress-admin-style' );
```
mauteri, line 410: I think this is fine for now, but feel like we are going to revisit this method as well as probably what is in the styling.

mauteri (changes requested):

includes/core/classes/class-setup.php
carstingaxion: Agree. 

While we have the beginning of support from core, there is still now way to persist a _Dismiss_ .

[grafik](https://make.wordpress.org/core/2023/10/16/introducing-admin-notice-functions-in-wordpress-6-4/)

https://make.wordpress.org/core/2023/10/16/introducing-admin-notice-functions-in-wordpress-6-4/

includes/core/classes/class-assets.php
carstingaxion: Ok. I combined the files into one, without doing anything further, because _I know_ we will have to revisit this code for the next release, when we look into https://github.com/GatherPress/gatherpress/issues/626

carstingaxion:
Oops, @mauteri please ignore the request for review. I clicked it by accident.

mauteri (approved):

mauteri:
@carstingaxion you are okay with me merging this one, yes?

PR #692, opened 2024-06-14
Use comments table for RSVP instead of custom RSVP table.

### LINKED ISSUE #687: Make RSVPs a custom comment type

- #523 
- #665 

### Designs

- RSVP status could become a custom taxonomy registered on the comments object type. Here In pseudo-code: 
   `<?php register_taxonomy_for_object_type( 'rsvp_status', 'comment‘ ); ?> `

#### Advantages:

- allows to create variations of comment related blocks
- Does not need a user-account to RSVP
- No need for the currently used custom DB table

#### Problems:
- potential to being accessible in public via comment_feed
- https://core.trac.wordpress.org/ticket/12668
- with no reliance on a logged-in user, need to rethink https://github.com/GatherPress/gatherpress/pull/597

#### further Reading
- like a [user taxonomy](https://usersinsights.com/wordpress-user-taxonomy/)

_No response_

### Code of Conduct

mauteri:
### Description of the Change
This change utilizes the Comments table with a comment type of `gatherpress_rsvp` in WordPress instead of a custom RSVP table. It also adds code for a comment tax_query, which isn't used at the moment, but will likely come in handy soon. This requires an update to `GatherPress Alpha` to migrated old RSVPs in custom table to Comments table, then delete the custom table.

Closes #687

### How to test the Change
* RSVP should work as it always has, but RSVPs will now be stored in Comments table.
* A custom taxonomy `_gatherpress_rsvp_status` is also being used for attending, not_attending, waiting_list
* Old RSVPs should not appear and need to be migrated with a change in GatherPress Alpha plugin.

### Changelog Entry

### Credits

Props @mauteri, @carstingaxion 

### Checklist:


stephenerdelyi (approved):
Just tested this by:
- reviewing that my existing test events (with attendees previously) are no longer displaying upon checkout + build
- signing up for a test event from the frontend
- ensuring that user was recorded
- added a test user under a different account
- successfully deleted / managed users from the gutenberg backend

No unexpected results were produced as a result of these updates. Marking as approved, though we should plan to update alpha to migrate over the existing data 🎉

jmarx (approved):

PR #698, opened 2024-06-19
Max attendance setting not working

mauteri:
### Description of the Change
Bug in Max attendance setting for events where it did not function.

### How to test the Change
* Change max attendance setting in global settings and see it appear as value on new events.
* Change value of max attendance on an event and that will be the max attendance for that event.
* Set value to 0 and there is no max attendance limit.

### Changelog Entry

> Fixed - Bug fix

### Credits

Props @mauteri

### Checklist:


steveerdelyi:
@mauteri this looks good on my end - I tested using your recommended test path above and got expected results

mauteri:
> @mauteri this looks good on my end - I tested using your recommended test path above and got expected results

Thanks, I'll look to get this out today :-)

PR #716, opened 2024-06-28
Feature 681: Adding uninstall method

### LINKED ISSUE #681: Add uninstall.php or uninstall routine

Not that I want anybody to uninstall GatherPress, but people should be able to do that without any worries.

Following [WordPress best-practices for uninstallation](https://developer.wordpress.org/plugins/plugin-basics/uninstall-methods/), _GatherPress_ should have something like this somewhere.

```php
// if uninstall.php is not called by WordPress, die
if ( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
    die;
}

$option_name = 'wporg_option';

delete_option( $option_name );

// for site options in Multisite
delete_site_option( $option_name );

// drop a custom database table
global $wpdb;
$wpdb->query( "DROP TABLE IF EXISTS {$wpdb->prefix}mytable" );
```
> In Multisite, looping through all blogs to delete options can be very resource intensive.

### Designs

Shouldn't be to much effort, if the existing & related methods can be re-used and we don't forget about the transients in #680 

https://github.com/GatherPress/gatherpress/blob/4942ce87071662cddcb8c993ae42893cad303830/includes/core/classes/class-setup.php#L196-L198

and 

https://github.com/GatherPress/gatherpress/blob/4942ce87071662cddcb8c993ae42893cad303830/includes/core/classes/class-setup.php#L307-L314

_No response_

### Code of Conduct

bluecollarcoders:
### Description of the Change
This PR adds an uninstall method to clear out any plugin options, settings specific to the plugin, and/or other database entities such as tables.

- screenshots or screencasts

Closes #

### Changelog Entry

> Added - New feature

### Credits

Props @bluecollarcoders 

### Checklist:

### DIFF
--- a/includes/core/classes/class-setup.php
+++ b/includes/core/classes/class-setup.php
@@ -237,10 +237,11 @@ class Setup {
 		}
 
 		// Check if the user opted to delete data on uninstall.
-		$delete_data = get_option( 'delete_data_on_uninstall', false );
-
+		$settings    = Settings::get_instance();
+		$delete_data = $settings->get_value( 'general', 'general', 'delete_data_on_uninstall' );
+		
 		// If the option is not enabled, exit early to prevent data on uninstall.
-		if( ! $delete_data ) {
+		if( intval( $delete_data )  !== 1 ) {
 			return;
 		}
 	
--- a/includes/core/classes/class-setup.php
+++ b/includes/core/classes/class-setup.php
@@ -83,6 +83,7 @@ class Setup {
 	protected function setup_hooks(): void {
 		register_activation_hook( GATHERPRESS_CORE_FILE, array( $this, 'activate_gatherpress_plugin' ) );
 		register_deactivation_hook( GATHERPRESS_CORE_FILE, array( $this, 'deactivate_gatherpress_plugin' ) );
+		register_uninstall_hook( GATHERPRESS_CORE_FILE, array( 'GatherPress\Core\Setup', 'uninstall_gatherpress_plugin' ) );
 
 		add_action( 'init', array( $this, 'load_textdomain' ), 9 );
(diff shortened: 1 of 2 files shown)


includes/core/classes/class-setup.php
```diff
@@ -218,6 +219,58 @@ public function deactivate_gatherpress_plugin(): void {
...
+	public function uninstall_gatherpress_plugin(): void {
+		// Ensure the uninstallation script is called from within WordPress.
+		if( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
+			die;
+		}
+
+		global $wpdb;
+
+		// Delete Options.
+		$option_name = 'wporg_option';
```
mauteri, line 253: This was just an example. We need to remove things we store in the options table (GatherPress settings). In addition to deleting options, should delete user meta, taxonomies, events, venues, etc. Some things like post meta will be deleted automatically with the posts. We might want to delete things in options table by looking for anything that prepends with `gatherpress_`. What do you think @carstingaxion?

carstingaxion: Take a look at https://github.com/stephenharris/Event-Organiser/blob/65c4e3b0f40a601bebf7d5fc2203da02bbb4d7e9/includes/event-organiser-install.php#L404 as a nice example.

carstingaxion: What I especially liked about the linked implementation was:

- [ ] additional check if plugin is network-enabled 
- [ ] deletion of user_meta

In addition:

For the multiple options I suggest an array function like `array_walk` calling `delete_option` on the list of all options and transients as well.

mauteri:
Cleaning up old PRs

PR #718, opened 2024-07-01
Remove language files and methods.

mauteri:
### Description of the Change
Now that GatherPress is on wordpress.org, we can use https://translate.wordpress.org/projects/wp-plugins/gatherpress/ to manage translations, rather than locally supporting translations

### How to test the Change
I'll need those that know more than I to confirm that we no longer need any of this code that is being removed.

### Changelog Entry

### Credits

Props @mauteri

### Checklist:


patriciabt:
I confirm the /languages folder and all its files can be removed.
I'll let someone else confirm the changes to the classes files.

PR #731, opened 2024-07-14
Reflect changes to the venue term, when a venue post gets deleted.

### LINKED ISSUE #730: venue terms are not deleted with venue posts
### Describe the bug

The terms of the shadow taxonomy `_gatherpress_venue` are not deleted, when the corresponding `gatherpress_venue` posts are deleted.

This is happening, because a trashed post gets its post_name changed to `'original' + '__trashed'`, which is not reflected on the venue terms, because the terms get only updated on status `'published'`. On post deletion, the slug of the post to delete gets compared to existing venue term slugs. Now, that nothing matches, the old and original venue term stays in the DB forever.

### Steps to Reproduce

1. Take a look at your wp_terms DB table
2. You'll find all venues, you ever created ;)

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

carstingaxion:
### Description of the Change

Closes #730

### Changelog Entry

> Fixed - The terms of the shadow taxonomy `_gatherpress_venue` were not deleted, when the corresponding `gatherpress_venue` posts were deleted

### Credits

Props @carstingaxion, @mauteri 

### Checklist:

### DIFF
--- a/includes/core/classes/class-venue.php
+++ b/includes/core/classes/class-venue.php
@@ -261,8 +261,8 @@ class Venue {
 			return;
 		}
 
-		// Only proceed if the venue post is being published.
-		if ( 'publish' !== $post_after->post_status ) {
+		// Only proceed if the venue post is being published or trashed.
+		if ( 'publish' !== $post_after->post_status && 'trash' !== $post_after->post_status ) {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


mauteri:
@carstingaxion added a small change to the logic to keep it one conditional check and updated unit test. Good find on this bug!!

carstingaxion:
Cool thing, you updated the test as well! Really.

Issue #684, opened 2024-06-10
NEW Event Date block 


carstingaxion:
### Is your enhancement related to a problem? Please describe.

A new or updated Event Date block should …:

- [ ] show the start date only
- [ ] Show the end date only
- [ ] Show start and end date
- [ ] use the time and date formats saved in settings, by default 
- [ ] Allow to set a custom format
- [ ] Should have proper documentation, incl. https://github.com/GatherPress/gatherpress/issues/727#issuecomment-2226563698

With that in place one could easily create many different layouts using one or multiple instances of the Event Date block within a singular or archive template, or even within a query block.

The UI for both dates should be merged into one, because people will probably ever want to define an end date, after they defined a start date. This should the save two-clicks for closing start- and opening end-date-fields.

### Designs

The existing `core/post-date` has the most of what is needed, but is not prepared well to create block-variations from. Maybe GatherPress can re-use some of its components.

Fixing the variation-problem in the [source at Gutenberg](https://github.com/WordPress/gutenberg/issues?q=is%3Aissue+is%3Aopen+post-date+label%3A%22%5BBlock%5D+Post+Date%22), would be best.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mauteri:
NOTES: https://github.com/WordPress/gutenberg/issues/60971

MervinHernandez:
✅ Viewed 684 - Event Date Block - Mike has it

MervinHernandez:
✅ Reviewed June 22 - Mike has it

mauteri:
This will need to be a custom block as a block variation does not appear to be possible. I can still explore a bit, but I think the translations aspects of this make it pretty much impossible. We can still explore ways of replicating things done in the Post Date block to improve our current block. 

https://github.com/WordPress/gutenberg/tree/trunk/packages/block-library/src/post-date

JordanPak:
I think I can take a look at this @mauteri @ashortpmc

JordanPak:
Discussed in today's weekly huddle:

We want the event date block to be flexible and highly customizable, like shown in [the example](https://github.com/GatherPress/gatherpress/issues/684#issuecomment-3144429390). Adding block-based attributes for the date/time formatting that override the plugin's settings makes sense, but we also have overrides at the **user** level (in their profile). This would basically break arrangements like the example where multiple event date blocks are used with different formatting, since every block would just get overridden multiple times, resulting in a ton of redundant and buggy displays.

[...]

PR #752, opened 2024-08-01
⚠️ Load later for i18n (Enhance i18n of attendees-email...)

### LINKED ISSUE #651: Set user-timezone & user-locale in emails

The emails ~~reminders~~ don't respect the user-defined timezone and not the users locale.

### Designs

_No response_

_No response_

### Code of Conduct

carstingaxion:
### Description of the Change

Originally planned to only close #651, some additional findings are solved with this PR, too:
- [Load users locale before sending emails](https://github.com/GatherPress/gatherpress/pull/752/commits/ba03e3861740ca79c02da28cff9196189f0b62de) _(the first task or better rabbit hole)_
- [Make sure translation files are loaded properly.](https://github.com/GatherPress/gatherpress/pull/752/commits/67c5ff7da38b3f8eddb4a4b6b3d6da7ba349be20) 
This is important for users who selected a different from the sites default language.
- [Set html language attributes correctly for the email](https://github.com/GatherPress/gatherpress/pull/752/commits/4fec72876054bd736b6af1ca4591f27448403fea)
- [Conditionally render the featured image in the email, only if it is set.](https://github.com/GatherPress/gatherpress/pull/752/commits/da5707162467ecd19b262160e592a1f710fc7bba)
- [Set the current user to the actual member to mail to,](https://github.com/GatherPress/gatherpress/pull/752/commits/eb209060375ec135afb6e995392336ccdfb92cd0) to make sure the GatherPress filters for date- and time- format, as well as the users timezone, are recognized by the functions inside `render_template()`.
- [Bug fix: Keep slashes in user-defined date- and time-formats](https://github.com/GatherPress/gatherpress/pull/752#issuecomment-2308986251)

### How to test the Change

Testing this PR might be a little tricky and is not possible with WordPress Playground because of the necessary sending & receiving of emails.

#### Preparation to test:

1. Setup at least two different users
2. One of the users creates an event, the other users attends. _(This will make sure, 2 emails are sent for the test)_
3. Assign a non-site-default language to one of the users in their `profile.php`
4. Go to _Updates_, get to the bottom of the page and hit the "**Update translations**" button.
5. Make sure that in the following screen, the translation files for the selected non-site-default language gets downloaded!

#### Testing this PR:

1. Edit the newly created event
2. **Compose** a message via Email
3. Send email
4. Check the results, one email should follow the user-defined language, while the other is sent in the default locale

### Changelog Entry

> Fix sending localised emails

### Credits

Props @carstingaxion 

### Checklist:

### DIFF
--- a/includes/core/classes/class-rest-api.php
+++ b/includes/core/classes/class-rest-api.php
@@ -457,7 +457,7 @@ class Rest_Api {
 				continue;
 			}
 			if ( $member->user_email ) {
-				$to = $member->user_email;
+				$to              = $member->user_email;
 				$switched_locale = switch_to_user_locale( $member->ID );
 
 				// Set the current user to the actual member to mail to,
--- a/includes/core/classes/class-rest-api.php
+++ b/includes/core/classes/class-rest-api.php
@@ -449,18 +449,6 @@ class Rest_Api {
 		}
 
 		$members = $this->get_members( $send, $post_id );
-		/* translators: %s: event title. */
-		$subject = sprintf( __( '📅 %s', 'gatherpress' ), get_the_title( $post_id ) );
-		$body    = Utility::render_template(
-			sprintf( '%s/includes/templates/admin/emails/event-email.php', GATHERPRESS_CORE_PATH ),
-			array(
-				'event_id' => $post_id,
-				'message'  => $message,
-			),
-		);
-		$headers = array( 'Content-Type: text/html; charset=UTF-8' );
-		$subject = stripslashes_deep( html_entity_decode( $subject, ENT_QUOTES, 'UTF-8' ) );
-
 		foreach ( $members as $member ) {
 			if ( '0' === get_user_meta( $member->ID, 'gatherpress_event_updates_opt_in', true ) ) {
 				continue;
@@ -468,8 +456,25 @@ class Rest_Api {
 
 			if ( $member->user_email ) {
 				$to = $member->user_email;
+				$switched_locale = switch_to_user_locale( $member->ID );
+
+				/* translators: %s: event title. */
+				$subject = sprintf( __( '📅 %s', 'gatherpress' ), get_the_title( $post_id ) );
+				$body    = Utility::render_template(
+					sprintf( '%s/includes/templates/admin/emails/event-email.php', GATHERPRESS_CORE_PATH ),
+					array(
+						'event_id' => $post_id,
+						'message'  => $message,
+					),
+				);
+				$headers = array( 'Content-Type: text/html; charset=UTF-8' );
+				$subject = stripslashes_deep( html_entity_decode( $subject, ENT_QUOTES, 'UTF-8' ) );
 
 				wp_mail( $to, $subject, $body, $headers );
+
+				if ( $switched_locale ) {
+					restore_previous_locale();
+				}
 			}
 		}
 
--- a/includes/core/classes/class-rest-api.php
+++ b/includes/core/classes/class-rest-api.php
@@ -448,15 +448,22 @@ class Rest_Api {
 			return false;
 		}
 
+		// Keep the currently logged-in user.
+		$current_user = wp_get_current_user();
+
 		$members = $this->get_members( $send, $post_id );
 		foreach ( $members as $member ) {
 			if ( '0' === get_user_meta( $member->ID, 'gatherpress_event_updates_opt_in', true ) ) {
 				continue;
 			}
-
 			if ( $member->user_email ) {
 				$to = $member->user_email;
 				$switched_locale = switch_to_user_locale( $member->ID );
+
+				// Set the current user to the actual member to mail to,
+				// to make sure the GatherPress filters for date- and time- format, as well as the users timezone,
+				// are recognized by the functions inside render_template().
+				wp_set_current_user( $member->ID );
 
 				/* translators: %s: event title. */
 				$subject = sprintf( __( '📅 %s', 'gatherpress' ), get_the_title( $post_id ) );
@@ -469,6 +476,9 @@ class Rest_Api {
 				);
 				$headers = array( 'Content-Type: text/html; charset=UTF-8' );
 				$subject = stripslashes_deep( html_entity_decode( $subject, ENT_QUOTES, 'UTF-8' ) );
+
+				// Reset the current user to the editor sending the email.
+				wp_set_current_user( $current_user->ID );
 
 				wp_mail( $to, $subject, $body, $headers );
 
--- a/includes/core/classes/class-setup.php
+++ b/includes/core/classes/class-setup.php
@@ -82,12 +82,12 @@ class Setup {
 	 * @return void
 	 */
 	protected function setup_hooks(): void {
-		register_activation_hook( GATHERPRESS_CORE_FILE, array( $this, 'activate_gatherpress_plugin' ) );
-		register_deactivation_hook( GATHERPRESS_CORE_FILE, array( $this, 'deactivate_gatherpress_plugin' ) );
+		// register_activation_hook( GATHERPRESS_CORE_FILE, array( $this, 'activate_gatherpress_plugin' ) );
+		// register_deactivation_hook( GATHERPRESS_CORE_FILE, array( $this, 'deactivate_gatherpress_plugin' ) );
 
 		add_action( 'init', array( $this, 'maybe_flush_rewrite_rules' ) );
 		add_action( 'admin_notices', array( $this, 'check_users_can_register' ) );
-		add_action( 'admin_init', array( $this, 'check_gatherpress_alpha' ) );
+		add_action( 'admin_notices', array( $this, 'check_gatherpress_alpha' ) );
 		add_action( 'wp_initialize_site', array( $this, 'on_site_create' ) );
 
 		add_filter( 'block_categories_all', array( $this, 'register_gatherpress_block_category' ) );
--- a/includes/core/classes/class-setup.php
+++ b/includes/core/classes/class-setup.php
@@ -82,8 +82,6 @@ class Setup {
 	 * @return void
 	 */
 	protected function setup_hooks(): void {
-		add_action( 'init', array( $this, 'load_gatherpress_textdomain' ), 0 );
-
 		register_activation_hook( GATHERPRESS_CORE_FILE, array( $this, 'activate_gatherpress_plugin' ) );
 		register_deactivation_hook( GATHERPRESS_CORE_FILE, array( $this, 'deactivate_gatherpress_plugin' ) );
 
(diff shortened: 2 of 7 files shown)


includes/core/classes/class-setup.php
```diff
@@ -111,6 +113,38 @@ protected function setup_hooks(): void {
...
+	 * for users how chosed another than the sites default language in their /profile.php.
+	 */
+	public function load_gatherpress_textdomain() {
+		// Keep as a fallback if translation files were not loaded from translate.wordpress.org ??
+		// load_plugin_textdomain( 'gatherpress', false, GATHERPRESS_DIR_NAME . '/languages' );
+	
+		// Or use it like that
+		// and DELETE the /languages folder,
+		// and "* Domain Path:       /languages" as well?
+		load_plugin_textdomain( 'gatherpress', false, false );
```
carstingaxion, line 145: Should we keep the `/languages` folder or remove it entirely @mauteri ?

mauteri: it should have been removed entirely since we are on .org repo now and no longer require it. thx!

carstingaxion:
Ooops ... deleted by accident; and restored.

carstingaxion:
While the original bug from #651 is already fixed at this point, I stumbled upon some strange time- and date formatting issue, after my last changes.

screenshot

That’s why the PR is still open and not yet finished.

carstingaxion:
Ok, found the problem:

In the gatherpress timeformat settings, `\` are ok and allowed.
grafik

While within the `profile.php`, they are not and get stripped off on save.
grafik

includes/core/classes/class-setup.php
```diff
@@ -82,6 +82,8 @@ protected function instantiate_classes(): void {
 	 * @return void
 	 */
 	protected function setup_hooks(): void {
+		add_action( 'init', array( $this, 'load_gatherpress_textdomain' ), 0 );
```
mauteri, line 85: Why `0`? I try not to be arbitrary with priorities and add a comment when the priority is not `10` so the next person knows the reason why the priority is the way it is. Usually, I try not to deviate far from `10` (likely to be `9` or `11` if it creates an issue with another priority that is the default `10`). If I want it to be the first or last thing run, I'll usually use `PHP_INT_MIN` or `PHP_INT_MAX`, but I try to not be too far off the thing that is keeping me from setting the priority to `10`.

includes/templates/admin/emails/event-email.php
```diff
@@ -20,13 +20,14 @@
...
-$gatherpress_venue = $gatherpress_event->get_venue_information()['name'];
+$gatherpress_event       = new Event( $event_id );
+$gatherpress_event_image = get_the_post_thumbnail_url( $event_id, 'full' );
+$gatherpress_venue       = $gatherpress_event->get_venue_information()['name'];
 
 ?>
 
 <!DOCTYPE html>
-<html lang="en-US">
+<html <?php language_attributes(); ?>>
```
mauteri, line 30: Good catch!

mauteri (changes requested):

includes/core/classes/class-setup.php
carstingaxion: Hmmm ... why did I do this here ... I need to have a look again.

Thanks for bringing this up! I often do the same and add comments about non-10 priorities for my future-self, but must have forgotten here.

includes/templates/admin/emails/event-email.php
carstingaxion: Thank you.

includes/core/classes/class-setup.php
```diff
@@ -111,6 +113,38 @@ protected function setup_hooks(): void {
...
+	 * you have to set the Requires at least: field in your readme.txt to 4.6 or more.
+	 * @see https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#plugins-on-wordpress-org
+	 *
+	 * BUT, ...!!!
+	 *
+	 * A call to load_plugin_textdomain() is ABSOLUTELY NEEDED to properly load all translation files
+	 * for users how chosed another than the sites default language in their /profile.php.
+	 */
+	public function load_gatherpress_textdomain() {
+		// Keep as a fallback if translation files were not loaded from translate.wordpress.org ??
```
carstingaxion: Remove old variant.

includes/core/classes/class-setup.php
```diff
@@ -111,6 +113,37 @@ protected function setup_hooks(): void {
...
+	 * If you don’t want to add a load_plugin_textdomain() call to your plugin
+	 * you have to set the Requires at least: field in your readme.txt to 4.6 or more.
+	 * @see https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#plugins-on-wordpress-org
+	 *
+	 * BUT, ...!!!
+	 *
+	 * A call to load_plugin_textdomain() is ABSOLUTELY NEEDED to properly load all translation files
+	 * for users who chosed another than the sites default language in their /profile.php.
+	 */
+	public function load_gatherpress_textdomain() {
```
swissspidy, line 143: You really don't need this. It's pointless and only has downsides. And saying "A call to load_plugin_textdomain() is ABSOLUTELY NEEDED" is purely **WRONG**.

The just-in-time translation loading in WordPress handles everything for you and only loads the translations when needed.

includes/core/classes/class-rest-api.php
```diff
@@ -448,28 +448,43 @@ public function send_emails( int $post_id, array $send, string $message ): bool
...
+				$to              = $member->user_email;
+				$switched_locale = switch_to_user_locale( $member->ID );
+
+				// Set the current user to the actual member to mail to,
+				// to make sure the GatherPress filters for date- and time- format, as well as the users timezone,
+				// are recognized by the functions inside render_template().
+				wp_set_current_user( $member->ID );
+
+				/* translators: %s: event title. */
+				$subject = sprintf( __( '📅 %s', 'gatherpress' ), get_the_title( $post_id ) );
```
swissspidy, line 469: This would benefit from additional context for translators (e.g. `_x( '📅 %s', 'email subject', 'gatherpress' )`).

Also worth noting that this is not really useful for translators. What would you even translate here? It's just a placeholder and an emoji.

includes/core/classes/class-setup.php
carstingaxion: Hmm... a little bit strange.

I absolutely trust you in translation stuff, because I know you have been advocating and coding a lot for that topic.

From my long doc-comment you can imagine that I read & tried a lot to come up with my solution and the UPPERCASE comment. But unfortunately I can not make this work, without that line of code. Or maybe I have the wrong expectations.

## How to test

0. Comment out `load_plugin_textdomain( 'gatherpress', false, false )`
1. Set site language to English in `wp-admin/options-general.php`
2. Set user prefered language to German in `wp-admin/profile.php`
3. Check that the Admin-UI is German now, except everything GatherPress 
4. Re-Enable `load_plugin_textdomain( 'gatherpress', false, false )`
5. Check that the Admin-UI is German now, including everything GatherPress

swissspidy: The issue is that you are triggering just-in-time translation loading **too early**.

WordPress tries to load the translations automatically as soon as you call `__()`. In your case, this happens somewhere before the current user is loaded, thus the admin won't be in German.

With this manual `load_plugin_textdomain()` call you're just papering over this issue, causing WP to try to load the translations twice.

The proper fix is to initialize your plugin later. This in the main plugin file works:

```
add_action(
	'plugins_loaded',
	static function () {
		// Initialize setups.
		GatherPress\Core\Setup::get_instance();
	}
);

// Or simply:

add_action(
	'plugins_loaded',
	array( '\GatherPress\Core\Setup', 'get_instance' )
);
```

carstingaxion: Ok, thank you very much Pascal, once again your explanation is better than any docs (I know) about that.

I followed your advise:

1. Set `add_action( 'plugins_loaded', array( '\GatherPress\Core\Setup', 'get_instance' ));`
2. Disabled my attempt `// add_action( 'init', array( $this, 'load_gatherpress_textdomain' ), 0 ); `
3. And started looking for the anoying bug again
4. and was able to track it down until `Assets->localize()`, which calls
   `'timezoneChoices'  => Utility::timezone_choices(),` and seems to be the culprit. Disabling this line brings all translations back into shape; having this enabled messes things up.
5. Going on ...

carstingaxion: Hello @mauteri ,

do we really need all possible timezones 

1. in the frontend markup?
2. in the backend markup?
3. on every page load?

https://github.com/carstingaxion/gatherpress/blob/02358ab60318a7464e31865ffea0f3bb8c961809/includes/core/classes/class-assets.php#L306

swissspidy: Isn't `Assets::localize()` only called in `Assets::add_global_object()` which is hooked to `wp_head`? So that should be late enough.

----

Aside: just adding a script tag like this is usually not a good idea. You should at least use `wp_print_inline_script_tag`. Even better, attach it to a specific script with `wp_add_inline_script`.

carstingaxion: https://github.com/user-attachments/assets/dd4685bc-dc0e-4cd2-911e-2d954a456e5e

carstingaxion: Found it: :tada: 

```php
$timezones_raw   = explode( PHP_EOL, wp_timezone_choice( 'UTC', get_user_locale() ) );  // i18n FIX!
```

carstingaxion: As expected, a change like to the load order will break things. At least the unit tests.

carstingaxion: > Isn't `Assets::localize()` only called in `Assets::add_global_object()` which is hooked to `wp_head`? So that should be late enough.

And on `admin_print_scripts`, which is the problematic trace.

> Aside: just adding a script tag like this is usually not a good idea. You should at least use `wp_print_inline_script_tag`. Even better, attach it to a specific script with `wp_add_inline_script`.

:+1:  I wanted to suggest the same, even that the relevant data - the timezones - are only used for the Event Date Block (#684)

includes/core/classes/class-rest-api.php
carstingaxion: I made an issue out of your idea: https://github.com/GatherPress/gatherpress/issues/827

and will add the context.

includes/core/classes/class-setup.php
carstingaxion: _This will be solved_, **when we accept turning the plugin upside-down and changing the load order.** 

OMG!

We need to have a :coffee: together @mauteri , I think.

carstingaxion:
This PR was split into multiple different PRs to make testing, understanding etc. easier.

PR #792, opened 2024-08-14
Downgrade dependency-review-action to version 4.3.3

carstingaxion:
### Description of the Change

Following the description at https://github.com/actions/dependency-review-action/issues/809#issuecomment-2285930024

> Version 4.3.4 does not work with invalid license names in the config and since that version,
SPDX expression licenses are considered invalid (see: https://github.com/actions/dependency-review-action/issues/792).
>
>In short, remove all SPDX expressions from the config!
Note that if you have packages that use SPDX expressions, this will report them as invalid,
as support is not yet implemented (see: https://github.com/actions/dependency-review-action/issues/263).
>
>In that case you need to downgrade to version 4.3.3 where SPDX expressions are handled as exact strings !!

~~Closes #~~ Hopefully fixes the [failing GitHub workflow run](https://github.com/GatherPress/gatherpress/actions/runs/10389051778/job/28776745529?pr=783) on invalid licenses, that aren't invalid in reality.

### Changelog Entry

> Changed - Downgrade version of dependency-review-action to prevent false-positives.

### Credits

Props @carstingaxion 

### Checklist:

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -17,8 +17,18 @@ jobs:
     steps:
       - name: 'Checkout Repository'
         uses: actions/checkout@v4
+
       - name: Dependency Review
-        uses: actions/dependency-review-action@v4
+        uses: actions/dependency-review-action@72eb03d02c7872a771aacd928f3123ac62ad6d3a # Version 4.3.3 # https://github.com/actions/dependency-review-action/releases/tag/v4.3.3
+        # Version 4.3.4 does not work with invalid license names in the config and since that version,
+        # SPDX expression licenses are considered invalid (see: https://github.com/actions/dependency-review-action/issues/792).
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-review.yml
```diff
@@ -17,8 +17,18 @@ jobs:
     steps:
       - name: 'Checkout Repository'
         uses: actions/checkout@v4
+
       - name: Dependency Review
-        uses: actions/dependency-review-action@v4
+        uses: actions/dependency-review-action@72eb03d02c7872a771aacd928f3123ac62ad6d3a # Version 4.3.3 # https://github.com/actions/dependency-review-action/releases/tag/v4.3.3
```
jeffpaul, line 22: Any reason you didn't change to `actions/dependency-review-action@v4.3.3`?

PR #805, opened 2024-08-16
Update to readme.md (change to video link and deletion of duplicate paragraphs)

patriciabt:
### Description of the Change
- the demo video link is now the Prettylink so we can change the target YouTube link when we have a new video.
- removed duplicate paragraphs (sections Contribute, What’s about the PRO version? Reminder that GatherPress is still in Alpha, upgrade notice)
- added under installation: from the WP plugin repository
- added the "The Map provider" and the "Not attending option" in the "Configure GatherPress" paragraph

Closes #

### Checklist:


readme.md
```diff
@@ -80,6 +80,11 @@ This plugin leverages the following third-party libraries for various functional
 
 ## Installation
 
+### From the WordPress.org plugin repository
+1. Install it in your WordPress instance: go to WP Admin Plugins > Add new plugin. Search for `gatherpress`. Once found, click `Install`.
+2. Activate the plugin.
+
+### From the file downloaded from our GitHub repository
 1. Download the plugin: you'll find the latest release on the GatherPress GitHub repository, under [Releases](https://github.com/GatherPress/gatherpress/releases) > Assets and download `gatherpress.zip`.
```
carstingaxion, line 88: Linking to https://github.com/GatherPress/gatherpress/releases/latest will redirect to the latest tagged release.

readme.md
```diff
@@ -186,28 +189,3 @@ Yes, GatherPress can be run on a network of sites. The additional database table
...
-
-### Reminder that GatherPress is still in Alpha
-
-As we continue to refine and develop the plugin, please use the [GatherPress Alpha](https://github.com/GatherPress/gatherpress-alpha) plugin alongside the core GatherPress plugin. The Alpha plugin manages breaking changes easily: just make sure it is up-to-date, activate it, go to the Alpha section under GatherPress Settings, and click "Fix GatherPress!" after updating GatherPress. This process helps us avoid technical debt as we work towards launching version 1.0.0 of the plugin.
-
-## Changelog
-
-See complete changelog at https://github.com/GatherPress/gatherpress/releases.
-
-## Upgrade Notice
```
carstingaxion, line 209: I would recommend removing this section entirely.

This upgrade notice will get rendered into the wp-admin/plugins.php only if the version here matches the version to update to. This can be helpful if a plugin would introduce breaking changes and could be re-enabled and used when GatherPress reaches a 1.0.0

carstingaxion (changes requested):

readme.md
carstingaxion: … sorry, taping the correct lines on my tablet wasn’t as intuitive as I thought. But you’ll get it.

carstingaxion:
Maybe we could something like the following? Open for improvements.

---

Contributing to the project can be fun & doesn't need code at all thanks to Playground. 

Help shaping GatherPress by adding your least recent & your next upcoming WordCamp, you attend(ed), to our set of demo-data powered by Playground. 

Take a look at the recent development and test actual PRs with Playground, just click on one of the preview-links in each PR and add your feedback.

patriciabt:
@carstingaxion we need the links then for the Playground

Yes to remove the mention of 1.0.0 but I think it's good to have a link to the changelog

carstingaxion:
Updated my comment with one link and will provide the Playground link soon (just realised a minor thing in the URL I wanted to copy).

Nice, that you agree on removing _the 1.0.0_ .

Absolutely, a link to the changelog is mandatory, I didn't recognise it was already there. Sorry.

carstingaxion:
I  added the Playground link as well @patriciabt

readme.md
```diff
@@ -29,7 +29,7 @@ This project is for the collaborative effort to build a compelling event managem
 
 We're creating the very network features we need to host events and gather well.
 
-https://www.youtube.com/watch?v=BnYS36C5d38&t=2s
+https://gatherpress.org/demovideo
```
carstingaxion, line 32: This will not render as a video on w.org 

We might have to change this regularly ... I think.

carstingaxion (changes requested):

readme.md
patriciabt: Are you sure? For example when pasting such a link on social media, the Youtube embed is rendered

The problem is that I need to wait for the release to make the video, if I want to avoid the version to show -alpha (or it's okay if it shows?), but then we need the new video link every time in the readme at the time of submitting the new version to .org
(or there is a way to have a youtube link in a draft before to upload a video? I'll check that)

carstingaxion:
I guess this can be closed as fixed, long time ago.

PR #808, opened 2024-08-16
Hide Admin Notice about alpha-plugin, also if installed via the branch-named zip

### LINKED ISSUE #807: seeing The GatherPress Alpha plugin notice
### Describe the bug

I am getting a notice in wp-admin  and wp-admin/network saying:

The GatherPress Alpha plugin is not installed or activated. This plugin is currently in heavy development and requires GatherPress Alpha to handle breaking changes. Please [download and install GatherPress Alpha](https://github.com/GatherPress/gatherpress-alpha) to ensure compatibility and avoid issues.

But as per the screenshot below both are activated (network activated infact). Maybe this is because I am using multisite. If so this needs to be corrected

Screenshot from 2024-08-16 19-39-12

### Steps to Reproduce

Network activate both GatherPress and GatherPress Alpha Version 0.30.0 on  multisite

Screenshot from 2024-08-16 19-39-12

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

carstingaxion:
While a plugin from w.org would have a unigue slug to check against,
this is different for .zips downloaded from github.com.

A download via "<> Code ⏷" > "Download ZIP" will include the branch in the file name,
while a download via "Releases" will provide a zip without a branch name attached.

### Description of the Change

This PR makes sure, that GatherPress checks against both possible slugs inside the condition to whether or not to show the admin notice about the alpha-plugin.

Closes #807

### Changelog Entry

> Fixed - Hide Admin Notice about alpha-plugin, also if installed via the branch-named zip

### Credits

Props @carstingaxion 

### Checklist:

### DIFF
--- a/includes/core/classes/class-setup.php
+++ b/includes/core/classes/class-setup.php
@@ -432,7 +432,7 @@ class Setup {
 	 */
 	public function check_gatherpress_alpha(): void {
 		if (
-			is_plugin_active( 'gatherpress-alpha/gatherpress-alpha.php' ) ||
+			is_plugin_active( 'gatherpress-alpha-main/gatherpress-alpha.php' ) ||
 			filter_var( ! current_user_can( 'install_plugins' ), FILTER_VALIDATE_BOOLEAN ) || (
 				false === strpos( get_current_screen()->id, 'plugins' ) &&
 				false === strpos( get_current_screen()->id, 'plugin-install' ) &&
--- a/includes/core/classes/class-setup.php
+++ b/includes/core/classes/class-setup.php
@@ -432,12 +432,7 @@ class Setup {
 	 */
 	public function check_gatherpress_alpha(): void {
 		if (
-			// While a plugin from w.org would have a unigue slug to check against,
-			// it different for .zips downloaded from github.com.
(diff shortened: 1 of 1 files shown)


includes/core/classes/class-setup.php
```diff
@@ -432,6 +432,11 @@ public function check_users_can_register(): void {
 	 */
 	public function check_gatherpress_alpha(): void {
 		if (
+			// While a plugin from w.org would have a unigue slug to check against,
+			// it different for .zips downloaded from github.com.
+			// A download via "<> Code ⏷" > "Download ZIP" will include the branch in the file name.
+			is_plugin_active( 'gatherpress-alpha-main/gatherpress-alpha.php' ) ||
+			// A download via "Releases" will provide a zip without a branch name attached.
```
mauteri, line 439: Might be better to change this to see if the GatherPress Alpha version constant is defined, then we don't have to worry about file name.

carstingaxion: Good idea.

carstingaxion:
Friendly ping @mauteri ;)

mauteri (approved):

PR #810, opened 2024-08-17
Feature/e2e tests workflow wp now

carstingaxion:
### Description of the Change

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:


carstingaxion:
This can’t work, …. because ….

> @adamziel: I see you're using the `importFile` step which is, at the moment, browser-only unfortunately. It uses the `DOMParser` class that's only available in the browser. I'm not sure if that already works in `wp-now`, but you could try importing that file via the [wpCLI](https://wordpress.github.io/wordpress-playground/blueprints-api/steps/#WPCLIStep) step and [the import command](https://github.com/wp-cli/import-command). Perhaps that step should do just that internally when ran in Node.js.
>
> https://github.com/WordPress/playground-tools/issues/195#issuecomment-2016209820

carstingaxion:
image
 
**webkit** is behaving strangly different.

carstingaxion:
Update docs with

- https://github.com/WordPress/wordpress-playground/pull/1289
- https://github.com/WordPress/gutenberg/pull/62692
- https://github.com/swissspidy/wp-performance-action/blob/try/wp-playground/action.yml

carstingaxion:
Some tests are still failing and I'm not sure, what is the reason for that.

But, as the quest was to find a working environment, I would call this to be done and open it up for reviews.

carstingaxion:
> [!NOTE]
> Ready for review does NOT mean ready for merge!

Just to note.

carstingaxion:
grafik
grafik

adamziel:
If the database works for a while and then it breaks, It's likely this issue: https://github.com/WordPress/wordpress-playground/issues/1596. I'd love to learn what happens there. Would you be able to log the first failing request and zip the playground site? It would also be supremely useful to peek into the `/internal` directory available through the PHP instance running inside the Playground CLI, but there's no easy API to do it from bash – we'd need a custom build that would zip it up and write to a local directory on the first error.

cc @wunderbart @swissspidy as these database error came up in your explorations, too.

test/e2e/tests/editor/blocks.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+		// Check if the inserter button is visible
+		await expect(inserter).toBeVisible();
+
+		// Ensure the inserter button is enabled before interacting with it
+		await expect(inserter).toBeEnabled();
+
+		// Click the inserter button
+		// Just click() without forcing led to timeouts, timeouts, timeouts...
+		await inserter.click({ force: true });
```
WunderBart, line 32: This combo is actually what `locator.click()` is already doing underneath. Before the actual click, Playwright performs a series of actionability checks (auto-waiting), including visibility and enabled state ([source](https://playwright.dev/docs/actionability#:~:text=%2D-,locator.click(),Yes,-%2D)). Therefore, this entire block is equivalent to:

```suggestion
		await inserter.click({ timeout: 10000 }); // Increase timeout if necessary.
```

test/e2e/tests/editor/events.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+test.describe('Events in the Editor', () => {
+	test.beforeEach(async ({ admin, page }) => {
+		await admin.createNewPost({ postType: 'gatherpress_event' });
+		await page.getByLabel('Add title').fill('Change title to allow saving');
+		await page
+			.getByLabel('Block: Event Date')
+			.locator('div')
+			.first()
+			.isVisible();
+		await page.getByRole('heading', { name: 'Date & time' }).isVisible();
```
WunderBart, line 15: This will have no impact, as the `locator.isVisible()` utility only checks whether an element is visible, returning a boolean value. To assert visibility, the `await expect(locator).toBeVisible()` assertion needs to be used.

test/e2e/tests/editor/events.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+			// Check if the inserter button is visible
+			await expect(panelToggle).toBeVisible();
+
+			// Ensure the panelToggle is enabled before interacting with it
+			await expect(panelToggle).toBeEnabled();
+
+			// Click the panelToggle
+			// Just click() without forcing led to timeouts, timeouts, timeouts...
+			await panelToggle.click({ force: true });
+		}
```
WunderBart, line 32: This block is equivalent to:

```suggestion
		// Open the Document -> Event settings panel.
		const panelToggle = page.getByRole('button', {
			name: 'Event settings',
			expanded: false,
		});

		if (await panelToggle.isVisible()) {
			await panelToggle.click();
		}
```

test/e2e/tests/editor/venues.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+	test('The admin should be able to create a new Venue.', async ({
+		editor,
+		page,
+	}) => {
+		await page.getByLabel('Add title').isVisible();
+		await page.getByLabel('Add title').fill('Test venue');
+		await page.getByLabel('Add title').press('Tab');
+
+		const venue = await page.$('.gatherpress-venue__name');
```
WunderBart, line 19: The `$` method is discouraged. Locator can be used instead:

```suggestion
		const venue = page.locator('.gatherpress-venue__name');
```

test/e2e/tests/frontend/rsvp.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+	test('A logged in user can perform RSVP action', async ({ page }) => {
+		await page.goto('/');
+		await page.getByRole('link', { name: 'Events' }).click({ force: true }); // Imported from of https://github.com/GatherPress/gatherpress-demo-data
+
+		await page.evaluate(() => window.scrollTo(0, 5000));
+		await page
+			.getByRole('link', { name: 'RSVP' })
+			.first()
+			.click({ force: true, timeout: 60000 });
```
WunderBart, line 51: Before the `click` is performed, [Playwright will try scrolling the element into view](https://playwright.dev/docs/input#mouse-click:~:text=scroll%20the%20element%20into%20view). Is the forced scroll necessary here?

test/e2e/tests/frontend/rsvp.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+		await page
+			.locator('a')
+			.filter({ hasText: 'Attend' })
+			.click({ force: true });
+		await page.getByText('Close').click({ force: true });
```
WunderBart, line 57: Forcing clicks should be used only in [specific edge cases](https://playwright.dev/docs/input#forcing-the-click:~:text=Sometimes%2C%20apps%20use%20non%2Dtrivial%20logic%20where%20hovering%20the%20element%20overlays%20it%20with%20another%20element%20that%20intercepts%20the%20click.). Do we know if it's necessary to use it here? I'm asking because it bypasses the auto-waiting mechanism which can lead to flaky results.

test/e2e/tests/wp-admin/settings.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+		});
+		const settingsMenu = menu.getByRole('link', {
+			name: 'Settings',
+		});
+		const settingsMenuUrl = await settingsMenu.getAttribute('href');
+
+		await expect(settingsMenu).toBeVisible();
+		await expect(settingsMenuUrl).toContain(
+			'admin.php?page=gatherpress_general'
+		);
```
WunderBart, line 24: The URL attribute should be asserted via the [`toHaveAttribute`](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-have-attribute) locator assertion so that it's properly awaited:

```suggestion
		const settingsMenu = menu.getByRole('link', {
			name: 'Settings',
		});

		await expect(settingsMenu).toBeVisible();
		await expect(settingsMenu).toHaveAttribute(
			'href',
			/admin\.php\?page=gatherpress_general/
		);
```

carstingaxion:
Wow!
Thank you very much for your detailed review @WunderBart !

Given that we never ever had any contact and that I was working just yesterday evening on this stuck PR, I do call this a great community moment!

I absolutely appreciate you flying by and taking the time, thank you very much!

WunderBart:
> Wow! Thank you very much for your detailed review @WunderBart !
> 
> Given that we never ever had any contact and that I was working just yesterday evening on this stuck PR, I do call this a great community moment!
> 
> I absolutely appreciate you flying by and taking the time, thank you very much!

Happy to help! I did spend a lot of time with Playwright so please don't hesitate to ping me anytime! 🙌 😄 


BTW, I noticed a lot of increased timeouts for the actions (mostly clicks) in this PR. Is it because the app is responding slowly? Maybe we could come up with some pre-test solution for that, or configure (action) timeouts globally.

test/e2e/tests/frontend/rsvp.spec.ts
carstingaxion: Hello again @WunderBart, thanks for looking at this!

Yes, I'm aware of this bad behavior, as you can see from some of the inline comments and my commit messages like [TRYOUT .click({ force: true }) which is ugly, but works :(](https://github.com/GatherPress/gatherpress/pull/810/commits/e405a0f8af711447f0142072e33eed35db56a470).

I experienced, that using `force:true` helped avoiding the before mentioned timeouts (error 500) to a minimum or even to not happen at all. With the commits from yesterday, I was a least able to reduce the failing tests to one, followed by an error 500. But it was only one 500 anymore, not 15-20 like in the workflow runs before.

But yes - all in all - I would like to remove all forcing too, if I (or maybe we) find a way around the fatal errors.

test/e2e/tests/editor/blocks.spec.ts
carstingaxion: I guess, something like this were my starting point.
Afterwards I added the pieces one by one to find out, what helps against the constantly happening timeouts.

Following your suggestion, I'm going to remove this superflous code now and am curious about the results.

test/e2e/tests/editor/events.spec.ts
carstingaxion: Cool, thats a lot more readable.
Thank you!

test/e2e/tests/editor/venues.spec.ts
carstingaxion: Thanks again; this is not only code-cleanup; this is pure education. Thx!

test/e2e/tests/wp-admin/settings.spec.ts
carstingaxion: Looks much better, thanks.

mauteri:
Instead of playground, I had a lot more success running the E2E tests with `wp-env`. https://github.com/GatherPress/gatherpress/pull/1173

Going to close this now that the environment issue is resolved and we can focus on writing E2E tests that run successfully locally and in the CI.

PR #820, opened 2024-08-27
Event date block and functionality refactor

### LINKED ISSUE #803: Save event start and end datetimes and timezone as meta

Currently, GatherPress stores datetimes and timezone to a custom table which is used to query events in a scalable way. This change is to ALSO store this data as post meta so it is more easily accessible to developers and in the API. We can use the same approach of pseudo post meta illustrated in this ticket https://github.com/GatherPress/gatherpress/issues/560

### Designs

_No response_

_No response_

### Code of Conduct

mauteri:
### Description of the Change
This change updates the Event Date block to look and feel more like the WordPress core date block in terms of styling. In this PR we also introduce meta data that is available in core post REST API for datetime start, end, and timezone. We also use meta data hook to update the custom event table for datetimes. This means we no longer need the datetime REST endpoint which was a bit clunky and required an additional call. That endpoint has now been removed. In addition, WP core stores powered by redux were introduced to start replacing the Broadcaster/Listener code that was custom written. This works a LOT better, and soon more things will get stores to replace Broadcaster/Listener functionality.

Closes #803 #684

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix

### Credits

Props @mauteri

### Checklist:


src/components/DateTimeRange.js
```diff
@@ -31,7 +31,7 @@ const DateTimeRange = () => {
 	const [dateTimeEnd, setDateTimeEnd] = useState();
 	const [timezone, setTimezone] = useState();
 
-	subscribe(saveDateTime);
+	saveDateTime();
```
mauteri, line 34: The subscribe calls the endpoint a LOT of times. Firing it just once here fires only once. This will all be looked at when the Event Date block is worked on. I'm planning on doing that next.

includes/core/classes/class-event.php
```diff
@@ -685,11 +685,34 @@ function ( $key ) {
...
+					$meta_key,
+					sanitize_text_field( $field )
+				);
+			} else {
+				add_post_meta(
+					$fields['post_id'],
+					$meta_key,
+					sanitize_text_field( $field )
+				);
+			}
```
mauteri, line 713: Even though this was set to `'single' => true` it was duplicating meta. Doing an exist check worked, but I don't love this solution. I would think `update_post_meta` should just work without duplicating...

carstingaxion:
@mauteri Do not miss to checkout the work at https://github.com/WordPress/gutenberg/issues/60971, which seems closely related to GatherPress' needs for a close-to-core event date block.

mauteri:
> @mauteri Do not miss to checkout the work at [WordPress/gutenberg#60971](https://github.com/WordPress/gutenberg/issues/60971), which seems closely related to GatherPress' needs for a close-to-core event date block.

Yup, saw this but it's still in the design phase. If this makes it into core we can definitely swap out the picker.

bluecollarcoders (approved):

jmarx (approved):

PR #822, opened 2024-08-27
Migrate screenshot generator workflow to use wp-playground/cli

carstingaxion:
### Description of the Change

Use WordPress Playground in favor of `wp-env` as the environment for the screenshot generator.

Switching to Playground not only makes this workflow **a lot faster**, it also allows to use a `blueprint.json` to setup the environment, which is way more comfortable, than writing a long list of `wp-cli` commands to do the same.

#### Needed before merge

- [ ] Finish the docs
- [x] https://github.com/WordPress/wordpress-playground/issues/1703

Included in this PR is also:

- the subtle change to use the _modern_ Admin-UI,
   which should become the default across all GatherPress Playgrounds
- hiding the admin-notices for this presentation-purpose, like @patriciabt suggested
- fixing the left sided menu background to fill the full height of the screen
 

grafik

~~Closes #~~

### Changelog Entry

> Changed - Run the screenshot generator workflow on WordPress Playground

### Credits

Props @carstingaxion 

### Checklist:


carstingaxion:
```bash

# Define the new data as a variable
new_data='[
  {"name": "item3", "value": 30},
  {"name": "item4", "value": 40}
]'

# Use jq to append the new data to the existing JSON file
jq --argjson newData "$new_data" '.items += $newData' data.json > updated_data.json

# Optionally, overwrite the original file
mv updated_data.json data.json

```

carstingaxion:
I splitted this into #837 and #838 for clarity.

PR #862, opened 2024-09-05
PATCH RELEASE 0.30.1 — Allow to create nice screenshots from 0.30.0

carstingaxion:
### Description of the Change

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:

### DIFF
--- a/includes/core/classes/class-setup.php
+++ b/includes/core/classes/class-setup.php
@@ -432,7 +432,7 @@ class Setup {
 	 */
 	public function check_gatherpress_alpha(): void {
 		if (
-			is_plugin_active( 'gatherpress-alpha/gatherpress-alpha.php' ) ||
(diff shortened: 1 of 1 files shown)


mauteri:
Not sure I agree on a release just for this. It means we need to package up a release for both the plugin and GatherPress Alpha, which will then make the user have to update GatherPress Alpha on their site as well. If we have any other small release we want to make I would include this, but I wouldn't make 0.30.1 just this.

carstingaxion:
Yes, I haven’t seen the full picture, which is slightly more complicated.

Even I’m unsure if I like this tight coupling, I’m ok with waiting or even closing this PR.

PR #877, opened 2024-09-17
DRAFT: (solves 736) Checkbox to toggle latitude and longitude in Venue block

stephenerdelyi:
### Description of the Change

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:


carstingaxion:
I tested your work the first time, and it showed the UI nicely.

Unfortunately, the current set of demo-data has no lat lng data. Wasn’t there an idea in the past to request missing data on the fly to help with backwards compatibility?

Next to that, I would prefer to have this fields in the Venue Settings, rather than in the map settings, as the data belongs to the venue, not the kind of map rendering.

src/blocks/venue/edit.js
```diff
@@ -212,6 +223,45 @@ const Edit = ({ attributes, setAttributes, isSelected }) => {
...
+								}
+								checked={mapCustomLatLng}
+								onChange={(value) => {
+									setAttributes({ mapCustomLatLng: value });
+								}}
+							/>
+						</PanelRow>
+						{mapCustomLatLng && (
+							<>
+								<TextControl
```
mauteri: Let's make this a `NumberControl`

`import { __experimentalNumberControl as NumberControl } from '@wordpress/components';`

src/blocks/venue/edit.js
```diff
@@ -212,6 +223,45 @@ const Edit = ({ attributes, setAttributes, isSelected }) => {
...
+									label={__('Latitude', 'gatherpress')}
+									value={latitude}
+									onChange={(value) => {
+										Broadcaster({ setLatitude: value });
+										updateVenueMeta({ latitude: value });
+									}}
+								/>
+								<TextControl
```
mauteri: `NumberControl` here too. If you leave this as `TextControl`, hitting a non-number crashes the block.

src/blocks/venue/edit.js
```diff
@@ -212,6 +223,45 @@ const Edit = ({ attributes, setAttributes, isSelected }) => {
...
+									label={__('Longitude', 'gatherpress')}
+									value={longitude}
+									onChange={(value) => {
+										Broadcaster({ setLongitude: value });
+										updateVenueMeta({ longitude: value });
+									}}
+								/>
```
mauteri, line 263: We should have a button that will "Reset" the coordinates. Just call the button "Reset" and have it take the address and run it through the API again to bring back original values.

src/blocks/venue/edit.js
```diff
@@ -23,7 +25,7 @@ import { isVenuePostType } from '../../helpers/venue';
 import VenueSelector from '../../components/VenueSelector';
 import VenueInformation from '../../panels/venue-settings/venue-information';
 import OnlineEventLink from '../../components/OnlineEventLink';
-import { Listener } from '../../helpers/broadcasting';
+import { Broadcaster, Listener } from '../../helpers/broadcasting';
```
mauteri, line 28: If you'd like to look at the `store` I created for updates to the `Event Date` block, you can start using that instead of the `Listener / Broadcaster`. If not, we can update to that later.

mauteri (changes requested):

src/blocks/venue/edit.js
mauteri: Seems that the changes in lat/long are not being retained. I did some updates, but I don't think anything that should have effected functionality. It changes, but then when I leave focus, it reverts... It might be good to use the redux stores rather than Broadcaster/Listener as that may be the issue.

mauteri (changes requested):

jmarx:
@mauteri We are pretty close on this. Everything works well with OSM. But we just need to make another adjustment for Google Maps. But feel free to take another pass at this and check out our implementation of the redux store. Please let me know if anything needs to be changes with that.

Also,
I personally don't think we need the `reset` button. If you test this pr, you'll find that the lat/long toggle(use default/use custom) will take care of all that.

src/blocks/venue/edit.js
jmarx: > Seems that the changes in lat/long are not being retained. I did some updates, but I don't think anything that should have effected functionality. It changes, but then when I leave focus, it reverts... It might be good to use the redux stores rather than Broadcaster/Listener as that may be the issue.

We implemented the store. So all these issues should be resolved. @mauteri

src/blocks/venue/edit.js
jmarx: > If you'd like to look at the `store` I created for updates to the `Event Date` block, you can start using that instead of the `Listener / Broadcaster`. If not, we can update to that later.

@mauteri Done.

src/blocks/venue/edit.js
mauteri: @jmarx I would like to add the Reset button. I get what you are saying, but that might not be super obvious to people. A reset button would be.

stephenerdelyi:
@mauteri @jmarx this is ready for final review. Got the typo / JS formatting / tests updated and cleaned things up.

mauteri (approved):

PR #882, opened 2024-09-17
Draft: Work in progress adding filter to customize Default DurationOptions.

ajskelton:
The filter works, but something in the DateTimeRange useEffect is reverting back to the default values.

### Description of the Change

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:


src/helpers/datetime.js
```diff
@@ -56,43 +57,6 @@ export const defaultDateTimeEnd = moment
...
- * value representing the duration in hours. The last option allows the user
- * to set a custom end time by selecting `false`.
- *
- * @since 1.0.0
- *
- * @type {Array<Object>} durationOptions
- * @property {string}         label - The human-readable label for the duration option.
- * @property {number|boolean} value - The value representing the duration in hours, or `false` if a custom end time is to be set.
- */
-export const durationOptions = [
```
mauteri, line 73: let's keep the name `durationOptions`, but convert this to a function. Then apply your filter to the function before returning.

src/components/Duration.js
```diff
@@ -35,11 +36,11 @@ const Duration = () => {
 		<SelectControl
 			label={__('Duration', 'gatherpress')}
 			value={
-				durationOptions.some((option) => option.value === duration)
+				DurationOptions.some((option) => option.value === duration)
```
mauteri, line 39: this will become `durationOpions().some...` after changes below.

src/components/Duration.js
```diff
@@ -35,11 +36,11 @@ const Duration = () => {
...
 					? duration
 					: false
 			}
-			options={durationOptions}
+			options={DurationOptions}
```
mauteri, line 43: `options={durationOptions()}`

mauteri (changes requested):
Thanks for contributing, Anthony! Take a look at my comments and let me know what you think.

ajskelton:
The comments made a lot of sense and I pushed up a new commit with the updates. With those in place the filter from a theme was working great, but there is one bug that I was able to drill down.

When the page loads the `gatherpress/datetime` store builds the DEFAULT_STATE. The duration value uses the `getDateTimeOffset()` function that calculates what the duration is. During this first instance the durationOptions() function does not use the filtered values from the theme, it still uses the defaults. When it doesn't find the corresponding value in the object returned from durationOptions it returns false. With the state set to false it doesn't render the dropdown and instead shows the Date and Time End field so you can't change the duration anymore. 

Do you think it would be better to save the duration value to the post meta along with the start and end dates so this calculation doesn't need to happen?

mauteri:
> The comments made a lot of sense and I pushed up a new commit with the updates. With those in place the filter from a theme was working great, but there is one bug that I was able to drill down.
> 
> When the page loads the `gatherpress/datetime` store builds the DEFAULT_STATE. The duration value uses the `getDateTimeOffset()` function that calculates what the duration is. During this first instance the durationOptions() function does not use the filtered values from the theme, it still uses the defaults. When it doesn't find the corresponding value in the object returned from durationOptions it returns false. With the state set to false it doesn't render the dropdown and instead shows the Date and Time End field so you can't change the duration anymore.
> 
> Do you think it would be better to save the duration value to the post meta along with the start and end dates so this calculation doesn't need to happen?

Hey @ajskelton, can you try to call this function to get the duration value? This should serve the same purpose you are looking for with storing meta. Basically, it takes the start/end datetime and returns the duration or false (same as you'd get from the meta) https://github.com/GatherPress/gatherpress/blob/main/src/helpers/datetime.js#L127 Let me know if that works. Thx!

ajskelton:
@mauteri This is the function that doesn't use the filtered values when the default state is being set.

I tested by adding some console.logs in the .find(). When you refresh the page and select the block you see the default duration values logging to the console instead of the filtered duration. 

So it sets the duration to false before checking against the filtered options. 

When it runs here: https://github.com/GatherPress/gatherpress/blob/4b894749193ca340d3cea082e1cac8ef5c6a0e4f/src/stores/datetime.js#L23, it must be running before the filter in the theme is registered.

mauteri:
> @mauteri This is the function that doesn't use the filtered values when the default state is being set.
> 
> I tested by adding some console.logs in the .find(). When you refresh the page and select the block you see the default duration values logging to the console instead of the filtered duration.
> 
> So it sets the duration to false before checking against the filtered options.
> 
> When it runs here:
> 
> https://github.com/GatherPress/gatherpress/blob/4b894749193ca340d3cea082e1cac8ef5c6a0e4f/src/stores/datetime.js#L23
> 
> , it must be running before the filter in the theme is registered.

Oh gotcha, let me give it some thought.

carstingaxion:
Hello @ajskelton & @mauteri,

are you aware of https://github.com/WordPress/gutenberg/issues/60971, which seems slightly related?

ajskelton:
@mauteri I found a fix for the issue. Instead of running the getDateTimeOffset() function in the DEFAULT_STATE, i changed that initial duration value to be null. Then I moved that function to the getDuration selector. When the Duration Component renders it always runs the getDuration selector to dynamically get the value based on the start and end times. And when this runs all of the theme filters have already been registered. 

We do need to check if the current state is false here in the selector. I first tried just using the getDateTimeOffset but when the value was set to false the state wasn't saving the false value and the DateTimeEnd was never rendering. 

I don't have a ton of experience working with custom stores so let me know if this is the wrong way to go about this.

@carstingaxion That new Date/Time Range component does look nice. I think it's still just design right now, I couldn't see any code commits to check out.

src/stores/datetime.js
```diff
@@ -78,7 +78,7 @@ const store = createReduxStore('gatherpress/datetime', {
 	selectors: {
 		getDateTimeStart: (state) => state.dateTimeStart,
 		getDateTimeEnd: (state) => state.dateTimeEnd,
-		getDuration: (state) => state.duration,
+		getDuration: (state) => state.duration === false ? false : getDateTimeOffset(),
```
mauteri, line 81:
```
getDuration: (state) => false === state.duration ? false : getDateTimeOffset(),
```

mauteri (changes requested):

mauteri:
This can be tested with this gist. https://gist.github.com/mauteri/5b8402d29de94502fff1e0da9e68f2d8

PR #884, opened 2024-09-18
Run on plugins_loaded to help fix PR #752

carstingaxion:
### Description of the Change

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:

### DIFF
--- a/gatherpress.php
+++ b/gatherpress.php
@@ -41,4 +41,7 @@ require_once GATHERPRESS_CORE_PATH . '/includes/core/classes/class-autoloader.ph
 GatherPress\Core\Autoloader::register();
 
 // Initialize setups.
-GatherPress\Core\Setup::get_instance();
+add_action(
+	'plugins_loaded',
+	array( '\GatherPress\Core\Setup', 'get_instance' )
+);


carstingaxion:
@mauteri This is the bare-bones testing PR, to help with the new load order, we talked about.

I tested locally, that when I also add the changes from https://github.com/GatherPress/gatherpress/pull/752/commits/3c2230a19163dcf0770d3bc65d0290f24e4e016f to that branch, the failing tests go down from 21 to 6.

But those 6 are the same as in #752

mauteri:
@carstingaxion I thought about this more and I think the issue is one of the hooks in one of the singletons being at wrong priority and maybe that is what we need to explore and not wrapping everything in `plugins_loaded`. All code in those singletons are executed in a hook, even class-setup.php. That class instantiates and runs `setup_hooks`. It also instantiates the other singletons and they do the same thing; they run their `setup_hooks`. So the issue is in one of those hooks and that would be the fix rather than executing ALL hooks in `plugins_loaded`. 

I can look at your previous PR and we can discuss next week and I can help you identify the issue you are running into and which hook is responsible. Thx!

carstingaxion:
> @carstingaxion I thought about this more and I think the issue is one of the hooks in one of the singletons being at wrong priority and maybe that is what we need to explore and not wrapping everything in `plugins_loaded`. All code in those singletons are executed in a hook, even class-setup.php. That class instantiates and runs `setup_hooks`. It also instantiates the other singletons and they do the same thing; they run their `setup_hooks`. 

Thank you for the explanation. In general this is not new to me and this structure is one of the reasons I like GatherPress so much - its easy to understand & read. I like it.

>So the issue is in one of those hooks and that would be the fix rather than executing ALL hooks in `plugins_loaded`.
>
> I can look at your previous PR and we can discuss next week and I can help you identify the issue you are running into and which hook is responsible. Thx!

Yes, I guess this will be the path and I think we might start at https://github.com/GatherPress/gatherpress/pull/752#discussion_r1745490197:

>     4. and was able to track it down until `Assets->localize()`, which calls
>        `'timezoneChoices'  => Utility::timezone_choices(),` and seems to be the culprit. Disabling this line brings all translations back into shape; having this enabled messes things up.

carstingaxion:
As things are not complicated enough,  I believe that I may have tried to prevent or workaround this 8 year old problem  (without knowing about) https://core.trac.wordpress.org/ticket/38643 reported by ... _checks notes_ ... @swissspidy 

Sometimes it's wild.

carstingaxion:
I was not able to reproduce the former bug, so closing this as not needed anymore.

_Thats annoying, when you've spent plenty of time .... for nothing._

PR #892, opened 2024-09-19
Use latest demo-data event as prominent example in screenshots

carstingaxion:
I wanted to do something _nice_, _something visually appealing_.
... so I updated the screenshot generator to use the latest demo-data event as prominent example in screenshots ;)

https://github.com/user-attachments/assets/f4e7ce3e-2944-45b6-abc8-858d9254117d

### Description of the Change

~~Closes #~~

### Changelog Entry

> Changed - Use latest demo-data event as prominent example in screenshots

### Credits

Props @carstingaxion 

### Checklist:

- [x] I have added tests to cover my change. ;)

### DIFF
--- a/.github/scripts/wordpress-org-screenshots/wporg.spec.ts
+++ b/.github/scripts/wordpress-org-screenshots/wporg.spec.ts
@@ -17,7 +17,7 @@ test.describe( 'Screenshots for the wordpress.org/plugins repository', () => {
             local_code,
             '.png'
         ].join('').toLowerCase();
-    }	// List all comments.
+    }
 
 
     test.beforeAll( async ( { requestUtils } ) => {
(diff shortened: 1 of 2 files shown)


mauteri:
@carstingaxion nice, this good to merge?

mauteri (approved):

carstingaxion:
> @carstingaxion nice, this good to merge?

@mauteri Thanks for looking at this. If ready or not is a little bit a chicken or egg question, if you now this saying.

Simply: No it’s not, because it will reference non-existent images from the main readme.md

The crux: It would be nice to have the screenshots without the alpha-notice. This will be possible with the next release. Because the screenshot-generator uses the latest release from .org, we should wait until the next is out. This leads to the need for a patch release for only screenshots and readme.

I don’t know.

mauteri:
> > @carstingaxion nice, this good to merge?
> 
> 
> 
> @mauteri Thanks for looking at this. If ready or not is a little bit a chicken or egg question, if you now this saying.
> 
> 
> 
> Simply: No it’s not, because it will reference non-existent images from the main readme.md
> 
> 
> 
> Solution: Run the screenshot generator now once, to create the missing images, afterwards merge this PR.
> 
> 
> 
> The crux: It would be nice to have the screenshots without the alpha-notice. This will be possible with the next release. Because the screenshot-generator uses the latest release from .org, we should wait until the next is out. This leads to the need for a patch release for only screenshots and readme.
> 
> 
> 
> I don’t know.

Got it, let's chat Monday and figure it out. Thx!

PR #888, opened 2024-09-19
Prepare for new block variations 1/3: hookable-patterns, slot-fills & docs

carstingaxion:
Preparation to bring some block variations from the [gatherpress-block-playground](https://github.com/carstingaxion/gatherpress-block-playground) into the plugin as part of #626.

### Description of the Change

This PR does:

1. Bring hookable block-patterns into place _(as preparation, still disabled)_
2. Import `PluginDocumentSettingPanel` from `wordpress/editor` package _(The former source was deprecated in 6.6, which led to the annoying console warnings all over the place)_
3. Introduce new slots inside the document sidebar for event & venue post type
4. add new `/docs/developer/blocks/...`

Closes #628

### Changelog Entry

> Added - Extendable infrastructure for future blocks

### Credits

Props @carstingaxion 

### Checklist:

- [x] I agree to follow this project's [**Code of Conduct**](https://github.com/GatherPress/gatherpress/blob/main/CODE_OF_CONDUCT.md).
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my change.
- [x] All new and existing tests pass.

PR #898, opened 2024-09-21
Prepare for new block variations 3/3:  block variation loading

carstingaxion:
Preparation to bring some block variations from the [gatherpress-block-playground](https://github.com/carstingaxion/gatherpress-block-playground) into the plugin as part of #626.

While current blocks can rely on `register_block_type()` to get everything in place, block variations need more manual attention.

### Description of the Change

With this PR, I try to outline a possible way to load all variations the same way.

Because most of the variations need some extra php to work, I went with a singleton, that gets automatically ...

~~copied from `/src` to `/build` and lives in the folder of the variation. For this reason the autoloader doesn’t know about the classes and that’s why I required them manually~~ 

... loaded from `includes/core/classes/blocks`.

This design-decision was made in a private chat between @mauteri and me.

~~Closes #~~

### Changelog Entry

> Added - Generalized the loading of block-variations

### Credits

Props @carstingaxion, @mauteri 

### Checklist:


carstingaxion:
He @mauteri could you please take a look at this idea of how to load the variations?

I based this branch on the work of #888, that’s why there are so many commits, but if possible, only focus on the last three. And don’t forget about the PR description ;)

thanks.

includes/core/classes/traits/class-block-variation.php
```diff
@@ -0,0 +1,171 @@
...
+defined( 'ABSPATH' ) || exit; // @codeCoverageIgnore
+
+use Error;
+
+/**
+ * Common class that handles registering and enqueuing of assets.
+ *
+ * @since 1.0.0
+ */
+trait Block_Variation {
```
mauteri, line 21: I think we can simplify this a bit and not include a new trait. I'd add code like this to `class-assets.php`, get the variations directory from the file system and loop through. There's a lot of helpers you can use in `class-assets.php` for getting dependencies and such from the asset. Let me know what you think.

mauteri (changes requested):

includes/core/classes/traits/class-block-variation.php
carstingaxion: This sounds good.
Now that you mention the class-assets.php, I’m wondering that I haven’t thought about this. But yes, seems totally plausible for all this asset-related stuff.
Thank you.

carstingaxion:
@mauteri Do you have any clue why the „semver“ package is missing now?  (Which makes all the workflows fail)

mauteri:
> @mauteri Do you have any clue why the „semver“ package is missing now?  (Which makes all the workflows fail)

I may have accidentally excluded it when I tried to resolve conflicts with the develop branch. Sorry about that.

carstingaxion:
> > @mauteri Do you have any clue why the „semver“ package is missing now?  (Which makes all the workflows fail)
> 
> I may have accidentally excluded it when I tried to resolve conflicts with the develop branch. Sorry about that.

No problem. I'll take a look.

carstingaxion:
Im starting now and hope to get this finished.

mauteri (approved):

carstingaxion:
Woooow! That was super fast again, @mauteri 
Thank you!

PR #931, opened 2024-10-01
NEW action PHPStan & 36 little :bug: fixes

carstingaxion:
### Description of the Change

>#### WordPress extensions for PHPStan
>
>Static analysis for the WordPress ecosystem.
>
>https://github.com/szepeviktor/phpstan-wordpress

This PR introduces a new GitHub action workflow to run PHPStan incl. *the* WordPress extension. 

On behalf of this static analysis I fixed a lot of little problems. While most of the fixes are simple & self-explanatory, the following 6/36 fixes may need a second pair of eyes:

- https://github.com/GatherPress/gatherpress/pull/931/commits/684d1ec78fbe3cc62ac84da98a69b5aca1076e71
- https://github.com/GatherPress/gatherpress/pull/931/commits/3242163db44d281b8ca2ca1806d4af952c7910ae
- https://github.com/GatherPress/gatherpress/pull/931/commits/33fcfeb46992657ce8d88b5d64c3c8a8fd79f116
- https://github.com/GatherPress/gatherpress/pull/931/commits/0ad3a3e80cb76893bf8b71e91051508372d3f1bc
- https://github.com/GatherPress/gatherpress/pull/931/commits/64f02715045ed5ad7a2bf822b9e14ffd5c7f4669
- https://github.com/GatherPress/gatherpress/pull/931/commits/2c3585cc6930cb3a523ec18d9fa7c54fb4d714ea

~~Closes #~~

### How to test the Change

Compare the results of the [first PHPStan test run](https://github.com/GatherPress/gatherpress/actions/runs/11116228009/job/30886124930), with the last run of the same action.

*Looking at the **Files changed**, may not be very helpful for this PR, instead going through the commits one-by-one is recommended here.*

### Changelog Entry

> Added - Introduced PHPStan as automated GitHub action workflow and fixed 36 little(!) :bug:s that became visible, on behalf of this action.

### Credits

Props @carstingaxion @mauteri

### Checklist:

- [x] I agree to follow this project's [**Code of Conduct**](https://github.com/GatherPress/gatherpress/blob/main/CODE_OF_CONDUCT.md).
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my change.
- [ ] All new and existing tests pass.

PR #933, opened 2024-10-03
Set & use missing cache group

carstingaxion:
One of the errors, that came up in #931 was :

```sh
Parameter # 3 $group of function wp_cache_set expects string, int given.
```
in `/includes/core/classes/class-rsvp.php`

Until now, `wp_cache_set` was called with its expiration time as third parameter, which is supposed to be the cache-group.

(see https://developer.wordpress.org/reference/functions/wp_cache_set/)

### Description of the Change

I thought, that it might be the easiest solution, to just use the `'default'` cache-group, which was probably meant to be done here.

:warning: **Unfortunately this led to a [unit test error](https://github.com/GatherPress/gatherpress/actions/runs/11155630228/job/31006861221?pr=933), that I wasn't able to solve (now).**

```sh
1) GatherPress\Tests\Core\Test_Rsvp::test_responses
Failed not_attending responses are empty after $user_id_2 was deleted.
Failed asserting that an array is empty.
```

~~Closes #~~

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

Props @carstingaxion 

### Checklist:


mauteri:
@carstingaxion we can close this, fixed it here https://github.com/GatherPress/gatherpress/pull/931

PR #937, opened 2024-10-04
NEW unit tests for the Import & Export classes

carstingaxion:
### Description of the Change

~~Closes #~~

### How to test the Change

```bash
npm run test:unit:php -- --group migrate
```

### Changelog Entry

> Added - New unit tests for the Import & Export classes

### Credits

Props @carstingaxion @mauteri 

### Checklist:


carstingaxion:
Hello @mauteri ,

can you please help me here, please?

I'm wondering if the failing unit-test should be _fixed_ rather in `class-import.php` or maybe better in `class-event.php` as it could be helpful to check if the needed array keys exist before saving.

What do you think?

includes/core/classes/class-export.php
```diff
@@ -96,6 +96,18 @@ public function prepare( WP_Post $post ): void {
...
+	 * Checks if the currently exported post is of type 'gatherpress_event'.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param  WP_Post $post Current meta key.
+	 * @return bool          True, when the currently exported post is of type 'gatherpress_event', false otherwise.
+	 */
+	protected function validate( WP_Post $post ): bool {
+		return ( Event::POST_TYPE === $post->post_type );
+	}
```
mauteri, line 109: Let's add this to the new `Validate` class. `Validate::event_post_type( $post )`

https://github.com/GatherPress/gatherpress/blob/develop/includes/core/classes/class-validate.php

test/unit/php/includes/core/classes/class-test-export.php
```diff
@@ -40,6 +42,205 @@ public function test_setup_hooks(): void {
...
+			10,
+			has_action( 'the_post', array( $instance, 'prepare' ) ),
+			'Failed to assert that the "the_post" action was added.'
+		);
+
+		$this->assertSame(
+			10,
+			has_filter( 'wxr_export_skip_postmeta', array( $instance, 'extend' ) ),
+			'Failed to assert that the "wxr_export_skip_postmeta" filter was added.'
+		);
```
mauteri, line 76: There's a helper to do this. Example [here](https://github.com/GatherPress/gatherpress/blob/develop/test/unit/php/includes/core/classes/class-test-block.php#L30).

mauteri (changes requested):

includes/core/classes/class-export.php
mauteri: never mind, we can use `Validate::event_post_id` here.

mauteri (approved):

PR #939, opened 2024-10-05
Add & updated unit tests for localized slugs

carstingaxion:
### Description of the Change

This PR adds new tests for existing methods which prepare the localized post type & taxonomy slugs, before saving them into GatherPress' options.

It also and additionally _fixes_ one left-over from the development of the methods. The default strings used for the slugs were new, additional strings to translate. This PR replaces the default slugs, with the already existing `Post Type Singular Name` and `Taxonomy Singular Name` labels.

~~Closes #~~

### Changelog Entry

> Added - ... and updated unit tests for the preparation of localized post type and taxonomy slugs

### Credits

Props @carstingaxion @mauteri

### Checklist:


carstingaxion:
Another _unit-test curiosity_, that isn't clear to me @mauteri :

Why does this one test fail now?

From an untouched file.
I called `$user_id = $this->factory->user->create();` three times, which might be the cause.

includes/core/classes/class-venue.php
```diff
@@ -157,8 +157,8 @@ public function register_post_type(): void {
 	 */
 	public static function get_localized_post_type_slug(): string {
 		$switched_locale = switch_to_locale( get_locale() );
-		$slug            = _x( 'venue', 'Post Type Slug', 'gatherpress' );
-		$slug            = sanitize_title( $slug );
+		$slug            = _x( 'Venue', 'Post Type Singular Name', 'gatherpress' );
+		$slug            = sanitize_title( $slug, 'save' );
```
mauteri, line 161: setting third argument as second argument. should be `sanitize_title( $slug, '', 'save' );`

Issue #528, opened 2024-02-04
Add setting to disable block template for events and venues


mauteri:

carstingaxion:
I would suggest following WordPress‘ *decisions, not options* philosophy and not adding a setting for this (for now).

If a user is disabling the built-in template, she would have to start with a blank canvas, because we do not have any core UI available yet, wether to re-set this template definition nor to choose some patterns as starter-content. That‘s why I linked the [Gutenberg issue](https://github.com/WordPress/gutenberg/issues/53396) over here.

I could easily imagine a add-on plugin to filter the `register_post_type()` arguments to unset `template` in a first step. And second, provide some patterns in addition, using the gatherpress-core blocks.

mauteri:
@carstingaxion thanks for this. This is good feedback. 

I'll hold off on this ticket for now. I do think this will come up soon that the default blocks and the order they appear will not be ideal for some folks (especially those using a FSE theme that want to add GP blocks to a site template). 

I too could also see this as a companion plugin for GP to manipulate the templates as you stated. I'm not sure the right answer and maybe something out of core (like the ticket you linked too) will provide the solution.

carstingaxion:
For the coming WordPress 6.7 release, GatherPress is now able to & should have prepared FSE templates for its post types: 

https://developer.wordpress.org/news/2024/08/29/registering-block-templates-via-plugins-in-wordpress-6-7/

mauteri:
This was completed in 0.34.0 now that a block pattern is presented on new event.

<img width="1082" height="1142" alt="Image" src="https://github.com/user-attachments/assets/6b709517-d2f1-4366-a564-48c42a93476e" />

PR #952, opened 2024-10-16
Following a little :bug: into a :rabbit: hole

carstingaxion:
**This PR should help illustrating a problem; it's not (primarily meant to be merged)!**

---

I was thinking about #528 and the long thread @mauteri had with _Ed_ on the GatherPress slack, where they talked (among other things) about the default blocks that gets pre-populated when a **New Event** is created.

After a while .... I asked myself why GatherPress is not using the modal, provided by WP core since 5.8, that allows to select a starter-pattern for `core/post-content`. With an addition (I guess for WP 6.7) a user can now even disable this modal permanently for a given post_type, which would help GatherPress' work on #528.

_**I just wanted to try this fast**_, and if it could be made compatible with the new `register_block_template()` to create a UI for GatherPress starter-patterns for the post-content.

---

It took me a loooong while to find two weird :bug:s that prevented the modal from appearing.

https://github.com/user-attachments/assets/c26d727b-9976-4260-8c27-8962ea448e5f

### Description of the Change

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:

### DIFF
--- a/includes/core/classes/class-block.php
+++ b/includes/core/classes/class-block.php
@@ -86,6 +86,7 @@ class Block {
 	 *
 	 * @since 1.0.0
 	 * @see   https://developer.wordpress.org/reference/functions/register_block_pattern/
+	 * @see   https://developer.wordpress.org/reference/classes/wp_block_patterns_registry/register/#parameters
 	 *
 	 * @return void
 	 */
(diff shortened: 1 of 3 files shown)


mauteri:
@carstingaxion I believe we can close this PR with the completion of this. https://github.com/GatherPress/gatherpress/pull/1584

carstingaxion:
> @carstingaxion I believe we can close this PR with the completion of this. #1584

Yes, that’s true @mauteri . 

One thing may be worth creating an issue as idea, if you haven’t already implemented something similar. Within this PR I wanted to make use of register_block_template() because it would add our templates to the UI of the template editor. This was meant to be the funding for user editable post-content templates (#1353)

PR #1020, opened 2025-02-21
Revert 889 as it breaks Event Lists block. This code will need to be …

mauteri:
…revisited when we work on Query loop block and still work with upcoming/past archive pages.

### Description of the Change
Reverts https://github.com/GatherPress/gatherpress/pull/889 due to sorting bugs.

- screenshots or screencasts
-->

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

Props @mauteri 

### Checklist:

### DIFF
--- a/includes/core/classes/class-event-query.php
+++ b/includes/core/classes/class-event-query.php
@@ -6,6 +6,8 @@
  * upcoming and past events, applying filters, and ordering events. It also handles adjustments
  * for event pages and admin queries.
  *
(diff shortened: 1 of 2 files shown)


carstingaxion:
Ok, what was the problem @mauteri ?

I'm willing to fix this but don't want to have to search for the problem.

mauteri:
I don't recall exactly, PR will need to be reversed in a branch and tested.

PR #1055, opened 2025-04-21
Move GP core function to init hook to fix translation error - Take two

### LINKED ISSUE #1021: Translation notice
### Describe the bug

Since I changed the site language to Dutch, I get the debug notice 'Function _load_textdomain_just_in_time was called incorrectly. Translation loading for the gatherpress domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the init action or later. Please see Debugging in WordPress for more information. (This message was added in version 6.7.0.)'. 

### Steps to Reproduce

Set website to a default language that has translations like Dutch
Set debug to true
See the notice in admin and frontend

_No response_

### Environment information

WP 6.7.2 
Dutch lang
GatherPress 0.31.0

### WordPress information

The problem is not the translation, I think. To me the problem seems to be that the function call 'GatherPress\Core\Setup::get_instance();' in gatherpress.php is executed straight away, on loading the file. This triggers the construct in class-settings.php. But the first method in the construct needs translation functionalities, which are not prepaired yet by WordPress. The second and third depend on the first. The solution that works for me is putting them in an init hook or admin_menu hook. I have not tested this to the max yet :-) 

	protected function __construct() {
		add_action( 'init', function () { //or admin_menu if only needed in admin
			$this->instantiate_classes(); // needs translation
			$this->set_current_page(); // depends on instantiate_classes()
			$this->set_main_sub_page(); // depends on instantiate_classes()
		});

		$this->setup_hooks();
	}

Another option might be to put GatherPress\Core\Setup::get_instance() in an init-hook with priority under 10.

### Code of Conduct

jmarx:
### Description of the Change
Move sub pages core function to init hook to remove translations notice.

Closes #1021 

### How to test the Change
Switch your language setting to Dutch and turn on `debug` mode and ensure you don't get a notice about the translation functionality loading too late.

### Credits
Props @mauteri, @jmarx 

- [X] I agree to follow this project's [**Code of Conduct**](https://github.com/GatherPress/gatherpress/blob/main/CODE_OF_CONDUCT.md).
- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my change.
- [X] All new and existing tests pass.

### DIFF
--- a/includes/core/classes/settings/class-base.php
+++ b/includes/core/classes/settings/class-base.php
@@ -121,6 +121,7 @@ class Base {
 	 * @return void
 	 */
 	protected function setup_hooks(): void {
+		add_action( 'init', array( $this, 'set_main_sub_page' ) );
 		add_action( 'admin_init', array( $this, 'init' ) );
 
 		add_filter( 'gatherpress_sub_pages', array( $this, 'set_sub_page' ) );
--- a/includes/core/classes/settings/class-base.php
+++ b/includes/core/classes/settings/class-base.php
@@ -121,7 +121,6 @@ class Base {
 	 * @return void
 	 */
 	protected function setup_hooks(): void {
-		add_action( 'init', array( $this, 'set_main_sub_page' ) );
 		add_action( 'admin_init', array( $this, 'init' ) );
 
 		add_filter( 'gatherpress_sub_pages', array( $this, 'set_sub_page' ) );
--- a/includes/core/classes/class-settings.php
+++ b/includes/core/classes/class-settings.php
@@ -61,7 +61,6 @@ class Settings {
 	protected function __construct() {
 		$this->instantiate_classes();
 		$this->set_current_page();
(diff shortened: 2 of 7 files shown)


includes/core/classes/settings/class-base.php
```diff
@@ -121,6 +121,7 @@ protected function get_sections(): array {
 	 * @return void
 	 */
 	protected function setup_hooks(): void {
+		add_action( 'init', array( $this, 'set_main_sub_page' ) );
```
mauteri, line 124: Why did you add this and the empty callbacks?

mauteri (changes requested):

includes/core/classes/settings/class-base.php
jmarx: @mauteri Fixed

mauteri (approved):

Issue #1069, opened 2025-04-25
Drop Block Guard z-index from 60 to 1


mauteri:
### Describe the bug

60 is too high of a z-index and was having issues with block toolbar popovers. 30 is the highest number that it can be and not cause issues with other blocks toolbar popovers. **Update:** More issues with blocks when Block Guard is at 30. There was a small issue with 6.7.2 and before that had us make the z-index 60 (A UI when you click escape in block editor). That is gone now in 6.8 so we no longer need z-index higher than 1.

This causes editorial issues, so would like to get this into `0.32.2` which we can release some time next week. I'd like to see if there are any other issues discovered that we'd also like to get in a that patch.

### Steps to Reproduce

N/A

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1072, opened 2025-04-26
Version 0.32.2 update

### LINKED ISSUE #1070: Fix Dropdown Item block with target and nofollow settings
### Describe the bug

The anchor feature in Dropdown Item block has target and nofollow settings, but they do not retain when checked. We'll make this fix for 0.32.2 as well as this should be fairly small change.

### Steps to Reproduce

1. Select Dropdown Item block
2. Check Advanced in URL edit
3. Check Open in new tab and/or Mark as nofollow
4. Checkboxes do not retain or show up in block

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

mauteri:
### Description of the Change
* We had z-index of the Block Guard at 60 because of a state in block editor where it changed UI a bit when clicking escape (this feature is now confirmed gone in 6.8). The z-index was conflicting poorly with other blocks. I tried 30 and then 29 and still got some rare conflicts. I dropped it to 1 since that original issue in 6.7.2 and less was now gone so z-index no longer needed to be high for that.
* Dropdown Item block was missing attributes for target and rel in link, which is now fixed with this update.
* Added sanitization to GatherPress settings page.
* Test up to 6.8.1

Closes #1070 #1069

### Changelog Entry

> Fixed - Bug fix

### Credits

Props @mauteri

### Checklist:


JordanPak (commented):
LGTM

PR #1094, opened 2025-05-02
Added a new check to determine if gatherpress has already been installed/activated

stephenerdelyi:
…

### Description of the Change

https://github.com/orgs/GatherPress/projects/1/views/1?pane=issue&itemId=107988139&issue=GatherPress%7Cgatherpress%7C1071

Closes GP-1071

### How to test the Change

Try to activate 2 different versions of GP at the same time and you will see an activation error message

### Changelog Entry

> Added - new check to search for other instances of GP installation

### Credits
@stephenerdelyi 

### Checklist:

### DIFF
--- a/gatherpress.php
+++ b/gatherpress.php
@@ -21,6 +21,11 @@
 // Exit if accessed directly.
 defined( 'ABSPATH' ) || exit; // @codeCoverageIgnore
 
+// Ensure no other versions of GatherPress are currently running.
+if ( require_once __DIR__ . '/includes/core/duplicate-check.php' ) {
+	return;
+}
+
 // Constants.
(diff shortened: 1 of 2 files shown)


mauteri:
@stephenerdelyi let's just add that [deactivate plugin](https://developer.wordpress.org/reference/functions/deactivate_plugins/) to this like we chatted about. Thanks!

stephenerdelyi:
> @stephenerdelyi let's just add that [deactivate plugin](https://developer.wordpress.org/reference/functions/deactivate_plugins/) to this like we chatted about. Thanks!

As we discussed in slack, there's no way to identify which plugin to deactivate, just the first one loaded which can cause other concerns. We decided since this is so rare, the current implementation is fine.

mauteri (approved):

PR #1103, opened 2025-05-06
Prep work for comment flow for open RSVP and small fixes

mauteri:
### Description of the Change
Mostly adjusts how RSVP comments are handled so they can be identified by email as well as user ID for new open RSVP flow that's to be developed.

New RSVP list table view.

Other small fixes and adjustments found with Event List styling and Block Guard were addressed.

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

Props @mauteri 

### Checklist:


patriciabt:
Testing now, I love the RSVPs admin page. 
Issue: I cannot re-approve someone I unapproved earlier
Feature request: a filter at the top, for event, date, topic, venue (like post filters for date and category)

includes/core/open-rsvp.php
```diff
@@ -0,0 +1,67 @@
...
+		$_SERVER['DOCUMENT_ROOT']
+	);
+
+	// Prevent directory traversal.
+	$gatherpress_document_root = str_replace(
+		array( "\0", '..' ),
+		'',
+		$gatherpress_document_root
+	);
+}
```
dd32, line 29: You can't realistically use DOCUMENT_ROOT to find the WordPress ABSPATH, as it may be in a sub-folder.

Pretty sure you can't use `require wp-load.php` in w.org hosted plugins either.

Instead, you should be hooking in early on the `wp_loaded` action and determining that it's a request you wish to add the filters to.

mauteri: This is still an early test (PR is still Draft), but since this code got your attention 😄 I'm looking to doing something similar to how `wp-comments-post.php` works, which WordPress POSTs to directly from a form to create a comment. The difference here is that I need to fire some filters to change the context from a normal comment to an RSVP which is a comment type. I wanted to follow this path as it will natively work with plugins like Akismet and won't run into issues around caching. I'm still thinking about best way to do this, but yeah, `DOCUMENT_ROOT` falls short in some cases, but trying to create a proxy file for `wp-comments-post.php` needs to call `wp-loaded` in order to fire filters. Maybe I just go to calling `wp-comments-post.php` directly, but include something in the request to set these filters. If you have other ideas let me know.

mauteri: @dd32 made some updates. Let me know what you think of this approach. Thx!

PR #1104, opened 2025-05-21
Don't include the .wordpress-org directory in the released versions of the plugin

dd32:
### Description of the Change

While looking at https://plugins.trac.wordpress.org/browser/gatherpress/#tags/0.32.2 and https://plugins.trac.wordpress.org/browser/gatherpress/trunk I noticed that you've included the `.wordpress-org` directory in the released versions of the plugin, although this is intended to be the `assets/` folder only.

See https://github.com/10up/action-wordpress-plugin-deploy?tab=readme-ov-file#excluding-files-from-deployment

### Changelog Entry

Fixed - smaller release ZIPs as we no longer ship the screenshots as part of the release.

### Checklist:

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -40,6 +40,7 @@ yarn.lock
 
 # Dot Files and Directories
 .gatherpress-org
+.wordpress-org
 .git
 .github
 .wordpress-version-checker.json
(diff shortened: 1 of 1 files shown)


dd32:
Additionally, the docs probably are best not shipped as well.
https://plugins.trac.wordpress.org/browser/gatherpress/tags/0.32.2/docs

mauteri (approved):

PR #1130, opened 2025-07-25
Add a privacy policy content suggestion

### LINKED ISSUE #875: IMHO, GatherPress should suggest some text to the websites privacy-page, which explains the use.

carstingaxion:
### Description of the Change

image

Closes #875

### Changelog Entry

> Added a privacy policy content suggestion

### Credits

Props @carstingaxion 

### Checklist:

### DIFF
--- a/includes/core/classes/class-setup.php
+++ b/includes/core/classes/class-setup.php
@@ -240,6 +240,7 @@ class Setup {
 		$content = '<p class="privacy-policy-tutorial">' . __( 'Inform your visitors about GatherPress\' use of OpenStreetMap services.', 'gatherpress' ) . '</p>'
 				. '<strong class="privacy-policy-tutorial">' . __( 'Suggested Text:', 'gatherpress' ) . '</strong> '
 				. sprintf(
+					/* translators: %1$s: privacy policy URL of the OpenStreetMap foundation */
 					__( 'When viewing maps on event or venue pages, your IP address and certain technical information (such as browser type and referrer URL) are transmitted to the OpenStreetMap Foundation, which operates the map service. This data is processed according to their <a href="%1$s" target="_blank">privacy policy</a>. For more information about what data OpenStreetMap collects and how it is used, please refer to their <a href="%1$s" target="_blank">privacy documents</a>.', 'gatherpress' ),
 					'https://osmfoundation.org/wiki/Privacy_Policy'
 				);
--- a/includes/core/classes/class-setup.php
+++ b/includes/core/classes/class-setup.php
@@ -243,7 +243,7 @@ class Setup {
 			return;
 		}
 		$content = '<h2>' . __( 'Inform your visitors about GatherPress\' use of OpenStreetMap services.', 'gatherpress' ) . '</h2>'
-				. '<p><strong class="privacy-policy-tutorial">' . __( 'Suggested Text:' ) . '</strong> ' // phpcs:ignore WordPress.WP.I18n.MissingArgDomainDefault
+				. '<p><strong class="privacy-policy-tutorial">' . __( 'Suggested Text:', 'default' ) . '</strong> '
 				. sprintf(
 					/* translators: %1$s: privacy policy URL of the OpenStreetMap foundation */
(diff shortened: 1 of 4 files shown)


includes/core/classes/class-setup.php
```diff
@@ -216,6 +217,41 @@ private function maybe_create_flush_rewrite_rules_flag(): void {
...
+	 * @since 1.0.0
+	 *
+	 * @return void
+	 */
+	public function add_privacy_policy_content() {
+		if ( ! function_exists( 'wp_add_privacy_policy_content' ) ) {
+			return;
+		}
+		$content = '<h2>' . __( 'Inform your visitors about GatherPress\' use of OpenStreetMap services.', 'gatherpress' ) . '</h2>'
+				. '<p><strong class="privacy-policy-tutorial">' . __( 'Suggested Text:', 'default' ) . '</strong> '
```
mauteri, line 246: Fix text domain

mauteri (changes requested):

includes/core/classes/class-setup.php
carstingaxion: No, it's intentional. 

GatherPress does not need to stress the translation contributors with defaults that are translated a thousand times.

mauteri: Okay, I would omit it then and just have `__( 'Suggested Text:' )` since `default` is the default and wp plugins guidelines is flagging it. Unless there's a reason not to do that.

carstingaxion: This is a little funny, because

I did exactly what you suggest in my [commits here](https://github.com/GatherPress/gatherpress/pull/1130/commits/27e5810917875d4e0d46dd6b42bbba5814db3214) before *(I guess)* phpcs complained about the missing `'default'` textdomain.
So I added it, expecting it to work now.

Ok, now what?

mauteri: Are you saying PHPCS complains if you leave it off and WP Plugin Guidelines complains if it's not "gatherpress"? Maybe don't include it and if PHPCS complains, add a PHPCS Ignore.

carstingaxion: According to the current `phpcs.ruleset.xml`, this should be allowed:

```xml
	<rule ref="WordPress.WP.I18n">
		<properties>
			<!-- Value: replace the text domain used. -->
			<property name="text_domain" type="array">
				<element value="gatherpress"/>
				<!-- No need to translate 100%-core strings -->
				<element value="default"/>
			</property>
		</properties>
	</rule>
```

carstingaxion: > Maybe don't include it and if PHPCS complains, add a PHPCS Ignore.

Ok, I did exactly this in my last 2 commits. Lets see what the guidelines check, will say ...

carstingaxion: image

Can you restart the workflow please @mauteri ?

mauteri: Just restarted it.

carstingaxion: Thanks.

But this keeps being suspicious, the wp.org plugin directory guidelines are now constantly failing with the error shown in the screenshot, which seems to be totally unrelated.

Any ideas @mauteri ?

mauteri: @carstingaxion Try changing tested up to 6.8.1 to just 6.8. Seems that is what it is erroring on (which is new, as it's been that way for awhile).

mauteri: If that fixes it, we can put in a PR to change that and pull it into all PRs we're working on.

carstingaxion: > All checks have passed
18 successful checks

**That did it.**

So next, I'm going top re-add the 'default' textdomain and remove the phpcs:ignore statement.

carstingaxion: > So next, I'm going top re-add the 'default' textdomain and remove the phpcs:ignore statement.

Which worked out very well, with no errors.

Do you still need additional information on the _Why & How_ of this idea @mauteri  or are you able to merge it?

mauteri: @carstingaxion I'll get this on my list to do final review and merge if looks good. Thx!

carstingaxion:
Hooray 🎉 and thanks for merging @mauteri !

PR #1134, opened 2025-07-27
New 'Upcoming' & 'Past' admin event list table views.

carstingaxion:
## *needs to be done|solved|fixed*

- [x] `adjust_event_sql()` is called twice on the same query, when sorting by event date, which leads to SQL errors
- [x] `adjust_event_sql()` has `orderby` hardcoded, which doesn't know about the `orderby` of the current query

### Description of the Change

[Bildschirmaufzeichnung vom 27.07.2025, 13:23:06.webm](https://github.com/user-attachments/assets/59e6e1d3-8556-45f7-a0dd-36dc245d3a73)

Closes #694

### Changelog Entry

> Added - New 'Upcoming' & 'Past' admin event list table views.

### Credits

Props @carstingaxion, @kofimokome 

### Checklist:

- [x] I agree to follow this project's [**Code of Conduct**](https://github.com/GatherPress/gatherpress/blob/main/CODE_OF_CONDUCT.md).
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my change.
- [x] All new and existing tests pass.

Issue #694, opened 2024-06-17
Sorting & Filtering of Admin List Tables


carstingaxion:
### Is your enhancement related to a problem? Please describe.

I’d like to have a little more flexibility when using the admin table lists for events and venues.

GatherPress should add drop-down filters …

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

carstingaxion:
~~GatherPress should add drop-down filters~~

No, GatherPress should just wait for the new core data-views.

PR #1171, opened 2025-08-13
Order events in admin-list-table by eventdate by default 

carstingaxion:
### Description of the Change

This PR changes the default sorting, that is used for event post listings in the backend. Events are now sorted by event date by default, instead of their publication date.

_**BEFORE:**_
image

**_AFTER:_**
image

~~Closes #~~ Part of #694 and directly related to #1134

### How to test the Change

Visit `https://example.org/wp-admin/edit.php?post_type=gatherpress_event` and make sure you see:

1. the events ordered by event date, 
2. with the oldest first

### Changelog Entry

> Changed - the default ordering of events in the Admin list view to be ordered by event date, not the published date.

### Credits

Props @carstingaxion 

### Checklist:


jmarx (approved):
LGTM

carstingaxion:
> LGTM

Thanks @jmarx!

mauteri:
@carstingaxion I see this is approved and looks good, is there a reason to keep this still in draft or is it ready for review and could be merged?

carstingaxion:
Let me have an additional look, I'm not sure if this is ready.

carstingaxion:
> @carstingaxion I see this is approved and looks good, is there a reason to keep this still in draft or is it ready for review and could be merged?

This is good to be merged.

But, it would only makes sense in combination with #1134, I guess.

includes/core/classes/class-event-setup.php
```diff
@@ -472,6 +473,32 @@ public function set_custom_columns( array $columns ): array {
...
+
+		// If the screen is already ordered, bail out.
+		if ( isset( $_GET['orderby'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			return;
+		}
+
+		// Define the event date as default ORDERBY and ORDER ascending.
+		$_GET['orderby'] = 'datetime';
+		$_GET['order']   = 'asc';
+	}
```
mauteri, line 721: @carstingaxion branch needs to be updated with develop so the new PR Coverage workflow runs (I think this PR was opened before that was implemented). If you'd like me to write a unit test for the new method let me know. Thanks!

carstingaxion: That would be a pleasure, if you could help @mauteri!

I will have to learn writing tests, if I want to contribute here in the future, this new check slightly increases my barrier to contribute, I guess.

mauteri: @carstingaxion I added some unit tests, but also found that the `asc` should be `desc` as we'd want the most recent. With `desc` it will show the oldest events first. I think the `asc` only works when you include this [PR](https://github.com/GatherPress/gatherpress/pull/1134) and also default to `upcoming`. Is that the intent?

mauteri: @carstingaxion Okay, I merged this [PR](https://github.com/GatherPress/gatherpress/pull/1134) first which allowed me to created a default sort with "Upcoming" as well, which is what I think you were after and what most people want; easy view with upcoming events which are what is relevant. I put it back to `asc` since when combined with `upcoming` makes a lot of sense. Did some other housekeeping and UI improvements. I'll merge this once tests and linting passes, but take a look at changes I made and let me know if you have any questions or concerns. Nice work!

mauteri (approved):

includes/core/classes/class-event-setup.php
carstingaxion: Thanks for your help @mauteri!!

You connected the dots very well, having the default sorting like this, was what I had in mind initially. Cool you also brought in the counters.

Very, very cool. I'm happy, and I hope GatherPress' users will be, too.

PR #1204, opened 2025-10-31
Feature/ai abilities api

jmarx:
### Description of the Change

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:


includes/core/classes/ai/class-abilities-integration.php
```diff
@@ -0,0 +1,1860 @@
...
+	 * @param array $params Optional parameters (currently unused).
+	 * @return array Response with venue list or error.
+	 */
+	// phpcs:ignore Generic.CodeAnalysis.UnusedFunctionParameter.Found,Squiz.Commenting.FunctionComment.Missing
+	public function execute_list_venues( array $params = array() ): array {
+		try {
+			$venues = get_posts(
+				array(
+					'post_type'      => Venue::POST_TYPE,
+					'posts_per_page' => -1,
```
mauteri, line 700: `-1` can be incredibly problematic as it does a full table scan and have issues with large sites. I would use [wp_count_posts](https://developer.wordpress.org/reference/functions/wp_count_posts/) which will handle the upper limit for the number of posts for a post type and caches that number.

includes/core/classes/ai/class-abilities-integration.php
```diff
@@ -0,0 +1,1860 @@
...
+				);
+			}
+		}
+
+		// Update datetime if provided using the Event class method.
+		$event = new Event( $event_id );
+
+		// Clear cache to ensure we get fresh datetime data.
+		$cache_key = sprintf( Event::DATETIME_CACHE_KEY, $event_id );
+		delete_transient( $cache_key );
```
mauteri, line 1306: Reading this made me realize we should update this `get_datetime` method to get data from post meta instead of the custom table, since datetime is now stored there too. I put in a ticket, so we will soon be able to get rid of the transient and just call meta. https://github.com/GatherPress/gatherpress/issues/1331

includes/core/classes/ai/class-admin-page.php
```diff
@@ -0,0 +1,257 @@
...
+			'gatherpress-ai-assistant',
+			GATHERPRESS_CORE_URL . 'build/style-ai-assistant-style.css',
+			$style_asset['dependencies'] ?? array(),
+			$style_asset['version'] ?? ''
+		);
+
+		wp_enqueue_script(
+			'gatherpress-ai-assistant',
+			GATHERPRESS_CORE_URL . 'build/ai-assistant.js',
+			array_merge( array( 'jquery' ), $script_asset['dependencies'] ?? array() ),
```
mauteri, line 101: Let's stay away from jQuery. No need for it.

includes/core/classes/ai/class-admin-page.php
```diff
@@ -0,0 +1,257 @@
...
+		}
+
+		$style_asset  = $this->get_asset_data( 'ai-assistant-style' );
+		$script_asset = $this->get_asset_data( 'ai-assistant' );
+
+		wp_enqueue_style(
+			'gatherpress-ai-assistant',
+			GATHERPRESS_CORE_URL . 'build/style-ai-assistant-style.css',
+			$style_asset['dependencies'] ?? array(),
+			$style_asset['version'] ?? ''
```
mauteri, line 95: Don't need `??` on these. If they're not there, it's a broken build.

includes/core/classes/ai/class-admin-page.php
```diff
@@ -0,0 +1,257 @@
...
+		);
+
+		wp_enqueue_script(
+			'gatherpress-ai-assistant',
+			GATHERPRESS_CORE_URL . 'build/ai-assistant.js',
+			array_merge( array( 'jquery' ), $script_asset['dependencies'] ?? array() ),
+			$script_asset['version'] ?? '',
```
mauteri, line 102: Don't need `?? ''`

includes/core/classes/ai/class-admin-page.php
```diff
@@ -0,0 +1,257 @@
...
+		);
+
+		wp_localize_script(
+			'gatherpress-ai-assistant',
+			'gatherpressAI',
+			array(
+				'ajaxUrl' => admin_url( 'admin-ajax.php' ),
+				'nonce'   => wp_create_nonce( 'gatherpress_ai_nonce' ),
+			)
+		);
```
mauteri, line 113: At some point switch this over to a REST API endpoint and don't use admin-ajax. Also, you can access the admin-ajax.php endpoint with the `ajaxurl` global in the admin.

includes/core/classes/ai/class-admin-page.php
```diff
@@ -0,0 +1,257 @@
...
+				</div>
+
+				<div class="gp-ai-status" id="gp-ai-status" style="display:none;">
+					<p class="gp-ai-processing">
+						<span class="spinner is-active"></span>
+						<?php esc_html_e( 'Processing...', 'gatherpress' ); ?>
+					</p>
+				</div>
+			</div>
+		</div>
```
mauteri, line 222: This much markup should be in a template and called in the method. Markup in class methods looks messy and is hard to maintain.

includes/core/classes/ai/class-admin-page.php
```diff
@@ -0,0 +1,257 @@
...
+		<?php
+	}
+
+	/**
+	 * Handle AJAX request to process AI prompt.
+	 *
+	 * @return void
+	 */
+	public function process_prompt_ajax(): void {
```
mauteri, line 231: Create REST API endpoint.

includes/core/classes/ai/class-admin-page.php
```diff
@@ -0,0 +1,257 @@
...
+		return ! empty( $api_key );
+	}
+
+	/**
+	 * Get asset data from build directory.
+	 *
+	 * @param string $asset Asset name without extension.
+	 * @return array Asset data including dependencies and version.
+	 */
+	private function get_asset_data( string $asset ): array {
```
mauteri, line 594: Thinking we refactor the [get_asset_data](https://github.com/GatherPress/gatherpress/blob/develop/includes/core/classes/class-assets.php#L420) method into our Utility class so we can use it here and in class-assets.php instead of having repeated code.

includes/core/classes/ai/class-date-calculator.php
```diff
@@ -0,0 +1,587 @@
...
+	 * @return array Array of date strings in Y-m-d format.
+	 */
+	private function calculate_recurring_dates( string $pattern, int $occurrences, \DateTime $start_datetime ): array {
+		$dates       = array();
+		$pattern_low = strtolower( trim( $pattern ) );
+
+		// Handle relative patterns first (next, last, this, tomorrow, yesterday).
+		$relative_weekday_pattern = '/^(next|last|this)\s+'
+			. '(monday|tuesday|wednesday|thursday|friday|saturday|sunday)$/i';
+		if ( preg_match( $relative_weekday_pattern, $pattern_low, $matches ) ) {
```
mauteri, line 117: Something like this would be a cleaner approach than this super long if/elseif statement.

```
private function get_pattern_handlers(): array {
	return [
		'/^(next|last|this)\s+(monday|tuesday|...)$/i' => 'handle_relative_weekday',
		'/^(tomorrow|yesterday)$/i' => 'handle_relative_day',
		// ...
	];
}

private function calculate_recurring_dates( string $pattern, int $occurrences, \DateTime $start ): array {
	foreach ( $this->get_pattern_handlers() as $regex => $method ) {
		if ( preg_match( $regex, $pattern, $matches ) ) {
			return $this->$method( $matches, $occurrences, $start );
		}
	}
	return array();
}
```

includes/core/classes/ai/class-date-calculator.php
```diff
@@ -0,0 +1,587 @@
...
+	 * @param int    $year    Year.
+	 * @param int    $month   Month.
+	 * @param string $weekday Weekday name (e.g., "monday").
+	 * @param int    $nth     Nth occurrence (1-5, or -1 for last).
+	 * @return string Date in Y-m-d format.
+	 */
+	private function get_nth_weekday_of_month( int $year, int $month, string $weekday, int $nth ): string {
+		if ( -1 === $nth ) {
+			// Last occurrence.
+			$date = new \DateTime( "last {$weekday} of {$year}-{$month}" );
```
mauteri, line 253: Add `use DateTime` at top of file and call `new DateTime` (here and everywhere)

includes/core/classes/ai/class-openai-handler.php
```diff
@@ -0,0 +1,563 @@
...
+
+	/**
+	 * Get GatherPress abilities as OpenAI functions.
+	 *
+	 * @return array Array of function definitions for OpenAI.
+	 */
+	private function get_gatherpress_functions() {
+		// Check if external AI plugin's calculate-dates ability is available.
+		// If so, use it instead of GatherPress's own implementation.
+		$calculate_dates_ability = Abilities_Integration::get_calculate_dates_ability();
```
mauteri: seems variable is only used once, so no need to set the variable, just include it in the array.

includes/core/classes/ai/class-openai-handler.php
```diff
@@ -0,0 +1,563 @@
...
+						'type'     => 'function',
+						'function' => $func,
+					);
+				},
+				$functions
+			),
+			'tool_choice' => 'auto',
+		);
+
+		$response = wp_remote_post(
```
mauteri: Use [wp_safe_remote_post](https://developer.wordpress.org/reference/functions/wp_safe_remote_post/) here.

includes/core/classes/ai/class-openai-handler.php
```diff
@@ -0,0 +1,563 @@
...
+			$openai_response = $this->call_openai_api( $messages, $functions, $api_key );
+
+			if ( is_wp_error( $openai_response ) ) {
+				return $openai_response;
+			}
+		}
+
+		// Max iterations reached.
+		return array(
+			'response' => 'Task partially completed. Maximum iterations reached.',
```
mauteri: Does this need translation function?

includes/core/classes/settings/class-ai.php
```diff
@@ -0,0 +1,163 @@
...
+							'options' => array(
+								'default' => '',
+							),
+						),
+					),
+				),
+			),
+		);
+	}
+}
```
mauteri: I would make the [AI Experiments](https://wordpress.org/plugins/ai/) plugin a dependency of this feature. Not sure when/if that code is making it into core eventually. But we should be utilizing core settings when we can and not including our own settings of the same thing. Maybe get some guidance here from @jeffpaul

src/ai/index.js
```diff
@@ -0,0 +1,154 @@
...
+
+		// Clear input
+		$prompt.val( '' );
+
+		// Disable submit button
+		$submit.prop( 'disabled', true );
+		$status.show();
+
+		// Send to backend
+		$.ajax( {
```
mauteri: Change to REST API and use `apiFetch`

src/ai/style.scss
```diff
@@ -0,0 +1,168 @@
+.gp-ai-assistant {
```
mauteri, line 1: Need to prepend with `gatherpress-` not `gp-`

webpack.config.js
```diff
@@ -70,6 +70,8 @@ module.exports = [
 				'style.scss',
 			),
 			utility_style: path.resolve( process.cwd(), 'src', 'utility.scss' ),
+			'ai-assistant': path.resolve( process.cwd(), 'src/ai', 'index.js' ),
+			'ai-assistant-style': path.resolve( process.cwd(), 'src/ai', 'style.scss' ),
```
mauteri, line 74: use underscores to keep format.

includes/core/classes/settings/class-ai.php
jmarx: @mauteri very hard disagree. But just wait a bit for my latest stuff :) I'll explain what im doing

PR #1231, opened 2025-11-20
Simplify rewrite rules flushing by leveraging WordPress core

### LINKED ISSUE #880: Simplify maybe_flush_rewrites() method

The current implementation of automated checks to see if it might be necessary to flush the rewrite rules is ok and working, but could be improved.

### Designs

Instead of setting an own` do_flush_the_rules` option to check against, GatherPress could rely on WordPress core functionality to re-create the rewrite rules on the fly, if they don’t exist. 

This is happening early on each request and checks for the existence of the `rewrite_rules` option.

I suggest that GatherPress _just_ delete the `rewrite_rules` core option, if the rules need to be flushed.
And let WordPress handle the rest.

_No response_

### Code of Conduct

pbrocks:
## Summary

Simplifies the rewrite rule flushing mechanism by replacing custom flag-based tracking with WordPress core's built-in rewrite rule regeneration.

Resolves #880

### Removed
- Custom `gatherpress_flush_rewrite_rules_flag` option
- `maybe_create_flush_rewrite_rules_flag()` method
- `maybe_flush_rewrite_rules()` public method in Setup class
- `init` action hook that checked flag on every request

### Modified
- **class-setup.php**: 
  - Replaced custom flag methods with new `schedule_rewrite_flush()` method
  - Simplified logic to just delete `rewrite_rules` core option
  - Added cleanup for legacy flag option
- **class-settings.php**: 
  - Updated `maybe_flush_rewrite_rules()` to use core option deletion
  - Improved documentation
- **class-test-setup.php**: 
  - Updated test to verify new `schedule_rewrite_flush()` behavior
  - Added test for legacy flag cleanup

## How It Works

Instead of maintaining a custom `gatherpress_flush_rewrite_rules_flag` option:

1. When rewrite rules need flushing, we simply delete the `rewrite_rules` core option
2. WordPress automatically detects the missing option on the next request
3. WordPress regenerates and saves the rewrite rules automatically
4. No custom tracking needed - WordPress handles it natively

## Benefits

1. **Simpler code**: Removes ~30 lines of custom tracking logic
2. **Better performance**: No database check on every `init` action
3. **WordPress best practices**: Leverages core's built-in mechanism
4. **Cleaner database**: One less custom option to manage
5. **More reliable**: Uses WordPress's tested rewrite regeneration system

## Testing

- [x] Plugin activation triggers rewrite rule regeneration
- [x] Changing URL structure settings regenerates rules
- [x] Plugin deactivation properly removes custom rewrite rules
- [x] Updated unit tests pass
- [x] PHPStan analysis passes (pre-existing errors in unrelated files)
- [x] Legacy flag option is cleaned up during migration

## Migration

The new `schedule_rewrite_flush()` method automatically cleans up the old `gatherpress_flush_rewrite_rules_flag` option, so existing installations will migrate seamlessly.

### DIFF
--- a/includes/core/classes/class-settings.php
+++ b/includes/core/classes/class-settings.php
@@ -833,7 +833,7 @@ class Settings {
 	 * rewrite rule regeneration on the next request. This is more efficient
 	 * than using a custom flag option.
 	 *
-	 * @since 0.33.0
+	 * @since 1.0.0
 	 *
 	 * @param mixed $old_value The old option value.
 	 * @param mixed $new_value The new option value.
--- a/includes/core/classes/class-setup.php
+++ b/includes/core/classes/class-setup.php
@@ -192,7 +192,7 @@ class Setup {
 	 * calling flush_rewrite_rules() directly and removes the need for a custom
 	 * flag option.
 	 *
-	 * @since 0.33.0
+	 * @since 1.0.0
 	 *
 	 * @return void
 	 */
--- a/package-lock.json
+++ b/package-lock.json
@@ -172,6 +172,7 @@
 			"integrity": "sha512-bXYxrXFubeYdvB0NhD/NBB3Qi6aZeV20GOWVI47t2dkecCEoneR4NPVcb7abpXDEvejgrUfFtG6vG/zxAKmg+g==",
 			"dev": true,
 			"license": "MIT",
+			"peer": true,
 			"dependencies": {
 				"@ampproject/remapping": "^2.2.0",
 				"@babel/code-frame": "^7.27.1",
@@ -2140,6 +2141,7 @@
 				}
 			],
 			"license": "MIT",
+			"peer": true,
 			"engines": {
 				"node": ">=18"
 			},
@@ -2163,6 +2165,7 @@
 				}
 			],
 			"license": "MIT",
+			"peer": true,
 			"engines": {
 				"node": ">=18"
 			}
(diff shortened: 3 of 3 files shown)


includes/core/classes/class-settings.php
```diff
@@ -824,15 +824,19 @@ public function url_rewrite_preview( string $name, string $value ): void {
...
 	 *
 	 * Fires after the value of the 'gatherpress_general["urls"]' option-part has been successfully updated
 	 * and only if it has changed since before.
 	 *
-	 * @since 1.0.0
+	 * Deletes the core rewrite_rules option to trigger WordPress's automatic
+	 * rewrite rule regeneration on the next request. This is more efficient
+	 * than using a custom flag option.
+	 *
+	 * @since 0.33.0
```
mauteri: Just keep these all at 1.0.0 (here and below), I know it's not true, but we haven't been tracking alpha version of the plugin and when we hit 1.0.0, this is everything that was worked on up to this. Thx!

includes/core/classes/class-setup.php
```diff
@@ -186,37 +185,22 @@ public function deactivate_gatherpress_plugin(): void {
...
-	 * @since 1.0.0
-	 *
-	 * @return void
-	 */
-	private function maybe_create_flush_rewrite_rules_flag(): void {
-		if ( ! get_option( 'gatherpress_flush_rewrite_rules_flag' ) ) {
-			add_option( 'gatherpress_flush_rewrite_rules_flag', true );
-		}
+		// Clean up old flag option if it exists from previous versions.
+		delete_option( 'gatherpress_flush_rewrite_rules_flag' );
```
mauteri, line 203: We can add this cleanup to [GatherPress Alpha](https://github.com/GatherPress/gatherpress-alpha)

includes/core/classes/class-settings.php
pbrocks: adjiusted, @mauteri

includes/core/classes/class-setup.php
pbrocks: do I need to do anything here?

mauteri: Nope, I can handle it. Thanks.

mauteri (approved):

PR #1335, opened 2026-01-09
updated get_datetime function to return static post meta data

### LINKED ISSUE #1331: Update get_datetime method in class-event.php to retrieve datatime data from post meta instead of custom table

Storing datetime values in meta data was introduced in this [issue](https://github.com/GatherPress/gatherpress/issues/803). However, the [get_datetime](https://github.com/GatherPress/gatherpress/blob/develop/includes/core/classes/class-event.php#L434) method (which is used a LOT) is still getting this data from the table and setting a transient to cache results. This method can be a lot less complicated by just getting the data from the meta value now. 

So in this ticket, we should remove the calls to the database table, remove transient cache as that will no longer be needed, and just simply call the meta data values which are the same as the database field names but prepended with gatherpress_.

### Designs

_No response_

_No response_

### Code of Conduct

stephenerdelyi:
(solves [issue/1331](https://github.com/GatherPress/gatherpress/issues/1331))

### Description of the Change

Updated `get_datetime()` function to return static post meta data as opposed to a cached custom database table call.

Resolves #1331 

### How to test the Change

Fire up GP and attempt to create a new event. You should see the time editable in the backend. Hit save and confirm you see the expected (set) time on the frontend.

### Changelog Entry

> Changed - `get_datetime()`

### Credits

@stephenerdelyi

### Checklist:

### DIFF
--- a/includes/core/classes/class-event.php
+++ b/includes/core/classes/class-event.php
@@ -440,10 +440,6 @@ class Event {
 	 *     - 'timezone'           (string) The timezone of the event, adjusted per user or site settings.
 	 */
 	public function get_datetime(): array {
-		if ( null !== $this->datetime_cache ) {
-			return $this->datetime_cache;
-		}
-
 		$data = array(
 			'datetime_start'     => '',
 			'datetime_start_gmt' => '',
@@ -456,10 +452,21 @@ class Event {
 			return $data;
 		}
 
+		if ( null !== $this->datetime_cache ) {
+			return $this->datetime_cache;
+		}
+
 		foreach ( array_keys( $data ) as $key ) {
 			$result = get_post_meta( $this->event->ID, Utility::prefix_key( $key ), true );
 
-			if ( ! empty( $result ) && Validate::datetime( $result ) ) {
+			if ( empty( $result ) ) {
+				continue;
+			}
+
+			// Validate datetime fields vs timezone field.
+			if ( 'timezone' === $key && Validate::timezone( $result ) ) {
+				$data[ $key ] = $result;
+			} elseif ( Validate::datetime( $result ) ) {
 				$data[ $key ] = $result;
 			}
 		}
@@ -883,11 +890,12 @@ class Event {
 				$fields,
 				array( 'post_id' => $fields['post_id'] )
 			);
-
-			$this->datetime_cache = null;
 		} else {
 			$value = $wpdb->insert( $table, $fields );
 		}
+
+		// Clear cache after insert or update.
+		$this->datetime_cache = null;
 		// phpcs:enable WordPress.DB.DirectDatabaseQuery.DirectQuery
 		// phpcs:enable WordPress.DB.DirectDatabaseQuery.NoCaching
 
--- a/test/unit/php/includes/tests/core/classes/class-test-import.php
+++ b/test/unit/php/includes/tests/core/classes/class-test-import.php
@@ -284,6 +284,9 @@ class Test_Import extends Base {
 		);
 		$instance->datetimes_callback( $post->ID, $meta_data_value );
 
+		// Create new Event instance to get fresh data after datetime save.
+		$event = new Event( $post->ID );
+
 		$expect = array(
 			'datetime_start'     => '2020-05-11 15:00:00',
 			'datetime_start_gmt' => '2020-05-11 19:00:00',


includes/core/classes/class-event.php
```diff
@@ -443,38 +443,17 @@ public function get_datetime(): array {
...
-			// phpcs:enable WordPress.DB.DirectDatabaseQuery.DirectQuery
-			// phpcs:enable WordPress.DB.DirectDatabaseQuery.NoCaching
-			// phpcs:enable WordPress.DB.PreparedSQLPlaceholders.UnsupportedIdentifierPlaceholder
-			$data = ( ! empty( $data ) ) ? (array) current( $data ) : array();
+		foreach ( $data as $key => & $value ) {
+			$result = get_post_meta( $this->event->ID, 'gatherpress_' . $key, true );
 
-			set_transient( $cache_key, $data, 15 * MINUTE_IN_SECONDS );
+			if ( ! empty( $result ) ) {
+				$value = $result;
```
mauteri: This can be a bit confusing as it's easy to miss the subtle `& $value` in the  `foreach`. Let's remove that `&`, use `array_keys( $data )` and just do `$data[ $key ] = $result;` as that's the more common and clear pattern.
```
foreach ( array_keys( $data ) as $key ) {
	$result = get_post_meta( $this->event->ID, Utility::prefix_key( $key ), true );

	if ( ! empty( $result ) && Validate::datetime( $result ) ) {
		$data[ $key ] = $result;
	}
}
```

includes/core/classes/class-event.php
```diff
@@ -443,38 +443,17 @@ public function get_datetime(): array {
...
-					$table,
-					$this->event->ID
-				)
-			);
-			// phpcs:enable WordPress.DB.DirectDatabaseQuery.DirectQuery
-			// phpcs:enable WordPress.DB.DirectDatabaseQuery.NoCaching
-			// phpcs:enable WordPress.DB.PreparedSQLPlaceholders.UnsupportedIdentifierPlaceholder
-			$data = ( ! empty( $data ) ) ? (array) current( $data ) : array();
+		foreach ( $data as $key => & $value ) {
+			$result = get_post_meta( $this->event->ID, 'gatherpress_' . $key, true );
```
mauteri: We have a helper for prefixing
```
$result = get_post_meta( $this->event->ID, Utility::prefix_key( $key ), true );
```

includes/core/classes/class-event.php
```diff
@@ -443,38 +443,17 @@ public function get_datetime(): array {
...
 
-			set_transient( $cache_key, $data, 15 * MINUTE_IN_SECONDS );
+			if ( ! empty( $result ) ) {
```
mauteri: We have a `datatime` helper to validate since we can't rely on the `datetime` field type from the database anymore.

`if ( ! empty( $result ) && Validate::datetime( $result ) ) {`

includes/core/classes/class-event.php
```diff
@@ -434,7 +434,7 @@ public function get_formatted_datetime(
 	public function get_datetime(): array {
 		global $wpdb;
```
mauteri: `global $wpdb` is no longer needed.

includes/core/classes/class-event.php
mauteri:
```
 * Fetches the event's start and end dates/times (local and GMT) along with
 * timezone information from post meta. Datetime values are validated before
 * being returned.
 ```

Comment (mauteri) [inline includes/core/classes/class-event.php:127]: Add so we can cache results. This gets called a lot.
```
/**
 * Cached datetime data.
 *
 * @since 1.0.0
 *
 * @var array|null
 */
private ?array $datetime_cache = null;
```

File: includes/core/classes/class-event.php [hunk 5]
```diff
@@ -434,7 +434,7 @@ public function get_formatted_datetime(
 	public function get_datetime(): array {
```
mauteri, line 442: Early return if we have results in cache.
```
	if ( null !== $this->datetime_cache ) {
		return $this->datetime_cache;
	}
```

includes/core/classes/class-event.php
```diff
@@ -443,38 +443,17 @@ public function get_datetime(): array {
...
+				$value = $result;
+			}
 		}
 
-		$data = array_merge(
-			$default,
-			(array) $data
-		);
-
 		$data['timezone'] = apply_filters( 'gatherpress_timezone', $data['timezone'] );
```
mauteri, line 474: Save results to cache.
```
$this->datetime_cache = $data;
```

mauteri (changes requested):

mauteri (approved):

PR #1357, opened 2026-01-22
Register commands for new event or venue creation

### LINKED ISSUE #735: NEW Commands for Palette (API)

While working on #629, I mentioned some missing commands for the [Command Palette (API)](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-commands/). 

I'd like to create at least a new venue via `Ctrl + k`

### Designs

https://github.com/user-attachments/assets/d0dca259-f84c-4e9c-b2b6-08c478fe2895

_No response_

### Code of Conduct

carstingaxion:
### Description of the Change

https://github.com/user-attachments/assets/835d77e8-189b-4316-8406-0e20e3cc2615

> ✅ Use dispatch().registerCommand() when hooks aren’t needed
> ✅ Use useCommand() for anything interactive or dynamic
>
> https://juanma.codes/2025/12/08/registering-custom-commands-for-the-wordpress-command-palette-in-different-contexts/#dispatch-registercommand-vs-usecommand

Closes #735 

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:


package.json
```diff
@@ -17,6 +17,9 @@
 		"url": "https://github.com/GatherPress/gatherpress/issues"
 	},
 	"main": "index.js",
+	"dependencies": {
+		"@wordpress/commands": "^1.38.0"
```
mauteri: Let's move this to `devDependencies`.

carstingaxion: Sure, I must have missed --save-dev on this one.
Thanks.

mauteri:
@carstingaxion i assigned this ticket to you so we can possibly get it into 0.35.0, so feel free to proceed. thx!

mauteri:
@carstingaxion heads up, I took this one forward since you mentioned you were short on time, and the approach changed along the way. Wanted to write up what we found so it doesn't look like your work just disappeared.

## What we found

While testing your commands in the palette, this showed up searching "add new":

- **Add new venue** (ours)
- **Add new event** (ours)
- **Go to: Events > Add New Event** (core's)

Core was already shipping an equivalent of one of them. Tracing where that came from: `wp-includes/script-loader.php` builds the palette's "Go to" commands by walking the `$menu` and `$submenu` globals and emitting one command per entry, formatted as `parent > child`. So every admin menu item automatically becomes a command.

That explains why events had a core command and venues did not. `wp-admin/menu.php` only builds the `all_items` plus `add_new_item` submenu pair for post types whose `show_in_menu` is `true`. Ours is a string:

```php
'show_in_menu' => 'edit.php?post_type=gatherpress_event',
```

Post types nested that way go through `_add_post_type_submenus()` instead, which registers exactly one entry, the listing screen, and never an "Add New" one. So the Events menu has "Add New Event" but the venue has no equivalent, and because the palette is generated from that menu, the missing menu item is also a missing command.

In other words the real gap was never in the palette. It was one missing admin menu item, and the palette was just reflecting it.

## What changed

Registering that missing submenu item fixes the admin menu and the palette in one go, and it makes the hand-written commands unnecessary, so those came out along with the `@wordpress/commands` dependency they needed. Venues stay nested under Events.

Two things fall out of doing it at the menu level: the label comes from the post type's `add_new_item`, so a venue post type renamed to "Location" reads "Add New Location" with no extra wiring, and the loop covers anything declaring `gatherpress-venue-information` rather than only the built-in post type.

## The part that got dropped

Your venue command opened `VenueNavigator` in a modal so a venue could be created without leaving the post. The menu-derived command navigates to the new venue screen instead, so that inline flow is gone. It was the more interesting half of your idea and closer to the video on #735, so if you still want it, it is worth its own issue rather than holding this one up. It would sit on top of this fine.

## Branch housekeeping

The branch was six months behind and merging develop surfaced 76 conflicts. Almost all of it was work that has since landed through other PRs (`src/formats/tooltip/`, `VenueNavigator`, `VenuePostsCombobox`, `VenueTermsCombobox` are all on develop now), so conflicts were resolved in develop's favor. I also dropped branch-only leftovers that were not part of this issue: `venue-v2`, `online-event-v2`, `rsvp-cleanup`, `VenuePanelRow`, and their tests and assets. Shout if any of those were still live work and I will get them back out of the history for you.

The PR is down to three files now, all PHP.

One testing note for anyone who picks this up: `Venue\Test_Setup` is `@group multisite`, so it does not run under the standard `phpunit.xml.dist`. The new tests passed locally only because I ran `phpunit-multisite.xml.dist` explicitly, and doing so caught two real failures.

Still yours in the history and on the issue.

PR #1369, opened 2026-02-04
Remove custom text-alignment controls in favor of core controls in event date block

carstingaxion:
The event date block uses a custom block attribute for text alignment, which renders as button inside the block toolbar, but has no effect onto the frontend rendering. The selected block attribute is not transformed into a css class on the frontend.

Instead of fixing this "old way to do text-alignment in blocks", this PR removes all "old-way" in favor of the core block support `textAlign`, which shows as an identical button inside the editor and is properly working on the frontend.

### Description of the Change

~~Closes #~~

### Changelog Entry

> Fixed - Text alignment for the event-date block

### Credits

Props @carstingaxion 

### Checklist:

### DIFF
--- a/src/blocks/event-date/block.json
+++ b/src/blocks/event-date/block.json
@@ -18,9 +18,6 @@
 		"eventStart": {
 			"type": "string"
 		},
-		"textAlign": {
-			"type": "string"
-		},
 		"displayType": {
 			"type": "string",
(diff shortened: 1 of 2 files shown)


mauteri (approved):

mauteri:
Nice catch, @carstingaxion!

Issue #1075, opened 2025-04-27
Topic Listing pages are sorted by Name


apermo:
### Describe the bug

I have 3 topics, a recurring game night. 

Called:

* Game night May
* Game night June
* Game night July

I expect them to be shown in exactly that order

They are shown in Order: June, July, May (which is the alphabetical order)

### Steps to Reproduce

Create the above example, put all in a topic and navigate to the topic page.

### Environment information

Operating System irrelevant

### WordPress information

Latest Plugin and WordPress Version

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

apermo:
I'll check in later and prepare a PR.

apermo:
My first assumption was wrong, but the result is still broken.

It is sorted descending by post date. So the newer an event was published, the higher it will be displayed in the topic archives.

I will create a PR, adding 2 drop downs to the settings page. Allowing to select the ORDERBY and ASC/DESC for the default archives.

mauteri:
I think by default custom taxonomies are ordered alphabetically. I would keep it default if that's the case, but there should also be a filter you can leverage to change the order. Let me know if you see or know something different. Thx!

carstingaxion:
@apermo This is defined in the `sort` attribute for the taxonomy during `register_taxonomy` and can be changed for your site via [`register_taxonomy_args`](https://developer.wordpress.org/reference/hooks/register_taxonomy_args/)

> Setting `sort` to `true` will have WordPress retain the order in which terms are added to objects. Setting `args` lets you specify an array of configuration parameters that, when present, override corresponding parameters in the `$args` array passed to `wp_get_object_terms` when using that function to query for terms in the specified taxonomy. This allows the order of terms to be displayed in the order they were defined in the editor.
>
> https://developer.wordpress.org/reference/functions/register_taxonomy/#comment-2687

mauteri:
Consolidating this into #1379, where the plan for it now lives.

**First, clearing up a mix-up in this thread.** @apermo retracted the title himself an hour after filing: the events aren't in alphabetical order, they're in publish-date order. My reply and Carsten's after that were both about **term** ordering (`register_taxonomy`'s `sort` arg, `register_taxonomy_args`), which is a different knob and doesn't touch this. The bug is about how the **posts** inside a term archive are queried. Sorry for sending you down that path.

[...]

PR #1379, opened 2026-02-13
Fix inherited queries using event-query-block

carstingaxion:
> @kofimokome 
> The events on the page https://gatherpress.org/event/?cst are sorted in ascending order.
> Recent events are on the last page.
>
> *in GatherPress slack*

### Description of the Change

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:

### DIFF
--- a/includes/core/classes/blocks/class-event-query.php
+++ b/includes/core/classes/blocks/class-event-query.php
@@ -103,13 +103,19 @@ class Event_Query {
 				true === $parsed_block['attrs']['query']['inherit']
 			) {
 				global $wp_query;
+error_log( 'Event_Query::pre_render_block - Inheriting query' );
+error_log( 'Event_Query::pre_render_block - Parsed block attrs: ' . print_r( $parsed_block['attrs'], true ) );
+error_log( 'Event_Query::pre_render_block - Original WP_Query args: ' . print_r( $wp_query->query_vars, true ) );	
 
 				$query_args = array_merge(
 					$wp_query->query_vars,
 					array(
-						'posts_per_page' => $parsed_block['attrs']['query']['perPage'],
-						'order'          => $parsed_block['attrs']['query']['order'],
-						'orderby'        => $parsed_block['attrs']['query']['orderBy'],
+						'posts_per_page'          => $parsed_block['attrs']['query']['perPage'],
+						'order'                   => $parsed_block['attrs']['query']['order'],
+						'orderby'                 => $parsed_block['attrs']['query']['orderBy'],
+						'gatherpress_event_query' => $parsed_block['attrs']['query']['gatherpress_event_query'],
+						'include_unfinished'      => $parsed_block['attrs']['query']['include_unfinished'],
+
 					)
 				);
 
@@ -135,6 +141,7 @@ class Event_Query {
 				// "Hijack the global query. It's a hack, but it works." Ryan Welcher.
 				// phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
 				$wp_query = new WP_Query( $filtered_query_args );
+				error_log( 'Event_Query::pre_render_block - Modified WP_Query args: ' . print_r( $wp_query->query_vars, true ) );
 			} else {
 				add_filter(
 					'query_loop_block_query_vars',


mauteri:
@carstingaxion can this one be closed? I added a setting that defaults event archive to upcoming, but can also be set to past and none. I believe this PR was debugging this functionality and related. https://github.com/GatherPress/gatherpress/pull/1587

carstingaxion:
I guess this was and maybe still is related to a yet unset gatherpress_event_query var.
( I wrote a longer comment on our chats, but can't find)

Let's test it together with https://github.com/GatherPress/gatherpress/issues/1075

If 1075 is working fine now, this one can be closed.

mauteri:
It might need to be extended for all event archives (topics, etc) which is fine.

carstingaxion:
> It might need to be extended for all event archives (topics, etc) which is fine.

Maybe even more general: for every query that has a Event supporting type, but not the event-query variation set.

mauteri:
> > It might need to be extended for all event archives (topics, etc) which is fine.
> 
> Maybe even more general: for every query that has a Event supporting type, but not the event-query variation set.

Need to think about that because the Settings are pretty tied to the Event/Venue post types as they appear in the Events admin menu. With other CPTs I think people will be explicitly setting things probably via filter which is also available right now for this.

carstingaxion:
Mentally, I'm not at custom post types, but just at how WordPress core generates the request on a (topic) taxonomy archive, if no custom template is registered.

mauteri:
> Mentally, I'm not at custom post types, but just at how WordPress core generates the request on a (topic) taxonomy archive, if no custom template is registered.

I just don't want it to get too confusing is all. I think we can sort this out nicely though.

mauteri:
@carstingaxion is this PR still relevant? Seems like mostly logging. Curious if we fixed the issue you were seeing in 0.34.0 with some work we did to event query. thx!

carstingaxion:
Testing with the [nightly Playground](https://playground.wordpress.net/?blueprint-url=https://raw.githubusercontent.com/GatherPress/gatherpress/develop/.wordpress-org/blueprints/blueprint-nightly.json):

1. `/topic/gatherpress/` lists 3 events (1 upcoming, 2 past), ordered by publish_date 
(same issue https://github.com/GatherPress/gatherpress/issues/1075)

Ideas:
1. GatherPress could register a block_template for the topic taxonomy programmatically and set the needed parameters for the query block in use.  This will make sure, a topic term archive shows only upcoming events (by default), which can still be adjusted by an editor inside the Site-Editor for that template.
2. OR we could extend the work from #1611 to apply for term archives as well.

mauteri:
> Testing with the [nightly Playground](https://playground.wordpress.net/?blueprint-url=https://raw.githubusercontent.com/GatherPress/gatherpress/develop/.wordpress-org/blueprints/blueprint-nightly.json):
> 
> 1. `/topic/gatherpress/` lists 3 events (1 upcoming, 2 past), ordered by publish_date
>    (same issue [Topic Listing pages are sorted by Name #1075](https://github.com/GatherPress/gatherpress/issues/1075))
> 
> Ideas:
> 
> 1. GatherPress could register a block_template for the topic taxonomy programmatically and set the needed parameters for the query block in use.  This will make sure, a topic term archive shows only upcoming events (by default), which can still be adjusted by an editor inside the Site-Editor for that template.
> 2. OR we could extend the work from [Default post type archive shows all posts regardless of upcoming/past filter on custom event-supporting post types #1611](https://github.com/GatherPress/gatherpress/issues/1611) to apply for term archives as well.

I think we should go with option 2 since we have a setting for Event Archive already and this is an event archive. It's predictable if following that setting.

carstingaxion:
> > OR we could extend the work from Default post type archive shows all posts regardless of upcoming/past filter on custom event-supporting post types #1611 to apply for term archives as well.

>I think we should go with option 2 since we have a setting for Event Archive already and this is an event archive. It's predictable if following that setting.

Ok, as long as we find a nice way to apply this to every registered taxonomy for the gatherpress-event-date supporting post types. Happy to help!

mauteri:
_Edit: I closed this by mistake a minute ago and reopened it. Sorry for the noise @carstingaxion, ignore that. Leaving the technical notes below since they're still useful._

Some notes on where this stands, relative to #1075.

The debugging here did its job. The original symptom, inherited queries not carrying the event query vars, is covered now by the archive mode work in #1611 and the per-`queryId` event type recording added in #1806.

What's left is the thing you re-tested on the nightly on July 25, and I don't think it can be fixed from this file. `Blocks\Event_Query::pre_render_block()` only runs for our own block, but a topic term archive never gets that far. The main query is already built wrong before any block renders.

That lands in `Event\Setup::handle_event_archive_redirect()`, which bails at the top:

```php
if ( ! is_post_type_archive() || ! post_type_supports( $post_type, 'gatherpress-event-date' ) ) {
	return;
}
```

Term archives fall straight through, so `fall_back_to_archive_mode()` never runs and the configured archive mode never applies. Widening that gate is option 2, the one we agreed on above.

One thing that should help with your "as long as we find a nice way to apply this to every registered taxonomy for the `gatherpress-event-date` supporting post types" concern: that resolver already exists in the codebase as `Calendar\Setup::has_post_type_for_taxonomy()`. It loops `get_post_types_by_support( 'gatherpress-event-date' )` against `is_object_in_taxonomy()`. It's `protected` inside the calendar subsystem today, so the work is lifting it somewhere shared rather than writing it from scratch.

Heads up either way that this branch has been conflicting with develop since February.

carstingaxion:
Good to read, that we do not need to take additional care about the supported post types.

@mauteri Could you create a PR with the ideas from Number 2, to widen the scope for all archives of registered taxonomies?

EDIT: Or completely update this one, to do something useful in the end. ;)

mauteri:
@carstingaxion I'll look to get this one in 0.36.0. Thx!

carstingaxion:
Maybe related and solvable in one go: the core/post-pagination blocks do show the previous or next event based on the post publication date, not the event date.

PR #1381, opened 2026-02-15
Return 404 for event archive URL unless matching page exists.

mauteri:
<!--
Please do your best to fill out this template.  Any PR that does not include enough information to be reviewed may be closed at a maintainers' discretion.  All new code ideally includes documentation and tests to ensure against regressions.
-->

### Description of the Change
Visiting the event base URL (e.g., `/event/`) now returns a 404 instead of showing a confusing archive page. If a WordPress page exists with the same slug, that page is served instead.                                                                                                                                          
                                                                                                                                                                   
Feed URLs (`/event/feed/`) continue to work as expected. 
<!--
We must be able to understand the design of your change from this description.  The maintainer reviewing this PR may not have worked with this code recently, so please provide as much detail as possible.

- screenshots or screencasts
-->

### How to test the Change
1. Visit `/event/` (or your configured event base URL) — should return 404                                                                                       
2. Create a page with slug `event` and publish it — visiting `/event/` should now show that page                                                                 
3. Visit `/event/feed/` — should still return the RSS feed    
<!-- Please provide steps on how to test or validate that the change in this PR works as described. -->

### Changelog Entry
<!--
Please include a summary for this PR, noting whether this is something being Added / Changed / Deprecated / Removed / Fixed / or Security related.  You can replace the sample entries after this comment block with the single changelog entry line for this PR. -->
> Fixed - Bug fix

### Credits
<!-- Please list any and all contributors on this PR so that they can be properly credited within this project. -->
Props @mauteri

### Checklist:
<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
<!--- If you are unsure about any of these, please ask for clarification.  We are here to help! -->

### DIFF
--- a/includes/core/classes/class-event-setup.php
+++ b/includes/core/classes/class-event-setup.php
@@ -535,7 +535,6 @@ class Event_Setup {
 		// No page exists with this slug, so trigger a 404.
 		$wp_query->set_404();
 		status_header( 404 );
-		nocache_headers();
 	}
 
 	/**
--- a/includes/core/classes/class-event-setup.php
+++ b/includes/core/classes/class-event-setup.php
@@ -69,6 +69,7 @@ class Event_Setup {
 		add_action( 'init', array( $this, 'register_post_meta' ) );
 		add_action( 'init', array( $this, 'register_calendar_rewrite_rule' ) );
 		add_action( 'parse_request', array( $this, 'handle_calendar_ics_request' ) );
+		add_action( 'template_redirect', array( $this, 'handle_event_archive_redirect' ) );
 		add_action( 'delete_post', array( $this, 'delete_event' ) );
 		add_action( 'wp_after_insert_post', array( $this, 'set_datetimes' ) );
 		add_action( sprintf( 'save_post_%s', Event::POST_TYPE ), array( $this, 'check_waiting_list' ) );
@@ -212,6 +213,9 @@ class Event_Setup {
 					'custom-fields',
 				),
 				'menu_icon'     => 'dashicons-nametag',
+				// Note: has_archive must be true for event feed URLs (/event/feed/) to work.
+				// The archive page itself is handled by handle_event_archive_redirect() which
+				// returns a 404 or redirects to a page with the same slug.
 				'has_archive'   => true,
 				'rewrite'       => array(
 					'slug'       => $rewrite_slug,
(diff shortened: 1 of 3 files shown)


carstingaxion:
I haven't looked into the code yet, but just reading your description makes me feel uncomfortable.

I don't think the default WordPress template hierarchy should be threaded as confusing archive page, I would consider this as the absolute normal default, which should not be changed, without very good reasons. 

What are the reasons for GatherPress?

I guess the original problem comes from the point, that an inherited event query does not know about any of GatherPress' custom query variables. And a default archive template, would use such an default inherited query. Just adding them to the inherited query should fix the problem, from what I understood.

mauteri:
> I haven't looked into the code yet, but just reading your description makes me feel uncomfortable.
> 
> I don't think the default WordPress template hierarchy should be threaded as confusing archive page, I would consider this as the absolute normal default, which should not be changed, without very good reasons.
> 
> What are the reasons for GatherPress?
> 
> I guess the original problem comes from the point, that an inherited event query does not know about any of GatherPress' custom query variables. And a default archive template, would use such an default inherited query. Just adding them to the inherited query should fix the problem, from what I understood.

@carstingaxion So the issue is that a query for upcoming events and a query for past events are different as one is asc and the other is desc with dateime either before now or after now. So there's not straight forward query for a general archive of events, unlike pretty much any other post type one could create which is based off the publish date. The reason to keep it true is due to the feeds, which have feeds for upcoming and past events. By setting this to 404 we are basically saying, this doesn't exist, which is fine as we already have archive settings in the Settings section of the plugin where you can dedicate a page to Upcoming or Past events (for classic themes). It also allows you to create a page here instead with this slug and do whatever you want. It is also a hook, so if a plugin doesn't want this at all, they can remove the hook and manipulate the query for whatever they like. In my mind, if there is nothing to show of value here out the box, there's no reason to resolve the URL to some weird archive.

I add a lot of the reasons in the code, so it's documented throughout.

carstingaxion:
I think it's absolutely wrong that you _just_ deactivated an ancient WordPress standard – the `post_type` archive!

image

mauteri:
> I think it's absolutely wrong that you _just_ deactivated an ancient WordPress standard – the `post_type` archive!
> 
> image

@carstingaxion You don't have to agree with me. I spoke my piece and gave all the reasons behind it. If you don't like it, unhook the action in a plugin and do what you like. Or provide a PR with good reasoning behind the query that makes sense for events.

PR #1383, opened 2026-02-15
Protect GatherPress meta from Custom Fields panel to prevent stale data overwrites.

mauteri:
### Description of the Change
When Custom Fields are enabled in the block editor, saving an event triggers two saves - one from JavaScript with updated data, then one from the Custom Fields metabox with stale data loaded at page load. This overwrites datetime changes.                                                                                   

Fix: Mark all `gatherpress_` meta keys as protected so they don't appear in Custom Fields. 

### How to test the Change
1. Edit an event or venue and open Preferences (⋮ menu → Preferences)                                                                                                     
2. Enable "Custom Fields" under Panels                                                                                                                           
3. Change the event end time and save                                                                                                                            
4. Reload the page — the end time change should persist                                                                                                          
5. Verify `gatherpress_*` fields don't appear in Custom Fields panel  

### Changelog Entry

> Fixed - Bug fix

### Credits

Props @mauteri @carstingaxion 

### Checklist:

### DIFF
--- a/.github/workflows/playground-preview.yml
+++ b/.github/workflows/playground-preview.yml
@@ -62,15 +62,10 @@ jobs:
         with:
           php-version: latest
           coverage: none
+          tools: wp-cli
 
-      - name: Install WP-CLI
-        run: |
-          curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar
-          chmod +x wp-cli.phar
-          sudo mv wp-cli.phar /usr/local/bin/wp
-
-      - name: Install latest version of dist-archive-command
-        run: wp package install wp-cli/dist-archive-command:dev-main
+      - name: Install dist-archive-command
+        run: wp package install wp-cli/dist-archive-command:^2
 
       - name: Setup Node
         uses: actions/setup-node@v4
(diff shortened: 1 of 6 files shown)


carstingaxion (commented):
Looking at the [used by](https://developer.wordpress.org/reference/functions/is_protected_meta/#related) section for this filter, makes me wonder if this is maybe to rigid, to protect all "gatherpress_" meta fields. 

But in general, I absolutely like the way, this looks clean and straightforward.

mauteri:
> Looking at the [used by](https://developer.wordpress.org/reference/functions/is_protected_meta/#related) section for this filter, makes me wonder if this is maybe to rigid, to protect all "gatherpress_" meta fields.
> 
> But in general, I absolutely like the way, this looks clean and straightforward.

I was thinking about that too, but in the end, I can't think of a meta field we'd put out that'd we'd want people to use the Custom Fields metaboxes or any other alternative ways to make changes to; we'd always want our UI to make changes to our meta fields. This also makes us not have to manage any new meta fields and adding it to a list.

If it becomes too much of an issue, we can always extend it with a filter to bypass some `gatherpress_*` meta, but make it opt-in to bypass it.

PR #1414, opened 2026-03-14
fix(ics): correctly escape newlines without mutating the letter 'n'

### LINKED ISSUE #1413: [Bug]: ICS Export incorrectly escapes every lowercase "n" as a newline "\n"
### Describe the bug

Hello GatherPress team! 👋 First off, thank you for the amazing work on this plugin. 

I've come across a text-formatting bug related to the `.ics` calendar export functionality. It appears that every instance of the lowercase letter `n` in the event text (like the description or location) is being incorrectly escaped/replaced with `\n`. This looked very weird in the calendar.

### Steps to Reproduce

1. Create a GatherPress event and include words in the description or location containing the lowercase letter "n" in the description (e.g., `"Veranstaltung"`, `"in"`, or `"Berlin"`).
2. Export/download the `.ics` file for this event.
3. Open the downloaded `.ics` file in a text editor or import it into a calendar application.
4. Notice that the letter `n` has been mutated into `\n`.

_No response_

### Environment information

- GatherPress version: 0.33.3
- WordPress version: 6.9
- Theme: twenty-twenty-five

### WordPress information

_No response_

### Code of Conduct

anzzyspeaksgit:
## Problem
The ICS export logic used `addcslashes( $text, ",;\\n" )` to escape content. In PHP, the double-quoted string `",;\\n"` evaluates to the characters `,`, `;`, `\`, and the literal letter `n`. As a result, `addcslashes` incorrectly escaped every lowercase 'n' character into `\n`, rather than escaping actual line feeds.

## Solution  
Replaced `addcslashes` with a strict `str_replace` sequence that explicitly matches actual newline (`\n`), carriage return (`\r`), comma, semicolon, and backslash characters and encodes them into their literal ICS counterparts (e.g. `\\n`, `\\,`). Carriage returns are stripped out to avoid trailing `\r` anomalies.

## Testing
Verified string replacement rules match RFC 5545 specifications for TEXT fields. A text containing the letter 'n' will now remain untouched while true line-breaks are converted safely to `\\n`.

Closes #1413

### DIFF
--- a/includes/core/classes/class-event.php
+++ b/includes/core/classes/class-event.php
@@ -737,7 +737,7 @@ class Event {
 	 * @return string The escaped text suitable for ICS content.
 	 */
 	protected function escape_ics_text( string $text ): string {
-		return addcslashes( $text, ",;\\n" );
+		return str_replace( array( "\\", ",", ";", "\n", "\r" ), array( "\\\\", "\\,", "\\;", "\\n", "" ), $text );
 	}
 
 	/**


includes/core/classes/class-event.php
```diff
@@ -737,7 +737,7 @@ public function get_ics_download_link(): string {
 	 * @return string The escaped text suitable for ICS content.
 	 */
 	protected function escape_ics_text( string $text ): string {
-		return addcslashes( $text, ",;\\n" );
+		return str_replace( array( "\\", ",", ";", "\n", "\r" ), array( "\\\\", "\\,", "\\;", "\\n", "" ), $text );
```
mauteri, line 740: Thanks for this! 

Could we update/add unit test for this to ensure it is fixed and that nothing else broke? If you need help let me know.

mauteri:
This ended up being address in this [PR](https://github.com/GatherPress/gatherpress/pull/1416).

PR #1464, opened 2026-04-15
Support interactive clientNavigation for the venue block

carstingaxion:
### Description of the Change

| BEFORE | AFTER |
|--------|--------|
| image | image |

 ~~Closes #~~

### Changelog Entry

> Added - Support interactive clientNavigation for the venue block

### Credits

Props @carstingaxion 

### Checklist:

### DIFF
--- a/src/blocks/venue/block.json
+++ b/src/blocks/venue/block.json
@@ -23,7 +23,10 @@
 			"postIdOverride": true
 		},
 		"align": [ "wide" ],
-		"html": false
+		"html": false,
+		"interactivity": {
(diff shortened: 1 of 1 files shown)


mauteri (approved):

mauteri:
@carstingaxion looks good, thanks! I assume we'll need to add support to other blocks? When testing with a Query Loop that includes all the GatherPress blocks (event-date, venue, rsvp, icons, modals, etc.), the "incompatible blocks" warning still appears because the venue block isn't the only one missing `clientNavigation`.

PR #1467, opened 2026-04-16
Extend admin list table columns, sorting, and views to all event post types.

mauteri:
### Description of the Change

Previously, `Event_Admin_List` hardcoded `Event::POST_TYPE` (`gatherpress_event`) throughout — in hook registration, sort detection, view filters, event counts, venue taxonomy resolution, and RSVP column links. This meant any custom post type using the `gatherpress-event-date` support would not get the custom admin list columns, sortable columns, Upcoming/Past view filters, or correct RSVP links.

This PR refactors `Event_Admin_List` to work dynamically for all post types with `gatherpress-event-date` support:

- `setup_hooks()` now registers only global hooks and defers per-post-type hook registration to a new `register_post_type_hooks()` method called at `init` priority 11, after all custom post types have been registered
- `default_sort()` and `handle_column_sorting()` use `post_type_supports()` instead of a hardcoded post type comparison
- `views_edit()` derives the current post type from `get_current_screen()->post_type` and passes it to `get_event_counts()`
- `get_event_counts()` now accepts a `$post_type` parameter and caches results per post type
- `venue_sorting_join_paged()` derives the venue taxonomy from the current query's post type via `$wp_query->get('post_type')`
- `custom_columns()` RSVP links use `get_post_type($post_id)` rather than the hardcoded constant

### How to test the Change

Register a custom post type with `gatherpress-event-date` support in a mu-plugin or test plugin, then navigate to its list table in the WordPress admin. Verify the **Event date & time**, **Venue**, and **RSVPs** columns appear, the **Upcoming** and **Past** view filters are present, and column sorting works. Confirm the same behavior is unchanged on the standard Events list table at `edit.php?post_type=gatherpress_event`.

### Changelog Entry

Changed - Admin list table columns, sorting, and Upcoming/Past view filters now apply to all post types with `gatherpress-event-date` support, not just the built-in `gatherpress_event` post type.

### Credits

Props @mauteri

### DIFF
--- a/includes/core/classes/class-event-admin-list.php
+++ b/includes/core/classes/class-event-admin-list.php
@@ -35,13 +35,13 @@ class Event_Admin_List {
 	use Singleton;
 
 	/**
-	 * Cached event counts for the current request.
+	 * Cached event counts keyed by post type for the current request.
 	 *
 	 * @since 1.0.0
 	 *
-	 * @var array<string, int>|null
+	 * @var array<string, array<string, int>>
 	 */
-	protected ?array $event_counts = null;
+	protected array $event_counts = array();
 
 	/**
 	 * Class constructor.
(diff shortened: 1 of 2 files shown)


includes/core/classes/class-event-admin-list.php
```diff
@@ -67,30 +67,47 @@ protected function setup_hooks(): void {
...
-		add_filter(
-			sprintf( 'manage_edit-%s_sortable_columns', Event::POST_TYPE ),
-			array( $this, 'sortable_columns' )
-		);
-		add_filter(
-			sprintf( 'views_edit-%s', Event::POST_TYPE ),
-			array( $this, 'views_edit' )
-		);
 		add_filter( 'query_vars', array( $this, 'query_vars' ) );
+		add_action( 'init', array( $this, 'register_post_type_hooks' ), 11 );
```
carstingaxion, line 71: Im unsure about this 11, I would have done this on 10 at the registered_{$post_type} hook instead.

mauteri: @carstingaxion thanks, this was done at 11 to ensure all post types with the support are registered prior to receiving the event admin appearance. That said, let me leave a comment explaining the change in priority.

mauteri: https://github.com/GatherPress/gatherpress/pull/1474/changes#diff-a238fca21df3c1e5aa87ea69d9c583606a7b4e5b525a4db33c61a4a5f1d1caa3R71

carstingaxion: @mauteri Thanks for your reply and the PR.
While it is best-practice that post types are registered on init:10, it may not be the case everywhere.

That's why I suggested to not use init at all, but https://developer.wordpress.org/reference/hooks/registered_post_type_post_type/ (I should have linked that in the first place to avoid misunderstanding, sorry)

mauteri: I'll take a look, thx

mauteri: Okay, fixing this in another PR. Sorry, I missed the second part of your comment with the proposed solution.

PR #917, opened 2024-09-25
Share Options across Multisite

carstingaxion:
Currently GatherPress saves and uses its settings exclusively per blog. 

### Description of the Change

This PR adds the ability to share Settings across a network of sites. _(In WordPress terminology: ... across multiple blogs of a site)_

#### NEEDED BEFORE MERGE

- [ ] Find a way to handle (separate, deprecate, ... ?) the event-pages options. This pair of settings is the only, that should not be inherited or handled on the network level at all
- [x] Decide what to do with the admin-settings-pages! Show inherited options as read-only, or hide completely in the UI?
- [ ] Add a network settings UI to enable or disable the feature (within this PR, this is currently done with `add_filter`)

https://github.com/user-attachments/assets/3a5ef773-072d-465b-bf90-b42e299e044d

~~Closes #~~

### How to test the Change

1. ...

### Changelog Entry

> Added - Share (or: inherit, set) the Settings across a whole WordPress Multisite

### Credits

Props @carstingaxion 

### Checklist:

- [x] I agree to follow this project's [**Code of Conduct**](https://github.com/GatherPress/gatherpress/blob/main/CODE_OF_CONDUCT.md).
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my change.
- [x] All new and existing tests pass.

PR #1500, opened 2026-04-19
Add network-wide GatherPress settings and per-option inheritance

mauteri:
### Description of the Change

Adds network-wide GatherPress settings for WordPress multisite, plus a handful of multisite-adjacent bug fixes surfaced while building it.

**Network-wide settings (main objective)**

- New `Network Admin → Settings → GatherPress` page that mirrors the per-site Settings (Events, Venues, Roles, RSVP, Credits) under a tabbed UI plus a dedicated **Network** tab.
- The Network tab holds the inheritance allowlist: a master toggle and a per-option checklist grouped by sub-page / section. Select-all and per-group select/deselect shortcuts included.
- Values set on the network admin tabs save to the network-wide `gatherpress_settings` site option via `update_site_option`. A `pre_option_gatherpress_settings` filter scoped to the network admin screen (via `load-{hook}`) routes reads transparently so the existing Settings class is unchanged.
- Subsite resolver: `Settings::get( $option )` checks `is_option_inherited( $option )` and, when true, reads from the site option instead of the blog option. An escape-hatch filter `gatherpress_network_is_option_inherited( $inherited, $option, $blog_id )` lets a companion plugin exempt a specific site.
- Subsite UX when an option is locked:
  - The field is rendered disabled + dimmed (`.gatherpress-field-inherited` wrapper, native `disabled` attribute on `input`/`select`) and shows the current network value.
  - Per-field note underneath. Super admins get `"Inherited from the [network]. Edit there to change this value."` with a link; regular site admins get `"Inherited from the network."` (no dead-end link).
  - Page-level admin notice on any GatherPress settings page. Everyone sees the first sentence; only users with `manage_network_options` see the appended `"Locked fields can only be changed from [Network Admin → Settings → GatherPress]."` sentence.
- `manage_network_options` capability gate on the network admin page; saves run through dedicated `network_admin_edit_*` handlers with per-action nonces.

**Multisite bug fixes landed along the way**

- **Missing `{prefix}gatherpress_events` table on subsites.** `on_site_create` was gated on `is_plugin_active_for_network()`, which can be unavailable depending on the context that creates the site (CLI, REST, early hooks). Now loads `wp-admin/includes/plugin.php` when needed and uses `plugin_basename( GATHERPRESS_CORE_FILE )` instead of a hard-coded path. `check_plugin_version` also auto-heals any subsite where the `gatherpress_version` option doesn't match current, running `create_tables()` idempotently via `dbDelta`.
- **`DateTimeZone('UTC+0')` 500 on save.** `Utility::get_system_timezone()` used to emit invalid WP-style strings like `UTC+0` / `UTC+5`. Rewrote it to return a PHP-valid identifier via a new `Utility::offset_to_timezone_string()`. Added `Utility::normalize_timezone_string()` to sanitize any legacy values already stored in event meta, and wrapped the `DateTimeZone` constructor calls in `class-event.php` through it.
- **`Moment Timezone has no data for UTC+0` console spam in the block editor on a fresh WordPress install.** Added an inline shim (`Assets::enqueue_timezone_shim()`) that runs after `wp-date`'s `setSettings` call and rewrites `UTC+0` / `UTC-0` → `UTC` before Gutenberg's `panels.js` reads it.

Closes #

### How to test the Change

Multisite verification:

1. On the main site, visit **Network Admin → Settings → GatherPress**. Confirm the tab row (Events / Venues / Roles / RSVP / Credits / Network), that Tools is omitted, and that the sidebar `GatherPress` link lands you on the Network tab.
2. On the Network tab, toggle `Enable` and check a handful of options (e.g. `date_format`, `show_timezone`). Use the top-of-page `Select all` / `Deselect all` and the per-sub-page toggles to verify the bulk helpers. Save.
3. On one of the other tabs, set a distinctive value (e.g. custom `Date Format`). Save.
4. Visit a subsite's `Events → Settings` page as a super admin:
   - Notice banner appears with both sentences and a link to network admin.
   - The fields you marked inherited are disabled, dimmed, showing the network value, with the per-field `Inherited from the [network]…` link.
   - Fields you didn't mark remain editable.
5. Log in to the same subsite as a regular site admin:
   - Notice banner shows only the first sentence (no network link).
   - Per-field notes show plain `"Inherited from the network."`.
6. Register a companion-plugin filter for `gatherpress_network_is_option_inherited` that returns `false` for one `(blog_id, option)` pair. Confirm that field renders fully editable on that site.

Multisite bug verification:

1. Create a new subsite (or visit a subsite whose `wp_<n>_gatherpress_events` table is missing). Load the admin at least once and confirm the table is created by the auto-heal. Event list now shows events (previously the JOIN against the missing table made the list appear empty while counts still showed rows).
2. With the main site set to Settings → General → `UTC+0`, save an event. Confirm no `DateTimeZone::__construct` 500 and no console warnings from `panels.js` / `date.min.js`.
3. Run Alpha's **Apply Updates** at network admin — confirm it iterates each site and completes without a fatal.

### Changelog Entry

> Added - Network-wide GatherPress settings with per-option inheritance control at `Network Admin → Settings → GatherPress`.
> Fixed - Missing `gatherpress_events` table on multisite subsites now auto-heals on admin load.
> Fixed - Saving events with a `UTC+0` / `UTC+N` timezone no longer raises a PHP fatal.
> Fixed - `Moment Timezone has no data for UTC+0` warning from the block editor on fresh WordPress installs.

### Credits

Props @mauteri, @carstingaxion (inspiration + messaging from [#917](https://github.com/GatherPress/gatherpress/pull/917)).

### Checklist:

### DIFF
--- a/docs/developer/settings/README.md
+++ b/docs/developer/settings/README.md
@@ -116,3 +116,4 @@ wp gatherpress settings import settings.json --apply --mode=replace
 
 - [Extending Settings](extending-settings.md) -- How to add custom tabs, sections, and fields
 - [Architecture](architecture.md) -- Storage, sanitization, defaults, and caching internals
+- [Multisite Network Settings](multisite.md) -- Network-wide values and per-option inheritance on WordPress Multisite
--- a/docs/developer/settings/multisite.md
+++ b/docs/developer/settings/multisite.md
@@ -7,9 +7,9 @@ On a WordPress Multisite install, GatherPress can expose settings at the network
 Two separate surfaces render the Settings form:
 
 - **Per-site** (unchanged) — `Dashboard → Events → Settings` on each site. Writes to the site's own `gatherpress_settings` blog option.
-- **Network-wide** (new) — `Network Admin → Settings → GatherPress`. Renders the same Events / Venues / Roles / RSVP / Credits tabs plus an additional **Network** tab. Writes to the network-wide `gatherpress_settings` *site* option via `update_site_option()`. Requires the `manage_network_options` capability.
+- **Network-wide** (new) — `Network Admin → Settings → GatherPress`. Renders the same Events / Venues / Roles / RSVP / Credits / Tools tabs plus an additional **Network** tab. Writes to the network-wide `gatherpress_settings` *site* option via `update_site_option()`. Requires the `manage_network_options` capability.
 
-The Tools tab (import / export) stays per-site only — import / export is blog-scoped.
+The Tools tab (import / export) is available at both scopes. In per-site admin it operates on the blog option; in network admin it reads and writes the network-wide site option. A hidden `scope` value on the Tools form (and in the `gatherpress_export_settings` / `gatherpress_import_settings` AJAX payload) tells the handler which store to target, and capability is gated per-scope (`manage_options` for blog, `manage_network_options` for network).
 
 ## The Network Tab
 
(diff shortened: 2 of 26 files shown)


carstingaxion:
Using disabled attribute on input and select fields may result in problems during save, because usually those values are not POSTed. 
Consider using readonly attribute instead.

Issue #1261, opened 2025-12-07
Rethink RSVP & the RSVP block


Menrath:
Currently different RSVP types are hard-coded, that is currently `user` and `email`. There is no real choice to allow or disallow certain RSVP types for an event, or in general. It is "managed" via which block is added to an event. Therefore, the current RSVP block is heavily tied to the "RSVP as WordPress user type".

### Designs

**Idea:**

1. Option to disable RSVP management (in general or for a specific event).
2. Option to choose available RSVP-types as admin setting (type `user` likely cannot be deactivated).
3. Optional: Maybe allow override to disable RSVP types per event.
4. Have a generic RSVP block that is pre-configured to handle all active RSVP types.

The block should offer good enough filters for companion plugins to customize it's appearance.

For core GatherPress without companion plugins this would mean:

* For logged in users the behaviour has not changed.
* For non-logged-in users the RSVP-button would offer both the "login?!" and the rsvp via email option

Companion plugins like gatherpress-activitypub could add a third option "RSVP via the Fediverse" etc.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

carstingaxion:
I would also love to get a way to disable all RSVP related stuff all at once! 

Suggested this also in #688

mauteri:
RSVP Form is self-validating, don't include it and you can't RSVP with it. RSVP requires being logged in and a block for the most part (or some trickiness with an API and valid nonce). That said, it's more important when include in a FSE template and being able to enable/disable RSVP with some front end disabled state if blocks remain present.

carstingaxion:
❓What do you mean?

carstingaxion:
> ... don't include the RSVP blocks for an event. But as we move to people using these blocks in FSE templates where the blocks are shared across events, then a plan for disabling is more important.

Ok ... and what do you suggest @mauteri, I should tell my editors about the remaining RSVP admin menu item and the RSVP admin column?

Menrath:
Maybe it helps taking the following features into account:

1. No RSVP management for this event
  a. show a reason why
  b. don't show a reason why
2. RSVP managed, but externally: show link and maybe info text.

So in this the RSVP block would not be desicive for whether RSVP is enabled but for how it is managed and the information about it.

It seems to be really hard to make the right choices for what should be part of "What you see is what you get", and what should be an option in the sidebar. Are there any references for best practice guidelines for WordPress available?

kofimokome:
[...]

PR #1509, opened 2026-04-21
Ideas on Generic RSVP

Menrath:
I just share this and want to open more publicly on how RSVP could be improved. This is heavily drafted. Not ready for review or more. Just for interested folks.


Menrath:
@mauteri I just reviewd your recent PR #1468 and #1469. This reminded me of where I got lost back then in other projects. I already have one follow up question: What do you think of replacing the post_meta

*`gatherpress_enable_rsvp`
* `gatherpress_enable_open_rsvp`

with more generic ones?

There would be one decision to be made at first: That is if RSVP is enabled, should that mean that RSVP via local WordPress user MUST always be enabled too. That means, that the "RSVP-as-a User-RSVP-Type" is always enabled when RSVP is enabled for an event. (Maybe also @patriciabt has an opinion here?)

Depending on that answer one could just use a single post-meta key. Which could also hold multiple values (which might indeed be handier than managin a list of comma-seperated rsvp-type-keys.

`gatherpress_enable_rsvp` ... could then be more than a boolean and holding the keys of rsvp-types.

@mauteri You also once mentioned managing RSVP-Types via blocks, or their presence. Is this idea rather abandoned now? A con this would have had is that adding new RSVP-Types in the future would be more difficult for already created events.

carstingaxion:
I really liked your idea from a comment in #1261

> 1. No RSVP management for this event
a. show a reason why
b. don't show a reason why
>2. RSVP managed, but externally: show link and maybe info text

Could this kind of externally managed state also be a RSVP Type following your scheme here?

mauteri:
@Menrath I fixed some merge conflicts from some work I did for you. So be sure to pull down when you continue to work on this. Thx!

Menrath:
@mauteri Thanks for moving the all RSVP related classes to their own namespace, I was about to ask whether we want to do this, but you just did it! :)

Menrath:
> Could this kind of externally managed state also be a RSVP Type following your scheme here?

It depends on how you mean that, i.e. to what depth integration is taking place. What information should be shared via the external provider and GatherPress, just an URL? or maybe more like, number of participants or even information about attendees themselves?

### The current architecture

Currently the save function has rather a lot of arguments:

```php
public function save(
  mixed $identifier,
  string $status,
  int $anonymous = 0,
  int $guests = 0,
  string $rsvp_type = 'user'
):
```

Because of this I asked myself, whether we rather should use objects that are passed to the RSVP->save() function. 
As a consequence the RSVP-Type classes could be written differently. To register RSVP-Types more WordPress style static hooks could be used, and the Singleton coding style could be removed from them.

And the responsibilities of the save function should be reduced to 

1. validate and RSVP (which can be delegated to private functions, waiting-list? anonymous-allowed?, etc.)
2. persist and cache it.

includes/core/classes/class-setup.php
```diff
@@ -34,13 +34,28 @@ class Setup {
...
+		 * Fires once GatherPress has finished bootstrapping its core classes.
+		 *
+		 * Subsystems use this to run setup work that depends on other
+		 * GatherPress classes already being instantiated — for example, the
+		 * RSVP type registry registers its core types here so companion
+		 * plugins can hook into `gatherpress_register_rsvp_types` afterwards.
+		 *
+		 * @since 1.0.0
+		 */
+		do_action( 'gatherpress_loaded' );
```
mauteri: This might make more sense as final thing to fire on `gatherpress.php` so it's always clear there that nothing should exist after we fire this and that GatherPress is fully loaded.

Menrath:
These days I started of separating domain logic of hows RSVP-Responses are stored and retrieved and the business logic how they are applied and processed in the context of an event. Two big pieces are still missing in the current state:

1. The Rsvp class is about to be cleaned up at lastly.
2. Renaming the Rsvp_Query to rather Repository, or creating an additional Repository which also does saving/persiting.

I'll try to keep the current public entry points of Rsvp for now. (save(), get()) , but they will just be wrappers for the new API I am designing. I hope that this makes everything much easier to read and follow and mantain in the future.

docs/developer/custom-url-endpoints/README.md
```diff
@@ -122,7 +122,7 @@ Example for a new redirection endpoint like `example.org/event/my-sample-event/o
 - ### 2. Define the callback for the endpoint
 
 	```php
-	use GatherPress\Core\Event;
+	use GatherPress\Core\Event\Event;
```
mauteri: This can stay `GatherPress\Core\Event` as there is an alias for it.

includes/core/classes/event/class-rest-api.php
```diff
@@ -787,14 +787,15 @@ public function update_rsvp( WP_REST_Request $request ): WP_REST_Response {
 			! $event->has_event_past()
 		) {
 			if ( 'attending' !== $status ) {
-				$guests = 0;
+				$guests = 0; // ToDo: is this intended?
```
mauteri: Yes, if you change your attendance status from attending, it removes any guess you said you were bringing.

Menrath:
@carstingaxion 

> Could this kind of externally managed state also be a RSVP Type following your scheme here?

No. Unless there is a 1:1 api for each attendee, no. But this should be not that hard to implement seperately.

includes/core/classes/settings/class-roles.php
```diff
@@ -136,7 +136,7 @@ public function get_user_role( int $user_id ): string {
 			}
 
 			foreach ( json_decode( $users ) as $user ) {
-				if ( intval( $user->id ) === $user_id ) {
+				if ( \intval( $user->id ) === $user_id ) {
```
mauteri: No need for `\` let's revert this. Also not sure how this is related to PR.

includes/core/classes/venue/map/provider/class-osm.php
```diff
@@ -236,7 +236,7 @@ private function paint_tile(
 		$dst_y = $ty * self::TILE_SIZE - $top_pixel;
 
 		imagecopy( $canvas, $tile, $dst_x, $dst_y, 0, 0, self::TILE_SIZE, self::TILE_SIZE );
-		imagedestroy( $tile );
+		unset( $tile );
```
mauteri, line 244: Not sure how this is related to PR and why it changed.

includes/core/classes/venue/map/class-map.php
```diff
@@ -1245,7 +1245,7 @@ protected function ensure_descriptor_for_combo(
 			}
 
 			$url = $this->save_image( $image, $address, $zoom, $width, $height, 1, $slug );
-			imagedestroy( $image );
+			unset( $image );
```
mauteri, line 1173: Not sure how this is related to PR and why it changed.

includes/core/classes/venue/map/class-map.php
```diff
@@ -1276,7 +1276,7 @@ protected function ensure_descriptor_for_combo(
 					self::RETINA_DENSITY,
 					$slug
 				);
-				imagedestroy( $image_2x );
+				unset( $image_2x );
```
mauteri, line 1206: Not sure how this is related to PR and why it changed.

mauteri:
@Menrath probably best to keep working on this but aim to get it into GatherPress 0.35.0 rather than 0.34.0 since we want to release that soon and this is a huge change.

mauteri:
Synced this branch with develop (149 commits — the back half of the 0.34.0 cycle plus the release) so the Generic RSVP exploration stays alive against current code. What was done, in three commits so the merge and the fixes are distinguishable:

1. **Merge commit** — five conflicts resolved:
   - `rsvp/class-rsvp.php` (`check_waiting_list()`): kept this branch's repository/`Intent`/`State` promotion loop. develop's conflicting hunk was the #1771 fix (Open RSVP promotions dropped by the empty-identifier guard), but this design promotes by **comment ID** via the repository, so that failure mode doesn't structurally apply — noted in a code comment.
   - `rsvp/class-query.php`: took develop's comment-query exclusion shape (`type__in`/`type__not_in` empty arrays) and dropped this branch's `post_type => Event::POST_TYPE` restriction, which would fight the post-type-supports work (RSVPs now attach to any `gatherpress-rsvp` post type).
   - `rsvp/class-list-table.php`: combined — kept this branch's new **Type** column and develop's dynamic, post-type-aware Event column label.
   - `gatherpress.php` / playground preview script: current version string; the PHP 8.1 floor this branch anticipated has meanwhile landed on develop for 0.35.0 (#1602), so that assumption now holds for real.
2. **Cleanup** — phpstan caught develop's token-confirmation flow still invalidating the response cache via `Rsvp::CACHE_KEY`, which this branch relocates to the dedicated `Rsvp\Cache` class; `Token` now calls `Cache::delete()`. Also removed a dead import.
3. **Test fix** — the two token cache-invalidation tests from develop updated to seed via `Cache::CACHE_KEY`.

Verification: full PHPUnit suite green (**1,524 tests, 6,529 assertions**), phpstan clean at level 5 / PHP 8.1, phpcs clean.

@Menrath no functional changes to your design were intended beyond the above — flag anything that misreads your intent.

mauteri:
Follow-up push per @mauteri: the five aliased classes (`Event`, `Rsvp`, `Venue`, `Venue\Map`, `Calendar`) now use their `register-class-aliases` short-form imports throughout the files this branch touches — the alias layer (and its phpstan stubs) exists precisely so code can read this way. One deliberate exception, commented in place: the prior-FQN regression test keeps the deep import because it asserts `Rsvp::class` equals the real class name, and `::class` on an alias import evaluates to the alias string — the one context where the two forms aren't interchangeable. Full suite still green (1,524 tests), phpstan and phpcs clean.

mauteri:
I pushed a round of maintenance commits to this branch to bring it current and tighten it up against our standards. Summary of what changed:

**Sync with develop** (48253493): merged in ~149 commits of develop. Notable conflict resolutions: the Token cleanup now goes through the new `Rsvp\Cache::delete()` (ce4e68f8), and tests were updated for the relocated `CACHE_KEY` (facbe1b8).

**Import style** (5c103f1c): switched deep imports like `use GatherPress\Core\Event\Event;` to the registered class aliases (`use GatherPress\Core\Event;`) across the branch. One deliberate exception: `class-test-rsvp.php` keeps the deep import because `Rsvp::class` on an alias yields the alias string, not the real FQN, which would break the identity assertion there.

**Standards pass** (305a1c42, d9d2bc7f): removed backslash prefixes from global function calls, normalized the ABSPATH guards, and completed docblocks across the whole `rsvp/` subsystem to the canonical shape with `@since 0.35.0`. `phpcs`, `phpstan`, and the full PHPUnit suite (1524 tests) are all green.

**Repository rework** (54d67ccb): converted the six private static helpers in `Rsvp\Repository` (`hydrate`, `hydrate_data`, `get_identity_from_comment`, `get_identity_provider`, `get_status`, `get_value_from_object_terms`) to instance methods. Nothing outside the class calls them, the call sites already mixed `$this->` and `self::`, and instance methods keep the door open for injected state and simpler test doubles. Also removed the unused `$default_args` property, restructured `save()` down to three returns, and fixed the `Status::TryFrom` casing.

**One real finding to flag** (documented in code, behavior intentionally left unchanged): the provider `tax_query` clause in `Repository::get_provider_query_args()` is currently a silent no-op. `WP_Tax_Query` skips flat clause arrays; clauses must be nested one level deeper. I tried fixing the nesting, and it surfaced two coupled gaps:

1. `Repository::save()` never assigns the provider term to the comment, so nothing ever matches a strict provider filter.
2. Even after persisting provider terms on save, any RSVP created before that change (including all existing installs after upgrade) has no provider term, so provider-filtered lookups miss them and an update turns into a duplicate insert.

So making the filter real needs a design decision: persist provider terms on save plus either a migration/backfill for existing RSVP comments or a query that also matches comments with no provider term (treating "no term" as the default user provider, matching the hydration fallbacks). Happy to discuss which way you want to take it, @Menrath.

mauteri (approved):

Menrath:
@mauteri Huge thanks for taking the time to fine-tune and review this, and for helping get the PR into a mergeable state! Really appreciate the effort!
I hope that this is not only handy for companion plugins but also good for the maintenance of GatherPress!

mauteri:
@Menrath No problem! If you see any issues or have questions with the final implementations just let me know :-)

PR #1565, opened 2026-04-30
Block transform: core/post-date → gatherpress/event-date

### LINKED ISSUE #1562: Register block transform: core/post-date 2 gatherpress/event-date

I found myself in the situation clicking the „Transform to …“ button on the post-date block more often and everytime I think: Damn, there is no transform registered here.

Please, let’s fix this.

### Designs

_No response_

_No response_

### Code of Conduct

mauteri:
### Description of the Change

Adds a block transform from `core/post-date` to `gatherpress/event-date`. The transform is wired in two places:

- **Block toolbar** — appears in the standard "Transform to" menu (the leftmost block-type icon) on any `core/post-date` block.
- **Inspector sidebar** — a "Convert to Event Date" button with a short description, gated to post types that support `gatherpress-event-date` so it only surfaces where the resulting block has a datetime source.

Attribute mapping: `format` → `startDateFormat`, `textAlign` carries over, `displayType` defaults to `'start'` (post-date is single-date). The sidebar button reuses the registered transform via `switchToBlockType`, so the mapping lives in one place.

Closes #1562

### How to test the Change

1. Edit any event post.
2. Insert a Post Date block.
3. Toolbar path: click the block-type icon (leftmost in the toolbar) → "Transform to" → **Event Date**.
4. Sidebar path: with the post-date block selected, look in the right sidebar for the "Convert to Event Date" button.
5. Verify the resulting Event Date block carries the date format and text alignment from the original.
6. Insert a Post Date on a non-event post type — the sidebar button should not appear.

### Changelog Entry

> Added - Block transform from `core/post-date` to `gatherpress/event-date`, available in the toolbar's "Transform to" menu and as a "Convert to Event Date" button in the inspector sidebar.

### Credits

Props @mauteri

### Checklist:

### DIFF
--- a/src/blocks/event-date/index.js
+++ b/src/blocks/event-date/index.js
@@ -8,6 +8,7 @@ import { registerBlockType } from '@wordpress/blocks';
  */
 import edit from './edit';
 import metadata from './block.json';
+import transforms from './transforms';
 import './style.scss';
 
 /**
@@ -26,4 +27,5 @@ import './style.scss';
 registerBlockType( metadata, {
 	edit,
 	save: () => null,
+	transforms,
 } );
--- a/src/blocks/event-date/transforms.js
+++ b/src/blocks/event-date/transforms.js
@@ -1,0 +1,41 @@
+/**
+ * WordPress dependencies.
+ */
+import { createBlock } from '@wordpress/blocks';
+
+/**
+ * Block transforms for the GatherPress Event Date block.
+ *
+ * Allows a core/post-date block to be converted into a gatherpress/event-date
+ * block via the editor's "Transform to" menu, preserving the user's PHP date
+ * format and text alignment.
+ */
+const transforms = {
+	from: [
+		{
+			type: 'block',
+			blocks: [ 'core/post-date' ],
+			transform: ( attributes ) => {
+				const { format, textAlign } = attributes;
+				const newAttributes = {
+					displayType: 'start',
+				};
+
+				if ( format ) {
+					newAttributes.startDateFormat = format;
+				}
+
+				if ( textAlign ) {
+					newAttributes.textAlign = textAlign;
+				}
+
+				return createBlock(
+					'gatherpress/event-date',
+					newAttributes
+				);
+			},
+		},
+	],
+};
+
+export default transforms;
--- a/src/editor.js
+++ b/src/editor.js
@@ -12,6 +12,7 @@ import EmailNotificationManager from './components/EmailNotificationManager';
 import './stores';
 import './supports/post-id-override';
 import './supports/post-date-override';
+import './supports/post-date-convert';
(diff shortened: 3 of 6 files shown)


carstingaxion:
Haven't tested this, yet, but just from reading I'm wondering if this extra sidebar button is really needed. My initial suggestion intended to add a core block transform, available through core UI. Maybe this added extra is not needed. 

Maybe I misunderstood the intention of the button (from just reading the PR description), but otherwise GatherPress should not carry that extra maintenance-burden, if not needed. Even its just a button.

mauteri:
> Haven't tested this, yet, but just from reading I'm wondering if this extra sidebar button is really needed. My initial suggestion intended to add a core block transform, available through core UI. Maybe this added extra is not needed.
> 
> Maybe I misunderstood the intention of the button (from just reading the PR description), but otherwise GatherPress should not carry that extra maintenance-burden, if not needed. Even its just a button.

I was wondering that when I approached it, but that area is designated for variations of the block, which the date and modified date are variations of the same block. Event Date is its own block so it can't be a variation of another block. I wasn't sure if there was a way to short-circuit it to kind of a pseudo-variation, but doesn't appear to work that way. So with that, I added a Transformer. 

I can undo this if you think it isn't really a thing or make sense. Just a head's up there is a setting in GatherPress Settings for the Date block (and Date function in PHP) to use the Event Date instead of Publish Date. I got it working with blocks recently (like within that past month or so). That might be the simplest solution. Let me know what you think.

carstingaxion:
@mauteri I've missed your answer here and am sorry for being late replying.

What I meant to use is the "Block Transforms API" https://developer.wordpress.org/block-editor/reference-guides/block-api/block-transforms/

As a part of the event-date blocks block.json file, we could declare (pseudocode) like:

```json
transforms: {
    from: [
        {
            type: 'block',
            blocks: [ 'core/post-date' ],
            transform: ( { content } ) => {
                return createBlock( 'gatherpress/event-date', {
                    content,
                } );
            },
        },
    ]
},
```

and handle the attribute-transformation inside that function, like the docs say.

mauteri:
@carstingaxion Oh okay, let me check that out. Thanks!

carstingaxion:
Consider reopening this issue, review the last idea and ship a lot less code into 0.34.0 @mauteri

mauteri:
@carstingaxion yes, I have a note for this in my todo list, so it's not forgotten. Will get a follow up ticket for this to revisit. Thanks!

mauteri:
Filed [#1654](https://github.com/GatherPress/gatherpress/issues/1654) on the 0.34.0 milestone to revert this PR and re-implement the transform via the Block Transforms API per @carstingaxion's [comment above](https://github.com/GatherPress/gatherpress/pull/1565#issuecomment-4454829402).

PR #1568, opened 2026-05-01
Settings for google maps API key

### LINKED ISSUE #1494: Venue map: optional Google Maps API key for full-featured embed
Split out of [#1481](https://github.com/GatherPress/gatherpress/issues/1481) so it can land as its own focused PR.

## Background

GatherPress's Google Maps integration currently uses the keyless Google Maps Embed API — specifically the [Maps Embed URL](https://developers.google.com/maps/documentation/embed/get-started) without an API key. This works out of the box for sites that don't want to sign up for a Google Cloud account, but comes with real limitations:

- **Only `roadmap` and `satellite` map types render** in the keyless embed. The block Inspector exposes four values (`roadmap`, `satellite`, `hybrid`, `terrain`) but `hybrid` and `terrain` silently fall back to one of the two supported types when no API key is configured.
- **No Street View or Places-powered overlays**, which are only available with an authenticated Embed API or JavaScript API request.
- **No customization beyond the bare center/zoom** — no custom styles, no per-feature overlays, no marker customization.
- **Rate / request limits** that apply to keyless embeds are stricter than authenticated tiers.

These are Google's limits, not a GatherPress bug — the keyless embed was the right call for a zero-friction default. But sites that do have a Google Cloud account should be able to opt in to the full feature set by supplying their own API key.

### Settings

- New Settings → Venues → Maps field: **Google Maps API Key** (password-masked `text` input, optional, empty by default).
- Field is hidden / disabled when **Mapping Platform** is OSM, since the key has no effect there.
- Description copy explains: optional, unlocks hybrid + terrain map types, increases rate limits, enables authenticated embeds. Links to Google's [API Key setup guide](https://developers.google.com/maps/documentation/embed/get-api-key).

### Render path

- Google renderer (interactive) reads the API key from Settings and, when present, swaps to the authenticated Embed API URL (adds `&key={key}`).
- When the key is empty, keep the current keyless embed — no regression for existing installs.
- `hybrid` and `terrain` map types only appear in the Inspector's **Map type** dropdown when: platform is Google AND an API key is configured. Otherwise the dropdown stays limited to roadmap/satellite (and documents the constraint).

### Attribute validation

- The block's `type` attribute keeps all four options in its enum so saved content with `hybrid` or `terrain` survives a key-addition round trip. The Inspector just hides the options when there's no key.

### Docs / user-facing copy

- Short help text in the block Inspector explaining why hybrid/terrain aren't available until an API key is set.
- Longer explanation on the Settings page linking to Google's setup guide.

## Security considerations

- API key is a **publishable / referrer-restricted** key, not a secret — it ends up in the client-side iframe URL. Documentation should strongly recommend the user apply HTTP referrer restrictions at key-creation time.
- Still store it in the GatherPress settings option (not as a secret store) since it's inherently client-visible.
- Don't emit the key into any REST response that non-admins can read; only emit into the iframe `src` on pages where the venue-map renders.

## Acceptance criteria

- [ ] Settings → Venues → Maps has a **Google Maps API Key** field (admin-only write).
- [ ] When populated: Google embed URL includes `&key=`, hybrid + terrain map types render correctly.
- [ ] When empty: current keyless behavior is preserved, hybrid + terrain map types hidden in the block Inspector.
- [ ] Settings description explains the trade-offs and links to Google's documentation.
- [ ] Help text in the block Inspector explains the key dependency when the user selects Google platform.
- [ ] Tests cover: keyless fallback, authenticated embed URL, Inspector option filtering based on key presence.

## Related

Split from [#1481](https://github.com/GatherPress/gatherpress/issues/1481). Follow-up to [#1480](https://github.com/GatherPress/gatherpress/pull/1480) and the venue-map block work.

jmarx:
### Description of the Change
Add field for users to save their google map API key. Once a key is entered, the map used for viewing will use the embed API, but visually there will be no difference. I also removed hybrid/terrain because they are not supported in the iframe world. We need to switch to javascript API to properly use those.

Closes #1494 

### How to test the Change
Enter your key in the new google maps API and notice that rendered map code will look like this with a valid key:
```html
<iframe src="https://www.google.com/maps/embed/v1/view?key=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx&amp;center=40.7762756%2C-73.9555771&amp;zoom=18&amp;maptype=roadmap" title="1453 3rd Avenue, New York, 10028" style="border: 0px; height: 300px; width: 100%;"></iframe>
```

With an empty key, you'll still get the legacy version:
```html
<iframe src="https://maps.google.com/maps?q=40.7762756%2C-73.9555771&amp;z=18&amp;t=m&amp;output=embed" title="1453 3rd Avenue, New York, 10028" inert="" style="border: 0px; height: 300px; width: 100%;"></iframe>
```

### Changelog Entry

> Added - New feature

### Credits

Props @jmarx 

### Checklist:

- [X ] I agree to follow this project's [**Code of Conduct**](https://github.com/GatherPress/gatherpress/blob/main/CODE_OF_CONDUCT.md).
- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my change.
- [X] All new and existing tests pass.


includes/core/classes/settings/class-venues.php
```diff
@@ -104,6 +104,42 @@ protected function get_sections(): array {
...
+								__(
+									'Restrict the key by HTTP referrer.',
+									'gatherpress'
+								),
+								sprintf(
+									'<a href="%1$s" target="_blank" rel="noopener noreferrer">%2$s</a>',
+									esc_url( 'https://developers.google.com/maps/documentation/embed/get-api-key' ),
+									esc_html__( 'Get an API key', 'gatherpress' )
+								)
+							)
```
mauteri, line 134:
```
'description' => wp_kses(
	sprintf(
		/* translators: %s: link to "Get an API key" documentation. */
		// phpcs:ignore Generic.Files.LineLength.TooLong
		__( 'Optional. Referrer-restricted key for Google Maps (Embed and Static APIs). Does not unlock interactive map styles; blocks use roadmap and satellite. Enables referrer restrictions and cached static images with the static provider. Restrict the key by HTTP referrer. %s', 'gatherpress' ),
		'<a href="https://developers.google.com/maps/documentation/embed/get-api-key" target="_blank" rel="noopener noreferrer">' . esc_html__( 'Get an API key', 'gatherpress' ) . '</a>'
	),
	array(
		'a' => array(
			'href'   => array(),
			'target' => array(),
			'rel'    => array(),
		),
	)
),
```

includes/core/classes/class-settings.php
```diff
@@ -178,12 +180,18 @@ public function add_editor_settings( array $settings ): array {
...
-			'mapTileAttribution' => self::get_map_tile_attribution(),
+			'timezoneChoices'       => Utility::timezone_choices(),
+			'siteTimezone'          => Utility::get_system_timezone(),
+			'pluginUrl'             => GATHERPRESS_CORE_URL,
+			'homeUrl'               => get_home_url(),
+			'mapTileUrl'            => self::get_map_tile_url(),
+			'mapTileAttribution'    => self::get_map_tile_attribution(),
+			'venuesMapsSettingsUrl' => admin_url(
+				sprintf(
+					'edit.php?post_type=gatherpress_event&page=%s',
```
mauteri: replace `gatherpress_event` with constant for post type

includes/core/classes/settings/class-venues.php
```diff
@@ -104,6 +104,42 @@ protected function get_sections(): array {
...
+						),
+						'field'       => array(
+							'label' => __( 'Google Maps API key:', 'gatherpress' ),
+							'type'  => 'password',
```
mauteri: I think this can just be text. Google Map API I believe is exposed in JS anyway so it's not that sensitive. You can leave the password as part of the GatherPress Settings API though. Make sure any docs are updated.

includes/core/classes/class-settings.php
```diff
@@ -635,10 +644,15 @@ public function render_field( string $option, array $option_settings ): void {
...
+
 		switch ( $type ) {
+			case 'password':
 			case 'text':
-				$params['size']    = $option_settings['field']['size'] ?? 'regular';
-				$params['preview'] = $option_settings['field']['preview'] ?? array();
+				$params['input_type'] = ( 'password' === $type ) ? 'password' : 'text';
+				$template_basename    = 'text';
+				$params['size']       = $option_settings['field']['size'] ?? 'regular';
+				$params['preview']    = $option_settings['field']['preview'] ?? array();
```
mauteri: I wouldn't combine these. You can repeat some of the things and not add additional logic breaking them up.

includes/core/classes/class-settings.php
```diff
@@ -655,12 +669,24 @@ public function render_field( string $option, array $option_settings ): void {
 				break;
 		}
 
+		if ( 'google_maps_api_key' === $option ) {
+			$gatherpress_google_api_initially_visible = ( 'google' === $this->get( 'map_platform' ) );
+			printf(
+				'<div class="gatherpress-settings-google-api-key" data-gatherpress-google-api-key-field="1"%s>',
+				$gatherpress_google_api_initially_visible ? '' : ' hidden'
+			);
+		}
```
mauteri, line 683: We don't need to hide it if it's not selected.

includes/core/classes/class-settings.php
```diff
@@ -691,6 +717,56 @@ public function render_field( string $option, array $option_settings ): void {
 			echo '<p class="description gatherpress-field-inherited__note">' . $inherited_message . '</p>';
 			echo '</div>';
 		}
+
+		if ( 'google_maps_api_key' === $option ) {
+			echo '</div>';
+		}
```
mauteri: Remove this too.

includes/core/classes/class-settings.php
```diff
@@ -691,6 +717,56 @@ public function render_field( string $option, array $option_settings ): void {
...
+	 */
+	public function print_venues_map_platform_toggle_script(): void {
+		if ( ! is_admin() ) {
+			return;
+		}
+
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended -- Read-only page context check.
+		$page = isset( $_GET['page'] ) ? sanitize_text_field( wp_unslash( (string) $_GET['page'] ) ) : '';
+
+		if ( '' === $page || 0 !== strpos( $page, 'gatherpress_' ) ) {
```
mauteri: Use `str_starts_with`. WordPress has a polyfill for some PHP 8+ functions so it's safe to use.

includes/core/classes/class-settings.php
```diff
@@ -691,6 +717,56 @@ public function render_field( string $option, array $option_settings ): void {
...
+		field.hidden = ! show;
+		var row = field.closest( 'tr' );
+		if ( row ) {
+			row.style.display = show ? '' : 'none';
+		}
+	}
+	select.addEventListener( 'change', gatherpressToggleGoogleMapsApiKeyRow );
+	gatherpressToggleGoogleMapsApiKeyRow();
+} )();
+</script>
```
mauteri: I believe you don't need this anymore. For now let's just show Google Maps API regardless. I think maybe we can add some conditional logic as part of the Settings API, but we can do that in a follow up. I need to think about it.

includes/templates/admin/settings/fields/text.php
```diff
@@ -23,14 +24,19 @@
...
 // drop inherited values out of the blog option on save.
 $gatherpress_readonly = ! empty( $disabled ) ? ' readonly' : '';
 ?>
 <div class="form-wrap">
 	<label for="<?php echo esc_attr( $option ); ?>"><?php echo esc_html( $label ); ?></label>
-	<input id="<?php echo esc_attr( $option ); ?>" type="text" name="<?php echo esc_attr( $name ); ?>" class="<?php echo esc_attr( $size . '-text' ); ?>" value="<?php echo esc_attr( $value ); ?>"<?php echo $gatherpress_readonly; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- static value. ?> />
+	<input id="<?php echo esc_attr( $option ); ?>" type="<?php echo esc_attr( $gatherpress_input_type ); ?>" name="<?php echo esc_attr( $name ); ?>" class="<?php echo esc_attr( $size . '-text' ); ?>" value="<?php echo esc_attr( $value ); ?>"<?php echo $gatherpress_password_attrs; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- static attribute names/values. ?><?php echo $gatherpress_readonly; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- static value. ?> />
 	<?php
 	if ( ! empty( $description ) ) {
 		?>
```
mauteri, line 36: Create an explicit template for password, don't mix with text.

src/blocks/venue-map/edit.js
```diff
@@ -261,6 +262,31 @@ const Edit = ( { attributes, setAttributes, context, clientId } ) => {
...
+	const googleMapTypeSelectOptions = [
+		{
+			label: __( 'Roadmap', 'gatherpress' ),
+			value: 'roadmap',
+		},
+		{
+			label: __( 'Satellite', 'gatherpress' ),
+			value: 'satellite',
+		},
+	];
```
mauteri: Add the hybrid and terrain to array for now, but let's programmatically remove them below for iframe, so when we switch to JS is just about logic of leaving them in there. Leave a comment explaining.

src/blocks/venue-map/render.php
```diff
@@ -137,17 +137,25 @@
...
-		'mapPlatform'  => Settings::get_instance()->get( 'map_platform' ),
+		'mapPlatform'  => $gatherpress_map_platform,
 		'pluginUrl'    => GATHERPRESS_CORE_URL,
 	);
 
+	if ( 'google' === $gatherpress_map_platform ) {
+		$gatherpress_block_attrs['googleMapsApiKey'] = (string) $gatherpress_settings->get(
+			'google_maps_api_key'
+		);
+	}
```
mauteri: Add logic into attrs array not an if statement. It doesn't matter if it's set for OSM or not.

mauteri (changes requested):

includes/core/classes/settings/class-venues.php
```diff
@@ -104,6 +104,35 @@ protected function get_sections(): array {
...
+								. '" target="_blank" rel="noopener noreferrer">'
+								. esc_html__( 'Get an API key', 'gatherpress' ) . '</a>'
+							),
+							array(
+								'a' => array(
+									'href'   => true,
+									'target' => true,
+									'rel'    => true,
+								),
+							)
```
mauteri, line 134: This is still not correct, please see the example I provided above. You don't need to escape hardcoded URLs and such, you should just be able to copy/paste what I have here in my previous comment.

includes/core/classes/class-settings.php
```diff
@@ -635,11 +644,17 @@ public function render_field( string $option, array $option_settings ): void {
 			'disabled'    => $inherited,
 		);
 
+		$template_basename = $type;
```
mauteri: Why did you create this variable and and change lines `678-682`?

includes/templates/admin/settings/fields/password.php
```diff
@@ -0,0 +1,52 @@
...
+	<input id="<?php echo esc_attr( $option ); ?>" type="password" name="<?php echo esc_attr( $name ); ?>" class="<?php echo esc_attr( $size . '-text' ); ?>" value="<?php echo esc_attr( $value ); ?>" autocomplete="off" spellcheck="false"<?php echo $gatherpress_readonly; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- static value. ?> />
+	<?php
+	if ( ! empty( $description ) ) {
+		?>
+		<p class="description"><?php echo wp_kses_post( $description ); ?></p>
+		<?php
+	}
+
+	if ( ! empty( $preview['template'] ) ) {
+		\GatherPress\Core\Utility::render_template(
```
mauteri: Include `use GatherPress\Core\Utility` at top of file and call just `Utility` here.

includes/templates/admin/settings/fields/text.php
```diff
@@ -2,18 +2,15 @@
 /**
  * Template for displaying a text input field in GatherPress settings.
  *
- * This template is used to display a text input field with an associated label and optional description
- * in GatherPress settings pages.
- *
```
mauteri, line 7: Why'd you remove this description (this left over from when password used this template too)?

src/blocks/venue-map/render.php
```diff
@@ -137,15 +137,18 @@
...
+	$gatherpress_settings     = Settings::get_instance();
+	$gatherpress_map_platform = (string) $gatherpress_settings->get( 'map_platform' );
+	$gatherpress_block_attrs  = array(
+		'address'          => $gatherpress_address,
+		'latitude'         => (string) ( $gatherpress_venue_meta['latitude'] ?? '' ),
+		'longitude'        => (string) ( $gatherpress_venue_meta['longitude'] ?? '' ),
+		'mapZoomLevel'     => $attributes['zoom'] ?? Map::DEFAULT_ZOOM,
+		'mapType'          => $attributes['type'] ?? Map::DEFAULT_MAP_TYPE,
+		'mapHeight'        => $attributes['height'] ?? Map::DEFAULT_HEIGHT,
+		'mapPlatform'      => $gatherpress_map_platform,
```
mauteri: Just call `(string) $gatherpress_settings->get( 'map_platform' )` here

src/blocks/venue-map/render.php
```diff
@@ -137,15 +137,18 @@
...
+		'pluginUrl'        => GATHERPRESS_CORE_URL,
+		'googleMapsApiKey' => (string) $gatherpress_settings->get( 'google_maps_api_key' ),
```
mauteri: Just call `Settings::get_instance()->get( 'google_maps_api_key' )` here

src/blocks/venue-map/render.php
```diff
@@ -137,15 +137,18 @@
...
-		'address'      => $gatherpress_address,
-		'latitude'     => (string) ( $gatherpress_venue_meta['latitude'] ?? '' ),
-		'longitude'    => (string) ( $gatherpress_venue_meta['longitude'] ?? '' ),
-		'mapZoomLevel' => $attributes['zoom'] ?? Map::DEFAULT_ZOOM,
-		'mapType'      => $attributes['type'] ?? Map::DEFAULT_MAP_TYPE,
-		'mapHeight'    => $attributes['height'] ?? Map::DEFAULT_HEIGHT,
-		'mapPlatform'  => Settings::get_instance()->get( 'map_platform' ),
-		'pluginUrl'    => GATHERPRESS_CORE_URL,
+	$gatherpress_settings     = Settings::get_instance();
+	$gatherpress_map_platform = (string) $gatherpress_settings->get( 'map_platform' );
```
mauteri: Dont' need these variables for one-time use below.

src/blocks/venue-map/view.js
```diff
@@ -16,7 +16,7 @@ import MapEmbed from '../../components/MapEmbed';
  * Static-mode wrappers are left alone — their pre-rendered `<img>` is the
  * final output. For interactive wrappers we read the JSON payload that
  * render.php emitted on the outer `<div>`, replace the wrapper's static
- * children with a React root, and mount `MapEmbed` (which drives Leaflet).
+ * children with a React root, and mount `MapEmbed` (Google iframe or Leaflet).
```
mauteri: remove `iframe` here, just make it `(Google Maps or Leaflet)`

src/components/GoogleMap.js
```diff
@@ -16,41 +16,121 @@ import { select } from '@wordpress/data';
...
+
+/**
+ * Map type slugs stored on the venue-map block (editor + legacy iframe).
+ */
+const BLOCK_MAP_TYPE_SLUGS = [
+	'roadmap',
+	'satellite',
+	'hybrid',
+	'terrain',
+];
```
mauteri, line 64: You have this basically repeated in `src/blocks/venue-map/edit.js`. I would take what we did there and add to this component and call it from here, or create some helper. Just don't repeat same code in two places like you have it.

src/components/GoogleMap.js
```diff
@@ -16,41 +16,121 @@ import { select } from '@wordpress/data';
...
+ * Google Maps legacy keyless embed uses single-letter `t=` query values.
+ * Hybrid/terrain block values are coerced to satellite/roadmap for parity with the Embed API path.
  *
  * @see https://developers.google.com/maps/documentation/embed/map-parameters
  */
 const GOOGLE_MAP_TYPE_CODES = {
 	roadmap: 'm',
 	satellite: 'k',
-	hybrid: 'h',
-	terrain: 'p',
```
mauteri, line 36: I would make these values part of that helper too that manages these, so we have one constant with display name, slug, and value.

mauteri (changes requested):

jmarx:
@mauteri Ok, I think I got it all now. Take another look please?

includes/templates/admin/settings/fields/text.php
```diff
@@ -2,18 +2,19 @@
...
- * @param string $option      The option name for retrieving the field's value.
- * @param string $value       The current value of the text input field.
- * @param string $description (Optional) Additional information or instructions for the field.
- * @param string $size        The size class for styling (e.g., 'regular', 'large', or 'small').
+ * @param string       $name        The name attribute for the input field.
+ * @param string       $label       The label text for the input field.
+ * @param string       $option      The option name for retrieving the field's value.
+ * @param string       $value       The current value of the text input field.
+ * @param string       $description (Optional) Additional information or instructions for the field.
+ * @param string       $size        The size class for styling (e.g., 'regular', 'large', or 'small').
```
mauteri: You can revert this file. Your comment above is AI being confused since you changed it and change it back.

mauteri (changes requested):
Almost there, just small fix to revert file and then we can test it fully and make sure it works well. thx

mauteri (approved):

PR #1678, opened 2026-05-24
Backfill @since docblocks with actual git-history versions

### LINKED ISSUE #1665: Replace blanket @since 1.0.0 docblocks with accurate versions from git history
### Describe the enhancement

Today nearly every `@since` tag in the codebase reads `@since 1.0.0` regardless of when the symbol was actually introduced — a legacy of an early assumption that 1.0.0 was close. The plugin is still pre-1.0 (currently 0.34.0), so the tag conveys no real version info. That noise propagates straight into the auto-generated extender docs at `docs/developer/hooks/`, where every documented filter and action currently claims to have shipped in `1.0.0` — extenders pinning to a specific GatherPress version have no real way to ask "was filter X around in 0.30.0?"

This issue proposes a one-time backfill: replace `@since 1.0.0` with the actual version each symbol shipped in, sourced from git tags.

### Scope (tier order)

1. **Tier 1 — public hooks.** Every `apply_filters()` / `do_action()` call site. Highest value because these flow into the published hook reference under `docs/developer/hooks/` on the next CI regen.
2. **Tier 2 — public class methods on the public API surface** (`Event`, `Rsvp`, `Venue`, `Settings`, REST API classes, and other classes documented for extenders).
3. **Tier 3 — internal helpers, protected/private methods, private utility functions.** Lower priority; reasonable to leave at `1.0.0` for now or roll into the same pass if cheap.

If Tier 1 is the only thing that lands under this issue and the rest gets its own follow-up, that's fine — the user-visible win is in Tier 1.

### How to derive the right version

For each `@since 1.0.0` occurrence:

1. Locate the symbol the docblock is documenting (filter name string, function name, `Class::method`).
2. Find the introducing commit, e.g. `git log --diff-filter=A --reverse -G '<symbol-introduction-pattern>' -- <file>`. For PHP methods, the symbol pattern is the method signature; for filters/actions, the string literal.
3. Run `git describe --contains <commit>` to get the first tag containing that commit, then strip the suffix to extract just the version (`0.30.0-N-gabcdef` → `0.30.0`).
4. Patch the `@since` line.

### Constraints / known caveats

- **Earliest tag is the floor.** Anything introduced before the earliest tag should fall back to the earliest version (~`0.27.0` if that's the floor — verify with `git tag --sort=v:refname | head -1`). State that explicitly in the script's fallback path.
- **Renames and namespace moves break `git log --follow`.** We did several flatten-into-subnamespace passes (e.g. the `Event_*` → `GatherPress\Core\Event\*` move). For affected symbols, fall back to `git log --all --diff-filter=A -G` across all paths, take the earliest match. Worth a quick human eyeball on the diff for the moved files.
- **Signature changes vs introduction.** `@since` reflects when the symbol *first appeared*, not when its signature was last touched. If a hook was added in 0.28.0 and gained an extra `@param` in 0.31.0, the docblock stays `@since 0.28.0`; signature evolution can be documented in a separate paragraph or `@since-newer-version` note (matches WP core convention).
- **Don't run this before 0.34.0 ships.** Anything genuinely added in 0.34.0 (e.g. the new `gatherpress_asset_utility_style_block_prefixes` filter from PR #1655) should be tagged `@since 0.34.0` directly. Running the bulk script before the 0.34.0 tag exists would convert those to whatever 0.33.x they were committed under and then need a second correction.

### Tactical approach

- Write a small one-off script that walks every `@since 1.0.0` and outputs the patched line. Eyeball the diff for renamed/moved symbols and correct manually where the script's heuristic is wrong.
- Coordinate with the auto-regen workflow at `docs/developer/hooks/` — let CI regenerate that directory on the merge to develop rather than committing regen in the same PR (per `AGENTS.md` convention).
- Split into Tier 1 / Tier 2 / Tier 3 PRs if the Tier 1 diff alone is reviewable; otherwise one bigger PR is fine.

### Code of Conduct

### Canonical docblock format

While touching `@since` lines, normalize each docblock to the project's canonical shape — short description first, then `@since` separated by a blank line, then the `@param` group separated by another blank line, then `@return` separated by another blank line:

```php
	/**
	 * Method description.
	 *
	 * @since 0.30.0
	 *
	 * @param int[] $param_1 Param description.
	 * @param int[] $param_2 Param description.
	 *
	 * @return int[] Description.
	 */
```

Rules:

- A blank line (`\t * `) between the short description and `@since`.
- A blank line between `@since` and the `@param` group.
- A blank line between the `@param` group and `@return`.
- No blank lines *inside* the `@param` group — all `@param` lines run consecutively.
- Same convention for JS JSDoc blocks (`@since`, `@param`, `@return`).

This is purely a formatting normalize — semantic content stays whatever each docblock already documents.

### No `@since 1.0.0` should remain after this work

Once this cleanup lands, **zero** `@since 1.0.0` strings should be left in the codebase. 1.0.0 has not been released — every `@since` must resolve to a tag that actually exists in `git tag` (the earliest is `0.27.0`; the current development line is `0.34.0`). If the script can't confidently date a symbol (e.g. heavy file renames break `git log --follow`), fall back to the earliest tag that contains the file in its current path rather than leaving `1.0.0`.

Acceptance criterion: `grep -rn "@since 1.0.0" includes/ src/ test/` returns no results after the cleanup PR merges (other than this issue's reference, if any).

mauteri:
## Summary

- Replaces 1,759 occurrences of placeholder `@since 1.0.0` across 297 PHP, JS, and SCSS files under `includes/`, `src/`, and `test/` with the version each symbol actually shipped in.
- PHP symbols (filters/actions, classes/interfaces/traits, methods/functions, constants, properties) were resolved by their introducing commit via `git log --all --reverse -G <pattern>`, then mapped to the first containing git tag via `git describe --contains`.
- Pre-release suffixes (`-alpha.N`, `-beta.N`, `-rc.N`) were normalized to the stable base version.
- Commits predating the earliest existing tag (`0.27.0`) floor to that tag; commits not yet in any tag (current `0.34.0` development line) map to `0.34.0`.
- JS/TS/SCSS files were dated by their first commit (per-symbol resolution there would need a separate parser; file-level is reasonable since most JS surfaces in this codebase correspond 1:1 to a file).
- The one-off backfill script lives in this PR's history but is not committed (per issue scope as "one-off script").

Acceptance criterion: `grep -rn "@since 1.0.0" includes/ src/ test/` now returns zero results.

Closes #1665

## Final version distribution

| Version  | Files |
|---|---|
| `0.27.0` (floor) | 101 |
| `0.28.0` | 4 |
| `0.29.0` | 7 |
| `0.29.1` | 1 |
| `0.30.0` | 13 |
| `0.31.0` | 7 |
| `0.33.0` | 66 |
| `0.33.1` | 1 |
| `0.33.3` | 1 |
| `0.34.0` | 145 |

## Tier 1 hook spot-checks

Manually verified a sample of public hook resolutions:

| Hook | Resolved | Real introduction |
|---|---|---|
| `gatherpress_event_feed_excerpt` | `0.33.0` | Commit `0608028e` (2025-08-03), contained by `0.33.0-alpha.1` ✅ |
| `gatherpress_calendar_url` | `0.34.0` | Commit `82629023` (2026-05-19), post-`0.33.3` ✅ |
| `gatherpress_async_geocode_failed` | `0.34.0` | Commit `af00ff6b` (2026-04-26), post-`0.33.3` ✅ |
| `gatherpress_template_path` | `0.34.0` | Commit `178cb239` (#1675, 2026-05-23) ✅ |
| `gatherpress_static_map_tile_url` | `0.34.0` | Static map subsystem ✅ |
| `gatherpress_pseudo_post_metas` | `0.34.0` | Recent settings work ✅ |

## Caveats

- For symbols renamed across versions (e.g. `Query` → `Event_Query` flatten before `0.27.0`), the script searches scoped to the current file path first; falls back to unscoped only if scoped finds nothing. This dates each symbol by when its *current* declaration first appeared in its *current* file — semantically "when did this current shape ship" rather than "when did this concept first exist."
- Same-named symbols in different files (e.g. `setup_hooks` exists on most subsystems) are dated independently per-file via file-scoped `git log -- <file>` lookups.
- Filter/action hook patterns require quotes around the hook name in the search regex so PHP variables that happen to share a name (e.g. `$gatherpress_template_path` the variable vs `'gatherpress_template_path'` the filter) don't cause cross-contamination.

## Test plan

- [ ] `grep -rn "@since 1.0.0" includes/ src/ test/` → empty.
- [ ] PHPCS still passes on a sample of touched files (no docblock formatting regressions).
- [ ] Existing PHPUnit suite still passes (no semantic changes; this is purely docblock text).
- [ ] On next CI run, `docs/developer/hooks/` regen should show the new versions in each per-hook reference page.


jmarx:
@mauteri is this what you call a tight pr? :) JK. Great work!

carstingaxion:
I don't know what motivated you, to do this now, but I appreciate that a lot.

I asked for the exact same one-time-automatism as I tested a Playground powered code reference for GatherPress using the developer.wordpress.org theme and phpdoc parser. The actual blocker to that idea was that all methods had since 1.0.0 tags which destroyed some of the logic.

Now we could easily have an automated code reference that looks and works 100% like w.org

mauteri:
> I don't know what motivated you, to do this now, but I appreciate that a lot.
> 
> I asked for the exact same one-time-automatism as I tested a Playground powered code reference for GatherPress using the developer.wordpress.org theme and phpdoc parser. The actual blocker to that idea was that all methods had since 1.0.0 tags which destroyed some of the logic.
> 
> Now we could easily have an automated code reference that looks and works 100% like w.org

I did this recently at work with another plugin with Claude and the git history, so I tried it here. It annoyed me that early on I made everything 1.0.0 because I thought it was close, but then it wasn't haha. This may not be perfect, but it's probably close enough and will be good going forward.

Issue #1670, opened 2026-05-23
Authenticate composer in 'wp package install' CI steps to avoid GitHub anonymous rate-limit flakes


mauteri:
### Describe the bug

Two CI workflows intermittently fail at the `wp package install wp-cli/dist-archive-command:^2` step with one of these messages:

```
Warning: Could not authenticate against github.com
Error: Package installation failed.
```

```
Error: Failed to get composer instance: Your github oauth token for github.com contains invalid characters: "***"
```

Pattern is "fails on push, passes on rerun" — characteristic of GitHub's unauthenticated API rate-limit (**60 requests/hour per source IP**) being exhausted by a single push that fans out to many parallel workflows. GitHub Actions runners share IP pools across thousands of repos, so the burst can blow through that quota.

### Affected workflows

- [`.github/workflows/wordpress-org-plugin-guidelines.yml:34`](https://github.com/GatherPress/gatherpress/blob/develop/.github/workflows/wordpress-org-plugin-guidelines.yml) — `Install dist-archive-command` step.
- [`.github/workflows/playground-preview.yml:52`](https://github.com/GatherPress/gatherpress/blob/develop/.github/workflows/playground-preview.yml) — same step.

Neither passes credentials to composer, so its underlying github.com fetch goes out anonymously.

### Fix

Pass the workflow's `GITHUB_TOKEN` to composer via the `COMPOSER_AUTH` env var on each affected step. That raises composer's effective quota from 60/hr per IP to 5000/hr per token, which is uncapped for CI volumes.

```yaml
- name: Install dist-archive-command
  env:
    COMPOSER_AUTH: '{"github-oauth": {"github.com": "${{ secrets.GITHUB_TOKEN }}"}}'
  run: wp package install wp-cli/dist-archive-command:^2
```

### Notes / caveats

- The token's read scope (which is what `permissions: contents: read` gives on these workflows) is enough for composer to clone public repos and fetch package metadata.
- This does **not** require relaxing the existing `permissions: contents: read` blocks — `secrets.GITHUB_TOKEN` is still scoped by them.
- If a similar pattern shows up elsewhere (e.g. a future `wp package install` for a different command, or a direct `composer global require ...`), apply the same env block.

### Steps to Reproduce

1. Trigger a push to a PR branch that touches paths watched by both workflows.
2. Watch `wp-org-plugin-guidelines` (and sometimes `playground-preview`) fail at the `wp package install` step.
[...]

PR #1710, opened 2026-05-30
fix(ci): drop COMPOSER_AUTH from dist-archive-command install (unblocks Playground Preview + wp.org Guidelines)

mauteri:
## Problem
Both `playground-preview.yml` and `wordpress-org-plugin-guidelines.yml` fail on:
```
Error: Failed to get composer instance: Your github oauth token for github.com contains invalid characters: "***"
```
The `COMPOSER_AUTH: GITHUB_TOKEN` (added in #1670 to avoid API rate limits) trips Composer's github-oauth token validation, so `wp package install wp-cli/dist-archive-command` aborts.

## Fix
`dist-archive-command` is a public package and needs no auth — remove the `COMPOSER_AUTH` env from the install step in both workflows. This restores the pre-#1670 working install.

If anonymous API rate limits become a real issue again, the right re-fix is git-credential auth (`git config url.insteadOf` with the token), which doesn't go through Composer's oauth validation.

### DIFF
--- a/.github/workflows/playground-preview.yml
+++ b/.github/workflows/playground-preview.yml
@@ -50,8 +50,6 @@ jobs:
           tools: wp-cli
 
       - name: Install dist-archive-command
-        env:
-          COMPOSER_AUTH: '{"github-oauth": {"github.com": "${{ secrets.GITHUB_TOKEN }}"}}'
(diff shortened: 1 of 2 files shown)


mauteri:
Folded into #1709.

PR #1716, opened 2026-05-31
Fix: render Duration select when the 2h default is filtered out

### LINKED ISSUE #1706: Duration select is not rendered, when 2h-default-duration is not selectable
### Describe the bug

**Date Time End** picker is immmediately rendered instead of the comfortable **Duration** select, when the `'gatherpress.durationOptions'` filter does not provide the 2h default duration (which is hardcoded into the logic).

*Took me some time to figure that tricky reason out.*

1. I would like to see the Duration select rendered with whatever (valid) values are given, independent from the 2h-default duration
2. I would recommend using the first or the last, but any value of the `'gatherpress.durationOptions'` array as the default duration

### Steps to Reproduce

Change the default duration 

```php
<?php
/**
 * Plugin Name: GatherPress Duration Options
 * Description: Adds custom duration options to the GatherPress plugin using the durationOptions filter.
 * Version: 1.0
 * Author: Mike Auteri & Carsten Bach
 * License: GPL2+
 */

 defined( 'ABSPATH' ) || exit;

 add_action( 'admin_print_footer_scripts', 'gatherpress_duration_options_inline_script', 100 );

 function gatherpress_duration_options_inline_script() {
	 // Only load on block editor screens
	 $screen = get_current_screen();
	 if ( ! method_exists( $screen, 'is_block_editor' ) || ! $screen->is_block_editor() ) {
		 return;
	 }
	 ?>
	 <script>
		 (function () {
			 if ( typeof wp === 'undefined' || ! wp.hooks ) return;

			 wp.hooks.addFilter(
				 'gatherpress.durationOptions',
				 'gatherpress/durationOptionsTest',
				 function ( options ) {
					 return [
						{
							label: '3 OR 2 hours',
							value: 3, // use this value to toggle your test
						},
						{
							label: '15 hours',
							value: 15,
						},
						{
							label: '23 hours',
							value: 23,
						},
						{
							label: '48 hours',
							value: 48,
						},
						{
							 label: 'Set an end time...',
							 value: false,
						},
					 ];
				 }
			 );
		 })();
	 </script>
	 <?php
 }
```

https://github.com/user-attachments/assets/c0a20371-7896-451c-b9ef-075588a8ef61

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

mauteri:
Closes #1706

## Problem

A new event's end is defaulted to `start + 2h`, and `DateTimeRange` only renders the **Duration** select when the current end matches one of the `durationOptions` presets (`matchedDuration ? <Duration /> : <DateTimeEnd />`). When a `gatherpress.durationOptions` filter omits the built-in 2-hour option, the default end matches no preset and the UI drops straight to the **Date Time End** picker instead of the Duration select.

## Why the obvious fix is not enough

`defaultDateTimeEnd` is a **module-load constant** (computed at import, seeded into the store's `DEFAULT_STATE`). In the reported repro the filter is registered late (a footer `<script>`), so it is not present when that constant is computed, but it *is* present at React render time. Feeding a default-duration helper into the frozen constant alone would not fix the reported case.

## Fix

- Add `getDefaultDuration()`: prefer `2` when it is an available option, else the first **numeric** option (skipping the `false` "Set an end time…" sentinel), else `2` as a last resort. This guarantees the default maps to a selectable preset.
- Seed a new event's `duration` and derived `dateTimeEnd` from `getDefaultDuration()` in the store's editor-ready `subscribe`, where the filter is already registered and `durationOptions()` is live. This is what makes it robust to late-registered filters.
- Route the remaining hardcoded 2h fallbacks (`getDefaultDateTimeEnd`, the `validateDateTimeStart`/`validateDateTimeEnd` adjustments) through the same helper.

No change to the default option set behavior: with 2h present, `getDefaultDuration()` returns `2` and everything is identical to before.

## Tests

- New `getDefaultDuration` coverage: 2h present, 2h filtered out (first numeric), `false` sentinel skipped, no-numeric last resort.
- Updated the store subscribe test to assert a new event seeds the default duration + derived end, and that a saved end is used verbatim (no seeding).
- `helpers/datetime.js` reports 100% statements/branches/functions/lines locally.

### DIFF
--- a/.github/changelog/add-default-duration-filter
+++ b/.github/changelog/add-default-duration-filter
@@ -1,4 +1,0 @@
-Significance: minor
-Type: added
-
-Add a `gatherpress.defaultDuration` JavaScript filter to override the default event duration (in hours). The value is honored when it matches one of the available `durationOptions`, otherwise it falls back to the first available duration. [#1706]
--- a/.github/changelog/add-default-duration-filter
+++ b/.github/changelog/add-default-duration-filter
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Add a `gatherpress.defaultDuration` JavaScript filter to override the default event duration (in hours). The value is honored when it matches one of the available `durationOptions`, otherwise it falls back to the first available duration. [#1706]
--- a/.github/changelog/add-duration-default-filter
+++ b/.github/changelog/add-duration-default-filter
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Add a `gatherpress.durationDefault` JavaScript filter to override the default event duration (in hours). The value is honored when it matches one of the available `durationOptions`, otherwise it falls back to the first available duration. [#1706]
--- a/.github/changelog/fix-duration-select-default-option
+++ b/.github/changelog/fix-duration-select-default-option
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Render the Duration select with whatever duration options are provided, defaulting to the first available option when the built-in 2-hour default is filtered out via `gatherpress.durationOptions`, instead of dropping straight to the end-time picker. [#1706]
--- a/docs/developer/event-duration.md
+++ b/docs/developer/event-duration.md
@@ -4,7 +4,7 @@ GatherPress lets developers customize the event **Duration** control in the
 block editor through two JavaScript filters:
 
 1. [`gatherpress.durationOptions`](#gatherpressdurationoptions) — change the list of selectable durations.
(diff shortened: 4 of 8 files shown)


carstingaxion:
Thanks for the fix!

Minor detail: I would have placed the filter doc inside /hooks, even it is not auto-generated. Just to follow the convention "where we document hooks and filters"

Sometime ago I asked the maintainer of extract-wp-hooks to start supporting JS filters as well,but had no response yet.

- https://github.com/akirk/extract-wp-hooks/issues/30

PR #1689, opened 2026-05-24
Generalize venue-detail + event-query toggle to any shadow-source post type

mauteri:
### Description of the Change
The 0.34.0 post-type-supports work made the shadow-source pattern broadly usable — a band can register a `tour` CPT with `gatherpress-event-date` + `gatherpress-shadow-source` supports, a theatre can register `production` and `season`, a conference can register `track`. All connect to events via the same shadow-taxonomy mechanism that already powered venues. The editor surface and event-query toggle were still venue-hardcoded though — extending the architectural promise to those two block surfaces is what this PR does.

**Venue-detail block (#1687):**
- New `sourcePostType` block attribute (default `gatherpress_venue`). When set explicitly (e.g. `sourcePostType: "production"`), the block resolves its connected source post via that CPT's shadow taxonomy.
- New `Shadow_Source::get_source_post_from_event_post_id( int $event_id, string $source_post_type ): ?WP_Post` — generic version of the venue-specific lookup. Skips sentinel terms, returns null gracefully.
- `Blocks\Venue::render_block()` now reads `sourcePostType` instead of hardcoded venue lookups; method names updated (`get_source_post`, `render_with_source_context`).
- `src/blocks/venue/edit.js` passes `sourcePostType` through the venue helper chain so the editor preview matches runtime — the existing helpers were already taxonomy-name-parameterized, just needed to feed in the right value.

**Event-query contextual toggle (#1688):**
- `Event\Query::pre_get_posts` now scopes by any shadow-source-supporting CPT's term when on a singular source page (not just venues). Uses the shared `Shadow_Source::get_taxonomy()` + `term_slug_from_post_name()` resolution path.
- Toggle label adapts via `usePostTypeLabel` — on a Tour edit screen it shows "Filter by Current Tour", on a Production "Filter by Current Production", venue fallback elsewhere.

**Editor-preview parity for the contextual toggle:**
- Frontend resolves the source post via `is_singular()` + `get_queried_object()`. The REST preview path the editor uses can't do that — there's no singular page yet. Without this fix the editor showed every event instead of only the ones scoped to the page being edited.
- Block now writes the editor's current post id + post type into the query attrs (`gatherpress_shadow_source_post_id`, `gatherpress_shadow_source_post_type`) when the toggle is on. `query_loop_block_query_vars` + `rest_query` thread them through and the Shadow_Source resolver falls back to them when `is_singular()` is false.
- Result: editor preview matches the rendered frontend for every shadow-source CPT.

[...]

PR #1725, opened 2026-06-02
Show "Exclude current event" toggle only when queried post type matches the currently edited post type

### LINKED ISSUE #1705: Show "Exclude Current Event" only, when context post_type and queried post_type match

I found the toggle avail. on a Production post_type, while querying for events.

This toggle is currently shown if the context post type supports `gatherpress-event-date` (as far as I can remember).

This is ok and fine but needs an additional guard to check if the current contextual postType is identical to query.postType.

Image

### Designs

_No response_

_No response_

### Code of Conduct

carstingaxion:
Fixes #1705

## Proposed changes:

The Event Query block's sidebar exposed two contextual toggles in places they could never apply. This PR gates each one on the relationship between the **queried** post type (`query.postType`) and the **contextual** post type the block is rendered in:

- **"Exclude Current Event"** now shows only when the contextual post type is an event AND `currentPostType === queryPostType`. Previously it appeared on any event-supporting context (e.g. while querying events from a Production), where "exclude the current event" is meaningless.
- **"Filter by Current &lt;Source&gt;"** (the shadow-source filter) now shows only when the host is a shadow-source post type AND `currentPostType !== queryPostType` (plus the existing template-context case). Querying Productions while sitting on a Production should not offer "Filter by the current Production".

Supporting changes:

- **Generalized the query var** `venue_filter` → `shadow_filter` across PHP, JS, and tests, since the filter now scopes by any shadow-source CPT (venue, production, season, ...), not just venues.
- **Made the contextual post type/ID readable by the core Query block.** `Blocks\Setup::enable_context_for_core_query_block()` adds `postType` and `postId` to `core/query`'s `uses_context` via `register_block_type_args` (registered there rather than on `init` so it runs before the block type is registered). The custom controls now read `context.postType` / `context.postId` instead of only the currently-edited post.
- **Reworked the help/label copy** to reflect the actual source and queried post types, e.g. "When placed inside Venue context, only shows Events tied to that Venue."

### Other information:

### Known limitation (still being investigated)

The mechanism works when the Event Query block lives in **post content** but not yet when placed inside a **block template** (e.g. the "Single Event" template). In the template/site editor there is no concrete post providing `postType` / `postId` block context, so the contextual resolution has nothing to bind to. This appears related to #1689 and has is being handled in #1753.

## Testing instructions:

1. Install one of the shadow-source companion plugins and create a few entries:
   - https://github.com/carstingaxion/gatherpress-productions
   - https://github.com/carstingaxion/gatherpress-seasons
2. Edit an **Event** post and add an Event Query block (or a `gatherpress/venue` block with a `sourcePostType` set, containing an Event Query block).
3. Confirm in the block sidebar:
   - "Exclude Current Event" appears only when the query post type matches the current post type.
   - "Filter by Current &lt;Source&gt;" appears only when the query post type differs from the host shadow-source post type.
   - The help text names the correct source and queried labels.
4. Confirm the editor preview and the frontend show the correctly filtered events.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Show "Exclude current event" toggle only when queried post type matches the currently edited post type

</details>

### DIFF
--- a/includes/core/classes/blocks/class-setup.php
+++ b/includes/core/classes/blocks/class-setup.php
@@ -53,7 +53,7 @@ class Setup {
 	 * @return void
 	 */
 	protected function setup_hooks(): void {
-		// Called from here because the Init:10 from inside Blocks/Event_Query would be too late for that filter to work.
+		// Called from here because Init:10 from inside Blocks/Event_Query would be too late for that filter to work.
 		add_filter( 'register_block_type_args', array( $this, 'enable_context_for_core_query_block' ), 10, 2 );
 		add_action( 'init', array( $this, 'register_block_classes' ) );
 		add_action( 'init', array( $this, 'register_block_patterns' ) );
@@ -323,10 +323,9 @@ class Setup {
 	public function enable_context_for_core_query_block( array $args, string $block_type ): array {
 		// Only modify the Query block.
 		if ( 'core/query' === $block_type ) {
-			error_log( 'enable_context_for_core_query_block called for block type: ' . $block_type );
 			$args['uses_context'] = array_merge(
-				$args['uses_context'] ?? [],
-				[ 'postType', 'postId' ]
+				$args['uses_context'] ?? array(),
+				array( 'postType', 'postId' )
 			);
 		}
 		return $args;
--- a/includes/core/classes/blocks/class-setup.php
+++ b/includes/core/classes/blocks/class-setup.php
@@ -313,7 +313,7 @@ class Setup {
 	 * This allows the block to receive context values like postType and postId,
 	 * which are necessary for some of the custom Event Query block controls.
 	 *
-	 * @since 0.33.0
+	 * @since 0.34.0
 	 *
 	 * @param array  $args       The arguments for registering the block type.
 	 * @param string $block_type The name of the block type being registered.
--- a/includes/core/classes/blocks/class-setup.php
+++ b/includes/core/classes/blocks/class-setup.php
@@ -53,6 +53,8 @@ class Setup {
 	 * @return void
 	 */
 	protected function setup_hooks(): void {
+		// Called from here because the Init:10 from inside Blocks/Event_Query would be too late for that filter to work.
+		add_filter( 'register_block_type_args', array( $this, 'enable_context_for_core_query_block' ), 10, 2 );
 		add_action( 'init', array( $this, 'register_block_classes' ) );
 		add_action( 'init', array( $this, 'register_block_patterns' ) );
 		// Priority 11 needed for block.json translations of title and description.
@@ -303,5 +305,30 @@ class Setup {
 		$post_id = isset( $block['attrs']['postId'] ) ? intval( $block['attrs']['postId'] ) : 0;
 
 		return $post_id > 0 ? $post_id : get_the_ID();
+	}
+
+	/**
+	 * Enables block context for the core Query block.
+	 *
+	 * This allows the block to receive context values like postType and postId,
+	 * which are necessary for some of the custom Event Query block controls.
+	 *
+	 * @since 0.33.0
+	 *
+	 * @param array  $args       The arguments for registering the block type.
+	 * @param string $block_type The name of the block type being registered.
+	 *
+	 * @return array Modified arguments for registering the block type.
+	 */
+	public function enable_context_for_core_query_block( array $args, string $block_type ): array {
+		// Only modify the Query block.
+		if ( 'core/query' === $block_type ) {
+			error_log( 'enable_context_for_core_query_block called for block type: ' . $block_type );
+			$args['uses_context'] = array_merge(
+				$args['uses_context'] ?? [],
+				[ 'postType', 'postId' ]
+			);
+		}
+		return $args;
 	}
 }
--- a/.github/changelog/1725-fix-exclude-current-event-toggle
+++ b/.github/changelog/1725-fix-exclude-current-event-toggle
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Show "Exclude current event" toggle only when queried post type matches the currently edited post type.
--- a/gatherpress.php
+++ b/gatherpress.php
@@ -58,15 +58,3 @@ if ( ! trait_exists( 'GatherPress\Core\Traits\Singleton' ) ) {
 GatherPress\Core\Setup::get_instance();
 
 
-function enable_context_for_core_query_block($args, $block_type) {
-
-	// Only modify the Query block
-	if ( 'core/query' === $block_type ) {
-		$args['uses_context'] = array_merge(
-			$args['uses_context'] ?? [],
-			[ 'postType', 'postId' ]
-		);
-	}
-	return $args;
-}
-add_filter('register_block_type_args', 'enable_context_for_core_query_block', 10, 2);
--- a/gatherpress.php
+++ b/gatherpress.php
@@ -56,5 +56,3 @@ if ( ! trait_exists( 'GatherPress\Core\Traits\Singleton' ) ) {
 
 // Initialize setups.
 GatherPress\Core\Setup::get_instance();
-
-
--- a/gatherpress.php
+++ b/gatherpress.php
@@ -56,3 +56,17 @@ if ( ! trait_exists( 'GatherPress\Core\Traits\Singleton' ) ) {
 
 // Initialize setups.
 GatherPress\Core\Setup::get_instance();
+
+
+function enable_context_for_core_query_block($args, $block_type) {
+
+	// Only modify the Query block
+	if ( 'core/query' === $block_type ) {
+		$args['uses_context'] = array_merge(
+			$args['uses_context'] ?? [],
+			[ 'postType', 'postId' ]
+		);
+	}
+	return $args;
+}
+add_filter('register_block_type_args', 'enable_context_for_core_query_block', 10, 2);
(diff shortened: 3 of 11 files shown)


mauteri:
@carstingaxion need help with the failing test?

carstingaxion:
Now I realised, that the "Filter by current ..." toggle should also be guarded, but the other way around. It should not be shown, when queried post type and current context post type match.

This image shows the problem: Querying for Productions, their should be no option to "Filter by the current Production"

image

I'm going to fix this on the way.

carstingaxion:
Ooops, and the label needs to be fixed, too:

> When placed on a shadow-source page, only shows events tied to that page.

carstingaxion:
I wanted to test the Event Query block and both toggles discussed in this issue inside an `gatherpress/venue` block (with different `sourcePostType` set).

Which did not work. The toggles gone. Nothing.
I remembered, that I saw some `getCurrentPostType()` calls recently that may be related. And indeed. 

I needed to open a can of worms to get the correct context `postType` and `postId` into the query block, but until now, ~~its only the UI, that works like it should. But neither the editor preview, nor the frontend rendering show the correct posts per the query settings~~ _it works_.

image

To test this, you will have to install at least one of the following plugins & create some productions/seasons
- https://github.com/carstingaxion/gatherpress-productions/archive/refs/heads/main.zip
- https://github.com/carstingaxion/gatherpress-seasons/archive/refs/heads/main.zip

<details><summary>Outdated screencast</summary>

[Bildschirmaufzeichnung vom 03.06.2026, 10:31:19.webm](https://github.com/user-attachments/assets/69056cef-ff7b-4785-aa7c-096f5875f369)

</details>

carstingaxion:
**Second big OOOPS:** The `gatherpress/event-date` block seems to consume only the currently edited, but not the contextual `postType` and `postId`. (The video in my previous comment shows the problem.)

**EDIT:** I looked into the code, which uses `context` properly. So I'm even more confused about this buggy behaviour.

**EDIT 2:** This problem seems to be independent from the `core/query` and `gatherpress/venue` blocks, because it seems to appear because I'm editing an event post. The dates shown in all queried events, match the one that holds the `core/query` blocks, no matter if nested inside a `gatherpress/venue` or totally pure inside the event itself.

**EDIT 3:** 
- https://github.com/GatherPress/gatherpress/issues/1733
- https://github.com/GatherPress/gatherpress/pull/1750

image

carstingaxion:
OMG!

Besides the fact, that the query is not filtered as it should, the result contains past AND upcoming events.

On the left side, the wrong event-date, seems the most obvious problem, but on the right side, the event-date is correctly showing. But given, that today is June 03, the query shows that events from past and upcoming are both included in the result.

image

carstingaxion:
_Getting closer._

- https://github.com/GatherPress/gatherpress/pull/1725/commits/be718213731f8070db3a69bc1e2d9869a9e9afb8 After I renamed the important query-var throughout php and not only JS, I got the filtering back to work.

image

carstingaxion:
> Ooops, and the label needs to be fixed, too:

Done :heavy_check_mark: 

- _When placed inside Venue context, only shows Events tied to that Venue._
- _When placed inside Season context, only shows Productions tied to that Season._

includes/core/classes/blocks/class-setup.php
```diff
@@ -304,4 +306,28 @@ public function get_post_id( array $block ): int {
...
 		return $post_id > 0 ? $post_id : get_the_ID();
 	}
+
+	/**
+	 * Enables block context for the core Query block.
+	 *
+	 * This allows the block to receive context values like postType and postId,
+	 * which are necessary for some of the custom Event Query block controls.
+	 *
+	 * @since 0.33.0
```
mauteri: Should be 0.34.0

carstingaxion:
Sadly, the whole mechanics do not work, when placed inside a template.

**HERE:** Example placed into the "Single Event" template. At first the WORKING queryings inside post-content, below the separator line are the same queries NOT WORKING.

_Propably related:_ https://github.com/GatherPress/gatherpress/pull/1689#issuecomment-4573830954

image

image


<hr>

- https://github.com/GatherPress/gatherpress/issues/1753

carstingaxion:
After I updated the production-under-test to carry a realistic date, the whole context post collapsed inside the event.

Smells like a new problem ...

image
image

<hr>

- #1754

carstingaxion:
"Exclude current Event"

image

"Filter by current season"

#### working within post_content

image

#### NOT working within template

image


<hr>

- https://github.com/GatherPress/gatherpress/issues/1753

mauteri:
@carstingaxion read through the whole debugging trail here. A few concrete things, separating the "definitely fix this" from the "why templates fight you."

### 1. There's a Rules-of-Hooks violation that likely explains the "toggles gone. Nothing." symptom

In `ShadowSourceFilterControls` ([`components.js:313-320`](https://github.com/GatherPress/gatherpress/blob/fix/show-exclude-current-event-toggle-only-when-queried/src/variations/core/query/components.js#L313)):

```js
const editorPostId = context?.postId || useSelect( ... );
const editorPostType = context?.postType || useSelect( ... );
```

`useSelect` only runs when the left side is falsy. Hooks must be called unconditionally and in the same order on every render. Block context resolves asynchronously, so `context.postId` flips from `undefined` to a number between renders, which changes whether the hook runs and corrupts the hook order. That's exactly the kind of "controls vanish / flicker" behavior you hit. It should also trip `react-hooks/rules-of-hooks` in lint.

Call both hooks unconditionally, then choose:

```js
const fallbackPostId = useSelect( ( wpSelect ) => wpSelect( 'core/editor' )?.getCurrentPostId(), [] );
const fallbackPostType = useSelect( ( wpSelect ) => wpSelect( 'core/editor' )?.getCurrentPostType(), [] );
const editorPostId = context?.postId || fallbackPostId;
const editorPostType = context?.postType || fallbackPostType;
```

### 2. Why it works in post content but not in a template

This is the real wall you hit in the last few comments, and it's a WordPress block-context limitation, not a bug in your logic.

- In the **post editor**, core seeds default block context with the edited post's `postId` / `postType`, so once `enable_context_for_core_query_block()` adds them to `core/query`'s `uses_context`, the query block actually receives them.
- In the **site/template editor**, there is no concrete post, so core never provides `postId` / `postType` context. Your fallback `getCurrentPostType()` then returns `wp_template`, not the event/production, and resolution has nothing to bind to.

You can trace the runtime side in [`Shadow_Source::resolve_post_from_query_context()`](https://github.com/GatherPress/gatherpress/blob/fix/show-exclude-current-event-toggle-only-when-queried/includes/core/classes/class-shadow-source.php): branch 1 needs `is_singular()` with the queried object being a shadow-source post (frontend only, and on a Single Event template the queried object is the *event*, which is not a shadow source, so this misses); branch 2 needs `gatherpress_shadow_source_post_id` / `_post_type` query vars, and those originate from block context that's absent in templates. So both branches come up empty inside a template. This is the same gap as the #1689 thread you linked.

Net: I don't think a small tweak makes editor-preview filtering work inside templates. The realistic split is (a) make the frontend resolve the shadow source even when the queried object is an event (resolve it from the event's relationship rather than from `get_queried_object()`), and (b) treat the in-template editor preview as best-effort. Worth deciding whether template support is in-scope for this PR or a follow-up, because it's a meaningfully bigger change than the toggle-gating this PR is named for.

### 3. Cleanup before merge

- Leftover debug line at [`components.js:623`](https://github.com/GatherPress/gatherpress/blob/fix/show-exclude-current-event-toggle-only-when-queried/src/variations/core/query/components.js#L623): `// const isShadowSourceContext = false;` (from the `WARNING: DEBUG code` commit). AGENTS.md is strict about no debug/console leftovers.
- The `sprintf` at [`components.js:400-408`](https://github.com/GatherPress/gatherpress/blob/fix/show-exclude-current-event-toggle-only-when-queried/src/variations/core/query/components.js#L400) passes three args (`singularLabel, pluralQueryLabel, singularLabel`) but the format only has two placeholders (`%1$s` is reused). Drop the trailing `singularLabel`; the translator comment only documents 1 and 2 anyway.
- Two identical commits `Move context-enabling filter into correct class` (53a27bf, 9927984) look like a rebase artifact, harmless but worth squashing.

### 4. Scope thought

The `venue_filter` → `shadow_filter` rename plus the `core/query` context plumbing are real generalizations and good, but they widen this PR well past its title ("Exclude current event toggle"). If template support turns into the rabbit hole it looks like, consider landing the two toggle guards + the rename here, and splitting template/context resolution into its own PR tracked against #1689. Keeps this one reviewable and unblocks the actual #1705 fix.

carstingaxion:
**_New issue #1754 _**: The REST requests that sets up the `gatherpress/venue` block with initial context fails, when: 

- the set `sourcePostType` does support `gatherpress-event-date`
- the requested source post has a past date

This has happened with Productions and Seasons as well, because they both use `gatherpress-event-date` AND `gatherpress-shadow-source` supports.

The failing request is only happening inside the editor. 

@mauteri I told you yesterday, that I encountered some reduced-opacity blocks within my tests. This is the reason.

carstingaxion:
**New issue #1755:** Consider renaming `gatherpress/venue` block slug to `gatherpress/source`. 

Everything else can be kept as is, because the default labels speak Venues, what matches the default setting of `sourcePostType`.

carstingaxion:
_**Conclusion:**_ Clean-up & stop here!

As @mauteri  concluded correctly, the problem of #1725 is already solved, everything else should have a new starting point.


- #1753
- #1754 
- #1755

carstingaxion:
> After I updated the production-under-test to carry a realistic date, the whole context post collapsed inside the event.
> 
> Smells like a new problem ...
> 
> image image

<hr>

- https://github.com/GatherPress/gatherpress/issues/1754

mauteri (approved):

Issue #1719, opened 2026-06-01
RSVP button disappears after rsvp


webteam-bd:
### Describe the bug

RSVP API request failed

### Steps to Reproduce

1. Make a event
2. View at frontend and RSVP
3. RSVP button disappears, console RSVP API request failed

### Screenshots, screen recording, code snippet

Screenrecording: https://website.belastingdienst.nl/testen/wp-content/uploads/sites/5/2026/06/Schermopname-2026-06-01-160306.mp4

### Environment information

Chrome
WordPress 7
Linux

### WordPress information

`
### wp-core ###

version: 7.0
site_language: en_US
user_language: en_US
timezone: Europe/Amsterdam
permalink: /%postname%/
https_status: true
multisite: false
user_registration: 0
blog_public: 1
default_comment_status: undefined
environment_type: staging
user_count: 6
dotorg_communication: User has blocked requests through HTTP to the URL: https://wordpress.org.

### wp-paths-sizes ###

wordpress_path: /srv/wordpress
wordpress_size: 65.25 MB (68417393 bytes)
uploads_path: /srv/wordpress/wp-content/uploads
uploads_size: 0.00 B (0 bytes)
themes_path: /srv/wordpress/wp-content/themes
themes_size: 28.00 B (28 bytes)
plugins_path: /srv/wordpress/wp-content/plugins
plugins_size: 28.00 B (28 bytes)
fonts_path: /srv/wordpress/wp-content/uploads/fonts
fonts_size: directory not found
database_size: 16.64 MB (17448960 bytes)
total_size: 81.89 MB (85866409 bytes)

### wp-active-theme ###

name: Twenty Twenty-Five (twentytwentyfive)
version: 1.5
author: the WordPress team
author_website: https://wordpress.org
parent_theme: none
theme_features: core-block-patterns, post-thumbnails, responsive-embeds, editor-styles, html5, automatic-feed-links, widgets-block-editor, block-templates, post-formats, editor-style
theme_path: /srv/wordpress/wp-content/themes/twentytwentyfive

### wp-plugins-active (3) ###

GatherPress: version: 0.34.0-beta 1 develop, author: The GatherPress Community
GatherPress Alpha: version: 0.34.0-alpha.2, author: The GatherPress Community

### wp-media ###

image_editor: WP_Image_Editor_GD
imagick_module_version: Not available
imagemagick_version: Not available
imagick_version: Not available
file_uploads: 1
post_max_size: 1G
[...]

PR #1729, opened 2026-06-03
Harden RSVP front-end against failures and malformed blocks

mauteri:
Addresses GatherPress/gatherpress#1719 ("RSVP button disappears after rsvp" / "RSVP API request failed"), reported on a WP 7.0 / PHP 8.3 staging site behind a proxy, running core `0.34.0-beta.1` alongside GatherPress Alpha `0.34.0-alpha.2`.

The root cause looks environment-specific (malformed/partial block markup and/or a proxy intercepting `/wp-json/` so the response isn't JSON — see the issue thread). These are **defensive fixes** that make the front end degrade gracefully and stop the symptom regardless of the underlying trigger. They are not a substitute for the reporter's raw blocks / Network response, but the first fix alone would have prevented the button from vanishing.

## Why the button disappeared
`renderRsvpBlock` in [view.js](src/blocks/rsvp/view.js) shows the inner block whose `data-rsvp-status` matches `currentUser.status` and hides all others — **with no fallback**. If the saved block is missing the inner block for the status the user transitions to, every state (including the `no_status` button) gets hidden and nothing replaces it.

## Changes
1. **Fallback in `renderRsvpBlock`** — when no inner block matches the current status, re-show the `no_status` block so the RSVP button never fully disappears.
2. **Guard `sendRsvpApiRequest`** ([interactivity.js](src/helpers/interactivity.js)) — `makeRequest` resolves to `undefined` when the nonce can't be fetched and to a non-success/error payload when a proxy/WAF returns non-JSON. Switched the success check to `res?.success` (no more throw on `res.success`) and surface a brief `alert()` on failure, matching the existing rsvp-form error handling, instead of a silent `console.warn`.
3. **Removed two shipped debug `console.log`s** (`[RSVP API] Response received` / `State updated`).
4. **Null-guard the `updateRsvp` success callback** — bail when the `[data-rsvp-status]` wrapper or `.wp-block-gatherpress-rsvp` container isn't in the DOM, so a `null` `closest()` doesn't throw and abort the post-RSVP UI update.

## Testing
- `npm run lint:js` clean on both files.
- Existing `src/blocks/rsvp` unit suites pass (42 tests).
- Both files are in `sonar.coverage.exclusions` (interactivity/view files are exercised via the Playwright `rsvp-flows` e2e, not unit tests), so no unit coverage gate applies. Behavior is best verified against the reporter's raw blocks once provided.

## Follow-ups (not in this PR)
- Get the reporter's raw blocks + the `/rsvp` Network response (status + body) to pin the real trigger.
- The core `0.34.0-beta.1` + Alpha `0.34.0-alpha.2` version mismatch is a strong suspect for the malformed markup — worth asking them to match versions / deactivate Alpha.

### DIFF
--- a/.github/changelog/rsvp-button-disappears-guard
+++ b/.github/changelog/rsvp-button-disappears-guard
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Stop the front-end RSVP button from vanishing when a request fails or the block markup is incomplete. The status switcher now falls back to the no_status state when no inner block matches, the request handler guards against missing/non-JSON responses and surfaces an error instead of failing silently, and the post-RSVP modal logic no longer throws on malformed markup.
--- a/src/blocks/rsvp/view.js
+++ b/src/blocks/rsvp/view.js
@@ -109,11 +109,19 @@ const { state, actions } = store( 'gatherpress', {
 				() => {
 					const parentWithRsvpStatus =
 						element.ref.closest( '[data-rsvp-status]' );
-					const rsvpStatus =
-						parentWithRsvpStatus.dataset.rsvpStatus;
-					const rsvpContainer = parentWithRsvpStatus.closest(
+					const rsvpContainer = parentWithRsvpStatus?.closest(
 						'.wp-block-gatherpress-rsvp',
 					);
+
+					// Bail if the expected wrappers aren't in the DOM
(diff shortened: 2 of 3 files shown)


abdullahkaludi:
@mauteri 
Testing with Playground!

PR #1734, opened 2026-06-03
Guard "Venue Settings" against venue post type being selected

### LINKED ISSUE #1724: Do not expose "Venue settings" if not a gatherpress_venue post type
### Describe the bug

The sidebar panel "Venue settings" gets exposed while using a custom block variation of the `gatherpress/venue` block, to provide context for a shadow-source related production post. 

Image

### block-editor Markup

```html
<!-- wp:gatherpress/venue /-->

<!-- wp:gatherpress/venue {"sourcePostType":"gatherpress_play"} -->
<!-- wp:post-title {"level":3,"isLink":true} /-->

<!-- wp:post-featured-image {"isLink":true} /-->
<!-- /wp:gatherpress/venue -->
```

### Steps to Reproduce

1. Use the [Playground Demo Link](https://playground.wordpress.net/?blueprint-url=https://raw.githubusercontent.com/carstingaxion/gatherpress-productions/main/.wordpress-org/blueprints/blueprint.json) of [gatherpress-productions](https://github.com/carstingaxion/gatherpress-productions)
2. Create a production 
3. Create an event, relate it to the production
4. Use the provided "Productions Details" block (which is a variation of the `gatherpress/venue` block, with `sourcePostType` set to `gatherpress_play`)
5. Select this block to find the sidebarpanel "Venue Settings"

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

carstingaxion:
Fixes #1724

## Proposed changes:

The `gatherpress/venue` block can be used through custom block variations that re-point it at a different shadow-source CPT via the `sourcePostType` attribute (for example the "Productions Details" block in [gatherpress-productions](https://github.com/carstingaxion/gatherpress-productions), which sets `sourcePostType: "gatherpress_play"`). When such a variation is selected in the editor, the sidebar still exposed the **Venue settings** panel (the `VenueNavigator` term picker), even though that control is meaningless for a non-venue source.

This PR guards the `InspectorControls` panel in [`src/blocks/venue/edit.js`](src/blocks/venue/edit.js) so it only renders when the block's resolved `sourcePostType` is an actual venue source, in addition to the existing `! isDescendentOfQueryLoop && ! isInFSETemplate() && isEventContext` conditions.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?

## Testing instructions:

1. Open the [Playground demo](https://playground.wordpress.net/?blueprint-url=https://raw.githubusercontent.com/carstingaxion/gatherpress-productions/main/.wordpress-org/blueprints/blueprint.json) for gatherpress-productions (or run the PR build with the gatherpress-productions plugin active).
2. Create a Production.
3. Create an Event and relate it to the Production.
4. Add the **Productions Details** block (a variation of `gatherpress/venue` with `sourcePostType` set to `gatherpress_play`) to the event.
5. Select the block and open the sidebar.

**Expected:** the **Venue settings** panel is NOT shown for the Productions Details block.

**Regression checks (should be unchanged):**
- A normal `gatherpress/venue` block on an event still shows **Venue settings**.
- Editing a venue post directly still behaves as before (panel hidden there because it is not an event context).

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Guard "Venue Settings" against venue post type being selected

</details>

### DIFF
--- a/.github/changelog/1734-guard-venue-settings-on-venue-information-supporting-types
+++ b/.github/changelog/1734-guard-venue-settings-on-venue-information-supporting-types
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Only show "Venue Settings" panel for post types supporting `gatherpress-venue-information`
--- a/src/blocks/venue/edit.js
+++ b/src/blocks/venue/edit.js
@@ -411,7 +411,7 @@ const Edit = ( props ) => {
 					) ) }
 			</BlockContextProvider>
 			<InspectorControls>
-				{ ! isDescendentOfQueryLoop && ! isInFSETemplate() && isEventContext && (
+				{ 'gatherpress_venue' === sourcePostType && ! isDescendentOfQueryLoop && ! isInFSETemplate() && isEventContext && (
 					<PanelBody
 						title={ __( 'Venue settings', 'gatherpress' ) }
 						initialOpen={ true }
--- a/src/blocks/venue/edit.js
+++ b/src/blocks/venue/edit.js
@@ -151,6 +151,8 @@ const Edit = ( props ) => {
 	// any shadow-source-supporting CPT without renaming.
 	const sourcePostType =
 		attributes?.sourcePostType || getVenuePostType( effectivePostType );
+	// Only a venue-information source has anything to configure in this panel.
+	const isVenueSource = usePostTypeSupports( 'gatherpress-venue-information', sourcePostType );
 	const venuePostType =
 		'venue' === overrideResolution?.kind
 			? overrideResolution.postType
@@ -411,7 +413,7 @@ const Edit = ( props ) => {
 					) ) }
 			</BlockContextProvider>
 			<InspectorControls>
-				{ 'gatherpress_venue' === sourcePostType && ! isDescendentOfQueryLoop && ! isInFSETemplate() && isEventContext && (
+				{ isVenueSource && ! isDescendentOfQueryLoop && ! isInFSETemplate() && isEventContext && (
 					<PanelBody
 						title={ __( 'Venue settings', 'gatherpress' ) }
 						initialOpen={ true }


mauteri:
@carstingaxion can you fill out PR description for more context? Thx!

mauteri:
@carstingaxion here's what's going on with this one.

**Your instinct is right, and the one-liner does fix the exact reported case.** The "Productions Details" block carries `sourcePostType: "gatherpress_play"`, and because [`src/blocks/venue/block.json`](https://github.com/GatherPress/gatherpress/blob/develop/src/blocks/venue/block.json) declares `sourcePostType` with a literal `"default": "gatherpress_venue"`, `attributes.sourcePostType` is always populated. So at [`edit.js:152`](https://github.com/GatherPress/gatherpress/blob/fix/scope-venue-settings-to-venues-only/src/blocks/venue/edit.js#L152) the variation resolves to `gatherpress_play`, the guard `'gatherpress_venue' === sourcePostType` is `false`, and the panel hides.

**The catch is that the check is too strict and re-couples the block to a hardcoded slug**, which is the one thing the post-type-supports architecture in `AGENTS.md` is built to avoid. Any site that registers a custom venue CPT (declaring `gatherpress-venue-information`) and points a venue variation at it with `sourcePostType: "my_custom_venue"` is a legitimate venue source, but `'gatherpress_venue' === 'my_custom_venue'` is `false`, so this guard would wrongly hide **Venue settings** for them. The question we want to answer is not "is the source literally `gatherpress_venue`?" but "does the source post type declare `gatherpress-venue-information`?"

**The tool for that is already in this same file.** Line [`edit.js:103`](https://github.com/GatherPress/gatherpress/blob/fix/scope-venue-settings-to-venues-only/src/blocks/venue/edit.js#L103) does it for the *host* post type:

```js
const isVenueContext = usePostTypeSupports( 'gatherpress-venue-information', context?.postType );
```

Do the same against the *source* type instead of the literal compare:

```js
// Only a venue-information source has anything to configure in this panel.
const isVenueSource = usePostTypeSupports( 'gatherpress-venue-information', sourcePostType );
```

then

```jsx
{ isVenueSource && ! isDescendentOfQueryLoop && ! isInFSETemplate() && isEventContext && (
```

This hides for `gatherpress_play` (no venue support) and shows for any real venue type, including custom ones.

Two notes while you're in there:

1. **Use the hook, not the imperative helper.** `usePostTypeSupports` (reactive), not `isPostTypeSupporting` (a bare `select`). This output drives render, and `AGENTS.md` documents that the non-reactive variant misses the post-type registry's first-render cache miss and leaves dim/visibility-gated blocks stuck. Lines 102-103 already use the reactive form, so this just follows the existing pattern.
2. The `|| getVenuePostType(...)` fallback at [`edit.js:152-153`](https://github.com/GatherPress/gatherpress/blob/fix/scope-venue-settings-to-venues-only/src/blocks/venue/edit.js#L152) is effectively dead for this attribute thanks to the block.json default. It only fires if `sourcePostType` is an empty string, so you're really only ever switching on the attribute value.

To get CI green: this is editor JS, so `npm run build` and confirm in a real editor (a lint + Jest pass alone won't move the panel). The changelog bot is still waiting on `.github/changelog/1734-from-description` (run `composer changelog:add`), and a small JS unit test asserting the visibility predicate for a venue source vs `gatherpress_play` would close the coverage the reviewers will ask about.

mauteri (approved):

PR #1744, opened 2026-06-04
Fix misleading single-event iCal URL in render_alternate_links test fixture

### LINKED ISSUE #1730: Correct the misleading single-event iCal URL in render_alternate_links test fixture
Raised by @carstingaxion in review of #1673: https://github.com/GatherPress/gatherpress/pull/1673#discussion_r3293488773

> This is not a valid URL. And (I hope) this is not generated as it makes no sense. Ping @mauteri

## The flagged line
In `test_render_alternate_links_prints_link_tags()` ([test/unit/php/includes/tests/core/classes/calendar/class-test-setup.php:1293](https://github.com/GatherPress/gatherpress/blob/develop/test/unit/php/includes/tests/core/classes/calendar/class-test-setup.php#L1293)):

```php
array(
    'url'  => 'https://example.org/event/sample/feed/ical/',
    'attr' => 'Example iCal Download',
),
```

A `…/event/sample/feed/ical/` URL labeled **iCal Download** mixes the *feed* URL shape with a *download* label, which is exactly why it reads as nonsensical.

## Investigation — good news: it is NOT generated
The URL is hand-written **test fixture data** passed straight into `render_alternate_links()`, which simply renders whatever `{url, attr}` array it's given — it does not derive URLs. The real code does not emit this shape for a single-event download:

- **Single-event download** comes from `Calendar::get_ical_url()` → `get_endpoint_url( 'ical' )`. Because the slug does not start with `feed/`, it takes the permalink branch ([class-calendar.php:324](https://github.com/GatherPress/gatherpress/blob/develop/includes/core/classes/calendar/class-calendar.php#L324)) and produces **`…/event/sample/ical/`** (no `/feed/`).
- The `…/feed/ical/` shape is only used, correctly, for **feeds** — sitewide, post-type archive, venue/tax-like, and per-term links via `get_post_comments_feed_link()` / `get_term_feed_link()`.

So this is a misleading test fixture, not a production bug — but it's worth fixing so the test documents the real URL contract and doesn't imply a wrong one.

## Proposed fix
- Change the download entry's fixture URL to the real single-event download shape, e.g. `https://example.org/event/sample/ical/` (drop the `/feed/` segment).
- Optionally add a small regression assertion that `Calendar::get_ical_url()` for a single event contains `/ical/` and **not** `/feed/ical/`, locking in download-vs-feed so a future refactor can't silently swap them.

## Acceptance criteria
- The `render_alternate_links` test fixture no longer presents a `/event/<slug>/feed/ical/` URL as an "iCal Download".
- (If added) a test confirms a single-event iCal download URL uses `/ical/`, not `/feed/ical/`.
- Confirmed: no production path emits `/event/<slug>/feed/ical/` for a single-event download.

e1337geek:
Fixes #1730

## Summary

The `test_render_alternate_links_prints_link_tags()` fixture used `/event/sample/feed/ical/` (the feed URL shape) for an entry labelled `"iCal Download"`. Those are two distinct URL contracts in GatherPress:

- **Single-event download** → `…/event/<slug>/ical/` (no `/feed/` segment)
- **Feed** → `…/event/<slug>/feed/ical/`

No production code was broken — `Calendar::get_ical_url()` already emits the correct shape — but the misleading fixture implied the wrong URL contract to anyone reading the test.

**Changes:**
- Corrects the fixture URL from `…/feed/ical/` → `…/ical/`
- Adds `test_get_ical_url_does_not_use_feed_slug()` as a regression lock: asserts the download URL never contains `"feed"`, so a future refactor that accidentally swaps the slugs will produce a named, descriptive failure
- Adds a `patch`-level changelog entry

## Testing steps

1. `npm run build && npm run test:unit:php -- --filter="test_render_alternate_links|test_get_ical_url"`
2. Confirm **3 tests, 8 assertions, OK**
3. Optionally mutate `get_ical_url()` to return `'feed/' . Setup::ICAL_SLUG` and confirm both new assertions fail with clear messages, then revert

### DIFF
--- a/.github/changelog/fix-ical-download-url-test-fixture
+++ b/.github/changelog/fix-ical-download-url-test-fixture
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Correct the misleading `/event/sample/feed/ical/` test-fixture URL in `test_render_alternate_links_prints_link_tags` to the real single-event iCal download shape (`/event/sample/ical/`). [#1730]
--- a/test/unit/php/includes/tests/core/classes/calendar/class-test-calendar.php
+++ b/test/unit/php/includes/tests/core/classes/calendar/class-test-calendar.php
@@ -116,6 +116,29 @@ class Test_Calendar extends Base {
 	}
 
 	/**
+	 * Regression for #1730 — single-event iCal download URL must use the
+	 * `ical` slug directly, never `feed/ical`.
+	 *
+	 * @covers ::get_ical_url
+	 * @covers ::get_endpoint_url
+	 *
+	 * @return void
+	 */
+	public function test_get_ical_url_does_not_use_feed_slug(): void {
+		$instance = new Calendar( $this->make_event() );
+
+		$url = $instance->get_ical_url();
+
+		$this->assertIsString( $url );
+		$this->assertStringNotContainsString(
+			'feed',
+			$url,
+			'Single-event iCal download URL must not contain "feed" — ' .
+			'download uses /ical/ directly; /feed/ical/ is for feed endpoints only.'
+		);
+	}
+
+	/**
 	 * Coverage for get_outlook_url — uses the `outlook` sibling slug pointing
 	 * at the same iCal template.
 	 *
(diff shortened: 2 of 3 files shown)


e1337geek:
Closing as superseded by #1743, which was merged first and includes the same fixture fix plus a more thorough regression test (verifies both the pretty-permalink and plain-permalink URL shapes, not just the absence of 'feed'). Thanks for the review!

PR #1750, opened 2026-06-04
Fix/1733 event date query loop context

### LINKED ISSUE #1733: Inside an Event post: The event-date block relies on the currently edited, but not the contextual post.
### Describe the bug

The `gatherpress/event-date` block seems to consume only the currently edited, but not the contextual `postType` and `postId`.

I looked into the edit component of the block, which uses `context` properly. So I'm even more confused about this buggy behaviour.

This problem seems to appear while **editing an event post**. The dates shown in all queried events, match the one that holds the `core/query` blocks, no matter if nested inside a `gatherpress/venue` or totally pure inside the event itself. This does not happen, when I add an Event Query to a new page.

### Steps to Reproduce

1. ....

Image

_Originally posted by @carstingaxion in https://github.com/GatherPress/gatherpress/issues/1725#issuecomment-4610718361_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

e1337geek:
Fixes #1733

## Problem

When editing an event post that contains a `core/query` block, every `gatherpress/event-date` block inside the query loop displays the dates of the event being edited rather than each queried event's own dates. The bug only appears in the block editor; the published frontend renders correctly.

The root cause was in the `useSelect` callback in `edit.js`. When the editor document is an event post, `isEventPostType()` returns `true` for the entire session. This caused the callback to branch into the live `gatherpress/datetime` store path for every `event-date` block on the canvas — including blocks inside a query loop that should be fetching each queried event's dates from the entity record via `context.postId`.

## Fix

The `useSelect` body was extracted into an exported `resolveEventDateData` helper in a new `src/blocks/event-date/helpers.js` file. This separates the routing logic from the React component and makes it directly unit-testable.

The fix adds two guards to the live-store path:

1. **`undefined === contextQueryId`** — `core/query` sets `context.queryId` to a number on every block in its template. When this value is defined, the block represents a queried post and must fetch from the entity record rather than the live editor store.

2. **`isDirectEditingEvent`** — computed from the reactive `select` parameter using the editor's current document type (not the block context type). This ensures the live-store path is correctly skipped in the site editor, where `context.postType` can be `gatherpress_event` (from the template context) but the editor document type is `wp_template`.

The block's `block.json` already declared `"queryId"` in `usesContext`; it simply was never checked.

## Testing instructions

```bash
npm run build && npm run test:unit:js -- --testPathPattern 'blocks/event-date/helpers'
```

Manual testing steps:

1. Open any published event post in the block editor.
2. Add a Query Loop block to the event's content. Configure it to query `gatherpress_event` post type.
3. Add a `gatherpress/event-date` block to the query template.
4. Verify each queried event shows its own date, not the outer event's date.
5. Confirm the outer event's `event-date` block (placed directly in the content, outside the query) still shows live-updating dates when the date is changed in the sidebar.
6. Open the same query loop on a plain page — confirm dates still render correctly there.

### Other information:

## Notes

- The `resolveEventDateData` extraction also normalizes all return paths to the same five-key shape (`dateTimeStart`, `dateTimeEnd`, `timezone`, `isLoading`, `isValidEvent`), resolving a pre-existing SonarCloud S3801 finding in the original `useSelect` callback.
- The site editor edge case (template context providing `gatherpress_event` post type without a query ID) is covered by a dedicated test and confirmed to skip the live-store path correctly.

### DIFF
--- a/src/blocks/event-date/helpers.js
+++ b/src/blocks/event-date/helpers.js
@@ -1,0 +1,89 @@
+/**
+ * Internal dependencies
+ */
+import { isEventPostType, findEventPostById } from '../../helpers/event';
+
+/**
+ * Resolves the event datetime data for the event-date block editor.
+ *
+ * Extracted from the useSelect callback in edit.js so the routing logic can be
+ * unit-tested independently of React rendering. Called once per render inside
+ * the component's useSelect with the reactive `select` function.
+ *
+ * @since 0.34.0
+ *
+ * @param {Function} select              WordPress data select function.
+ * @param {Object}   context             Block context (postId, postType, queryId).
+ * @param {number}   postId              Resolved post ID (attribute override or context).
+ * @param {boolean}  hasExplicitOverride Whether postId comes from block attributes.
+ *
+ * @return {Object} Datetime data: dateTimeStart, dateTimeEnd, timezone, isValidEvent, isLoading.
+ */
+export function resolveEventDateData( select, context, postId, hasExplicitOverride ) {
+	const postType =
+		context?.postType ||
+		select( 'core/editor' )?.getCurrentPostType();
+
+	const supportsEventDate = !! select( 'core' )
+		.getPostType( postType )?.supports?.[ 'gatherpress-event-date' ];
+
+	if ( ! postId ) {
+		return { isValidEvent: false };
+	}
+
+	// When editing an event directly (not inside a query loop), use the
+	// datetime store for live updates.
+	if ( isEventPostType() ) {
+		const datetimeStore = select( 'gatherpress/datetime' );
+		return {
+			dateTimeStart: datetimeStore.getDateTimeStart(),
+			dateTimeEnd: datetimeStore.getDateTimeEnd(),
+			timezone: datetimeStore.getTimezone(),
+			isValidEvent: supportsEventDate,
+		};
+	}
+
+	// postId override on a host that itself doesn't support event-date
+	// (e.g. a regular page or template part). Resolve the override target
+	// across event-supporting post types so the block can light up.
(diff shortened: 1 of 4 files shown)


carstingaxion:
Cool, you jumped in here, too. Thank you @e1337geek !

I'll have a deeper look today, but this looks promising.

carstingaxion:
I've tested this manually and can confirm, the block is working again!

mauteri (dismissed):

mauteri:
@e1337geek not sure why it didn't write its comment here with coverage report, but missing a little JS coverage, which is causing PR Coverage workflow to fail.

```
Configuration loaded:
  Minimum coverage: 100%
  Exclude patterns: 43 patterns

Checking coverage for JavaScript files changed from origin/develop...

Found 2 changed JavaScript file(s):
  - src/blocks/event-date/edit.js
  - src/blocks/event-date/helpers.js

Files requiring coverage check: 1
  - src/blocks/event-date/helpers.js

❌ src/blocks/event-date/helpers.js - 95.65% coverage (22/23 statements)
   Uncovered lines: 64

=== Summary ===
✅ Passed: 0 file(s)
⏭️  Excluded: 1 file(s)
❌ Low coverage: 1 file(s)
❌ Missing coverage: 0 file(s)

=== JavaScript Files Requiring Additional Test Coverage ===

File: src/blocks/event-date/helpers.js
  Coverage: 95.65% (22/23 statements)
  Required: 100%
  Uncovered lines: 64
  Total uncovered: 1 lines

Please add tests to cover the missing lines in the files listed above.
Minimum required coverage: 100%
```

mauteri (approved):

PR #1757, opened 2026-06-05
Fix initial context for event date supporting types inside venue block

### LINKED ISSUE #1754: gatherpress/venue block does not get initial context
### Describe the bug

The REST requests that sets up the `gatherpress/venue` block with initial context fails, when: 

- the set `sourcePostType` does also support `gatherpress-event-date`
- the requested source post has a event-date in the past

This has happened with _Productions_ and _Seasons_ as well, because they both use `gatherpress-event-date` AND `gatherpress-shadow-source` supports.

The failing request is only happening inside the editor, not on the frontend. 

### Steps to Reproduce

1. Install the Productions companion plugins, create a two entries: one with a past, one with an upcoming event-date (named _Premiere_ here).
   - https://github.com/carstingaxion/gatherpress-productions
2. Edit an **Event**, relate it to one of the productions
3. Add a _**Production Details**_ block, which is a `gatherpress/venue` block with `sourcePostType` set.
4. Inside the editor: Confirm that the context for the production is loaded properly, if you relate the upcoming production to that event
5. Inside the editor: Confirm that the context for the production is NOT loaded properly, if you relate the past production to that event
6. On the frontend: Confirm that context resolution works, no matter of upcoming or past Production.

> After I updated the production-under-test to carry a realistic date, the whole context post collapsed inside the event.
> 
> image image

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

carstingaxion:
>The REST requests that sets up the `gatherpress/venue` block with initial context fails, when: 
>
>- the set `sourcePostType` does also support `gatherpress-event-date`
>- the requested source post has a event-date in the past

Fixes #1754

## Proposed changes:

* Make the `gatherpress/venue` block aware, if the currently requested post type supports `gatherpress-event-date`
* Set the query variable `gatherpress_event_query` to either `'all'` or `null`, for all requests made by the block to find its context post
* Allow the a value of `'all'` for the `gatherpress_event_query` query variable during a REST request

### Other information:

This use case will not apply to a default GatherPress setup. But using the multiple options, GatherPress provides with `post_type_supports`, one can reach such a use case or situation relatively fast. 

>This has happened with _Productions_ and _Seasons_ as well, because they both use `gatherpress-event-date` AND `gatherpress-shadow-source` supports.

A combination that likely does not take place in core GatherPress, but makes a lot of sense as extension point.

## Testing instructions:

1. Install the Productions companion plugins, create a two entries: one with a past, one with an upcoming event-date (named _Premiere_ here).
   - https://github.com/carstingaxion/gatherpress-productions
2. Edit an **Event**, relate it to one of the productions
3. Add a _**Production Details**_ block, which is a `gatherpress/venue` block with `sourcePostType` set.
4. Inside the editor: Confirm that the context for the production is loaded properly, if you relate the upcoming production to that event
5. Inside the editor: Confirm that the context for the production is also loaded properly, if you relate the past production to that event
6. On the frontend: Confirm that context resolution works, no matter of upcoming or past Production

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>

### DIFF
--- a/.github/changelog/1754-fix-initial-context-for-event-date-supporting-types-inside-venue-block
+++ b/.github/changelog/1754-fix-initial-context-for-event-date-supporting-types-inside-venue-block
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fix initial context for event date supporting types inside venue block
--- a/includes/core/classes/blocks/class-event-query.php
+++ b/includes/core/classes/blocks/class-event-query.php
@@ -246,7 +246,7 @@ class Event_Query {
 			? $block_query['postType']
 			: get_post_types_by_support( 'gatherpress-event-date' );
 
-		// Type of event list: 'upcoming' or 'past',
+		// Type of event list: 'upcoming', 'past', or 'all',
 		// @see wp-content/plugins/gatherpress/includes/core/classes/class-event-query.php.
 		$query_args['gatherpress_event_query'] = $block_query['gatherpress_event_query'];
 
@@ -330,7 +330,7 @@ class Event_Query {
 		// Generate a new custom query will all potential query vars.
 		$custom_args = array();
 
-		// Type of event list: 'upcoming' or 'past',
+		// Type of event list: 'upcoming', 'past', or 'all',
 		// @see wp-content/plugins/gatherpress/includes/core/classes/class-event-query.php .
 		$custom_args['gatherpress_event_query'] = $request->get_param( 'gatherpress_event_query' );
 
@@ -410,7 +410,7 @@ class Event_Query {
 
 		// Add custom GatherPress query parameters.
 		$query_params['gatherpress_event_query'] = array(
-			'description' => __( 'Type of events to query: upcoming or past', 'gatherpress' ),
+			'description' => __( 'Type of events to query: upcoming, past, or all', 'gatherpress' ),
 			'type'        => 'string',
 			'enum'        => array( 'upcoming', 'past', 'all' ),
 			'default'     => 'upcoming',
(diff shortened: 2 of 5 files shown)


carstingaxion:
This one is ready @mauteri

mauteri:
Pushed a small follow-up commit (63a323e7) on top of your branch — the approach is solid, this just hardens the JS side:

- **`null` → conditional spread in `useVenuePostFromTermId`**: `buildQueryString` only drops `undefined`; `null` serializes to `gatherpress_event_query=` (empty string). On event-date post types that fails the REST enum validation with a 400 whenever the lookup fires before `getPostType()` has resolved (easy to reproduce on a cold editor load). The param is now spread in conditionally, so it's `all` when the source post type supports `gatherpress-event-date` and absent otherwise.
- **Tests for the new behavior**: the existing mock updates kept the suite green but never exercised the `'all'` branch. Added three cases (supporting, non-supporting, still-resolving post type) — `src/helpers/venue.js` is back at 100% branch coverage.
- **Doc touch-ups**: the REST param description and two inline comments now mention `all` alongside `upcoming`/`past`.

Verified end to end against a local site with a `production` post type (event-date + shadow-source supports): past-dated source post now resolves in the editor request (`gatherpress_event_query=all` → 200 with the post), and the frontend render path was unaffected as you noted. Thanks for the fix — nice catch on the root cause!

mauteri (approved):

carstingaxion:
Cool. Thanks for the addition. I would have thought that the php REST param replacement in the blocks query class would have eliminated the null values, but haven't reached the cold start problem in the editor.

Thanks!

PR #1799, opened 2026-06-13
Lower default Venue Map zoom from 18 to 16 (#1746)

### LINKED ISSUE #1746: Lower default Venue Map zoom from 18 to 15 for better neighborhood context
### Describe the bug

When a venue address is entered and the map focuses on that location, new Venue Map blocks default to zoom level 18. At that level the map is very tight (roughly street-level), so users only see a few blocks and miss useful surrounding context (nearby streets, landmarks, transit, etc.).

Expected: A slightly wider default (e.g. 15) so the pin stays clear but more of the neighborhood is visible without manual adjustment.

Actual: New Venue Map blocks use zoom 18 by default (editor slider and map framing).

This is a UX/default-value change, not a broken feature — the zoom control works; the default is just closer than most users need when previewing a venue.

Site admins can already override this under GatherPress → Settings → Venues → Maps → Default Zoom Level (currently 18 in a fresh install); the request is to change that built-in default so new sites and new blocks start at 15 without requiring a settings change.

### Steps to Reproduce

Edit an event or venue post in the block editor.
Add or open a Venue pattern/group with Venue Address and Venue Map.
Enter a full address (e.g. 1453 3rd Avenue, New York, 10028) so the map geocodes and centers.
Select the Venue Map block and open block settings → Zoom level.
Observe the default is 18 and the map shows a very close view of the immediate block.
Optional: Compare with zoom 15 on the same address — more surrounding area is visible with the pin still readable.

Image

Image

### Environment information

Macbook
tahoe 26.5.1

### WordPress information

7.0

### Code of Conduct

mauteri:
Fixes #1746

## What

Lowers the built-in default Venue Map zoom from **18** to **16** so a freshly geocoded venue shows useful neighborhood context (nearby streets, landmarks, transit) instead of a tight street-level view of one block.

**Why 16 and not 15:** the issue suggested ~15, and we tried it, but at 15 the map read a touch too zoomed out — street names and similar labels weren't reliably visible, which loses the "where exactly is this" readability. **16** is the balance: it pulls back from 18's street-level tightness to show surrounding context while keeping street names and nearby labels legible.

## How

`Venue\Map::DEFAULT_ZOOM` is the single source of truth — changing it `18 → 16` cascades to:

- the **"Default Zoom Level"** setting default (`Settings → Venues → Maps`, via `Map::DEFAULT_ZOOM`),
- the prewarm fallback (`class-prewarm.php`),
- the runtime block-attribute defaults (`apply_block_attribute_defaults()`).

The venue-map `block.json` static `zoom` default is updated to **16** to match.

**Existing sites are unaffected:** the setting is only persisted when an admin explicitly saves a value; otherwise `Settings::get()` resolves the field default (`Map::DEFAULT_ZOOM`) at read time. So sites that saved a custom Default Zoom Level keep it, and everyone else picks up 16 with no migration.

## Tests

- Updated the settings-default assertion (`Test_Venues`) and the venue-map edit-test default-attribute fixture to 16.
- `Test_Venues` / `Test_Prewarm` / `Test_Map` green (39 tests); venue-map JS suite green (45 tests). The `18x300`-keyed descriptor fixtures in `helpers.test.js` are arbitrary cache-key test data, not the default, and are intentionally left unchanged.

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

-   [x] Changed

#### Message

Lowered the default Venue Map zoom level from 18 to 16 so new venue maps show more neighborhood context. Sites that already set a custom Default Zoom Level are unaffected.

</details>

### DIFF
--- a/.github/changelog/1799-from-description
+++ b/.github/changelog/1799-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+Lowered the default Venue Map zoom level from 18 to 16 so new venue maps show more neighborhood context. Sites that already set a custom Default Zoom Level are unaffected.
--- a/includes/core/classes/venue/map/class-map.php
+++ b/includes/core/classes/venue/map/class-map.php
@@ -74,7 +74,7 @@ class Map {
 	 * @since 0.34.0
 	 * @var int
 	 */
-	const DEFAULT_ZOOM = 18;
+	const DEFAULT_ZOOM = 16;
 
 	/**
 	 * Default height (in pixels). Used by the generator and the block.
--- a/src/blocks/venue-map/block.json
+++ b/src/blocks/venue-map/block.json
@@ -12,7 +12,7 @@
 	"attributes": {
 		"zoom": {
 			"type": "number",
-			"default": 18
+			"default": 16
 		},
 		"type": {
 			"type": "string",
--- a/test/unit/js/src/blocks/venue-map/edit.test.js
+++ b/test/unit/js/src/blocks/venue-map/edit.test.js
@@ -143,7 +143,7 @@ import Edit from '@src/blocks/venue-map/edit';
 import { isVenuePostType } from '@src/helpers/venue';
 
 const DEFAULT_ATTRIBUTES = {
-	zoom: 18,
+	zoom: 16,
 	type: 'roadmap',
 	width: 0,
 	height: 300,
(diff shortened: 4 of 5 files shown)


mauteri:
Quick note on the value choice: we initially tried **15**, but at that level the map was a touch too zoomed out — street names and similar labels weren't reliably showing, so it lost the "where exactly is this" readability. **16** turned out to be the sweet spot: it still pulls back from the street-level tightness of 18 to show surrounding neighborhood context, while keeping street names and nearby labels legible. Going with 16 as the new default.

PR #1851, opened 2026-07-03
Harden RSVP frontend against missing state markup

mauteri:
Fixes the frontend half of #1719.

## What happened

On the reporter's site, the RSVP block's `serializedInnerBlocks` attribute is corrupted by a platform-level content filter (one level of backslash escaping stripped on every save), so `json_decode` fails server side and only the currently selected status container renders. The RSVP REST request itself succeeds, but the success handler then does:

```js
actions.openModal( null, rsvpContainer.querySelector( '[data-rsvp-status="attending"] ...' ) );
```

With the attending container missing, that is `openModal( null, null )`, which threw `TypeError: Cannot read properties of null (reading 'target')`, and the catch-all in `sendRsvpApiRequest` reported it as "RSVP API request failed" even though the RSVP was saved. Reproducible with the existing e2e fixture `test/e2e/helpers/create-event-with-rsvp.php` (its `serializedInnerBlocks:""` yields the same single-state DOM), no corrupted content needed.

## Changes

- **modal-manager/view.js**: `openModal` mirrors `closeModal`'s null handling (`event?.target` plus a bail when no element resolves) instead of throwing.
- **rsvp/view.js**: when the attending state's trigger is absent, fully close the current modal (`closeModal( null, ref, true )`) instead of calling `openModal` with null.
- **helpers/interactivity.js**: `onSuccess` runs in its own try/catch; a UI-update failure after a successful request logs `RSVP post-success UI update failed:` instead of firing the request-failure alert. The alert still fires for genuine request failures.
- **rsvp-template/template.js**: restore the centered display name. Newer WordPress moved `core/comment-author-name` text alignment from the `textAlign` attribute to the typography block support, so the template's `textAlign: 'center'` was silently dropped on insert and the name rendered left-aligned. The alignment is now declared both ways (legacy attribute for the WP 6.7 floor, `style.typography.textAlign` for current WordPress); verified via `wp.blocks.createBlock` attribute validation that the new form survives where the old one was stripped. Existing events keep their saved (left-aligned) markup; re-centering there is a one-click toolbar change or a block reinsert.

## Testing

- New unit tests for the `rsvp` and `modal-manager` view stores using a namespace-merging `@wordpress/interactivity` store mock (first tests for these view files), covering all new branches, plus two new `sendRsvpApiRequest` cases (onSuccess invoked; onSuccess throwing does not alert).
- Full Jest suite: 890 passing. Lint clean. `npm run build` (experimental modules) compiles.
- Verified end to end in a browser against a single-state event: one click produces one POST, the RSVP saves as attending, no alert, no console error, modal closes cleanly. Before this change the same click threw the TypeError above and showed the false-failure alert.

### DIFF
--- a/.github/changelog/1851-center-rsvp-display-name
+++ b/.github/changelog/1851-center-rsvp-display-name
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Center the RSVP Response display name again on WordPress versions where the comment author name block moved text alignment to the typography block support.
--- a/.github/changelog/1851-rsvp-missing-state-markup
+++ b/.github/changelog/1851-rsvp-missing-state-markup
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Prevent a successful RSVP from showing a false "RSVP API request failed" alert and a broken modal when the block's state markup is malformed or incomplete.
--- a/src/blocks/modal-manager/view.js
+++ b/src/blocks/modal-manager/view.js
@@ -18,7 +18,13 @@ const { actions } = store( 'gatherpress', {
 				event.preventDefault();
 			}
 
-			element = element ?? event.target;
+			element = element ?? event?.target;
+
+			// Bail when called with neither an event nor an element (e.g. a
+			// querySelector miss at the call site) instead of throwing (#1719).
+			if ( ! element ) {
+				return;
+			}
 
 			const modalManager = element.closest(
 				'.wp-block-gatherpress-modal-manager',
--- a/src/blocks/rsvp-template/template.js
+++ b/src/blocks/rsvp-template/template.js
@@ -38,8 +38,18 @@ const TEMPLATE = [
 							'gatherpress',
 						),
 					},
+					// Centering is declared both ways on purpose: newer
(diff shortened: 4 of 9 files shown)


webteam-bd:
Hello, I am unable to download the zip due to the message Invalid request: The specified URL is not allowed. Is that correct?

carstingaxion:
> Hello, I am unable to download the zip due to the message Invalid request: The specified URL is not allowed. Is that correct?
> 
> 

@webteam-bd Sorry to read, that this download is not working. But for now you can also use beta.2 from yesterday, which includes the changes from this PR.

https://github.com/GatherPress/gatherpress/releases/tag/0.34.0-beta.2

I'll investigate what is going on with the defect download link.

_**EDIT:**_ Since this PR was created, we made one change to our PR Playground preview workflow in #1852, that included a change to the URL of the zipped download. Updating this PRs branch, would update the Playground comment, but as the branch is already deleted, this will not work anymore.

mauteri:
@webteam-bd sorry, yes, older download links will not work due to the new structure, but as @carstingaxion mentioned, we just released beta.2. All new PRs downloads as of #1852 will work as intended.

PR #1863, opened 2026-07-03
Target screenshot PRs at develop with only .wordpress-org changes

mauteri:
Follow-up to #1856. That PR got the screenshot workflow to run end to end for the first time, which exposed the next latent bug: **all six generated screenshot PRs (#1857–#1862) have merge conflicts.**

## Why

The per-locale branch was created with `git checkout -b` from the workflow run's own checkout — which is develop-based — while `gh pr create -B main` targeted main. Develop is currently ~1,080 commits ahead of main (main only moves on release), so every screenshot PR dragged the entire develop delta into a main-targeted PR, and main's own unique commits conflicted. On top of that, `git add -A` also committed the repo-wide image recompression that the Compress step performs, meaning the six PRs would conflict with *each other* after the first merge.

## Changes

- **Screenshot PRs now target `develop`**, matching the repo's branching model (main is release-only); the images reach main with the next release like everything else.
- The Compress step runs **before** the branch switch, so the committed screenshots are the compressed ones.
- The PR branch is cut from `origin/develop` (`git checkout -B … origin/develop` after a shallow fetch), with the generated `.wordpress-org/` directory preserved across the switch via a temp copy and a `git reset --hard` to clear incidental compression modifications that would block the branch switch. This also means a run dispatched from any other ref can't leak that ref's commits into the PR.
- The commit is scoped to `git add .wordpress-org` — each PR contains exactly its locale's screenshot diff and nothing else.
- The generated PRs get the **Skip Changelog** label, since the changelog check enforces an entry on every develop-targeted PR.

## After merge

Re-dispatch the workflow, close #1857–#1862 in favor of the clean replacements, and delete their branches plus the stray `fix/wp-org-screenshots-fr_FR-f944f336…` branch from the earlier quoting-bug run.

### DIFF
--- a/.github/workflows/wordpress-org-screenshots.yml
+++ b/.github/workflows/wordpress-org-screenshots.yml
@@ -201,12 +201,12 @@ jobs:
       # Using "continue-on-error:true" results in 'conclusion' being a success in any case, while the 'outcome' can differ.
       # And, as you already guessed, if: failure() looks at 'conclusion'.
       #
-      # The PR branch is cut from origin/main and carries over ONLY the
-      # screenshot directory. The run's own checkout is develop-based, and a
-      # branch created from it drags every develop-not-in-main commit into the
-      # main-targeted PR — with develop ~1000 commits ahead, that made every
-      # screenshot PR conflict with main (and, via `git add -A` picking up the
-      # repo-wide image recompression, with each other).
+      # The PR branch is cut from origin/develop (screenshots flow into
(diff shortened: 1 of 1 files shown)


carstingaxion:
@mauteri Can you add another review round to make sure the existing docs at .../contributor/screenshot-generation is still accurate with the updated workflow?!

carstingaxion:
Ah ... and ... great you fixed it @mauteri.

PR #1894, opened 2026-07-04
Always screenshot the working tree; drop the wordpress.org install

mauteri:
Settles the open question from #1891 per the release model: **develop is the single source of truth**. On release, develop merges into main carrying the code and its matching screenshots together; the tag cut from main ships wordpress.org assets that depict exactly the released version; develop then moves on to the next version and its screenshots follow.

The old blueprint installed GatherPress **from wordpress.org**, so `.wordpress-org/` screenshots always showed the *previous* release — releasing 0.34.0 would have shipped 0.33.3 imagery. That's structurally incompatible with the model above.

## Changes

- **Shared blueprint** (`.github/scripts/wordpress-org-screenshots/blueprint.json`): `installPlugin` (wp.org slug) → `activatePlugin` (the mounted plugin), and demo data bumped 0.33.0 → **0.34.0** (matching the playground-preview blueprint).
- **Reusable workflow**: always builds and mounts the working tree (`--mount=.:/wordpress/wp-content/plugins/gatherpress`). The `mount-plugin` input is removed — under this model there is no correct `false` case, so the option was only a misconfiguration waiting to happen. The docs suite's separate blueprint is deleted; both callers use the shared one.
- **Contributor docs**: describe the invariant, including the release-checklist implication (dispatch both screenshot workflows before cutting a release so the merge to main carries fresh images).

## Validation

- Locally: the unified blueprint boots Playground with the mounted working tree (verified running 0.34.0-beta.1), imports the 0.34.0 demo data, and the full wordpress.org suite generates all six screenshots (which now include the new Events/Roles settings tabs).
- After merge: dispatching **WordPress.org screenshot generation** regenerates all 8 locales from develop via auto-merging PRs — that's the step that brings `.wordpress-org/` in line with 0.34.0.


carstingaxion:
@mauteri Will this work? 

the old workflow got the language Packs from dot org, based on the Released Version.
An unreleased Version will have Zero language Packs, making the whole workflow useless.

mauteri:
Probably not, I didn't review this yet but did mention that to Claude when it was being generated. I'm at a party so it was just working on this independently.

carstingaxion:
I liked what #1891 did, but this one gets too far.

I think we can probably use the develop version for the English screenshots used in the docs, that would be IMHO the only sense-making use-case.

mauteri:
@carstingaxion I hear you, but the deploy timing is what changes my mind here. The 10up action pushes `.wordpress-org/` to w.org at release time, from the release ref. So if screenshots come from the wp.org install, every release ships the previous version's images. That's how they went stale in the first place. With develop mounted, we dispatch the workflow as part of release prep and the merge to main carries the code and matching screenshots together. Nothing touches w.org between releases, so users never see unreleased UI.

For refreshing screenshots on the current release mid-cycle (like a locale hitting the translation threshold), just dispatch the workflow from `main` instead of `develop`. It mounts whatever ref it runs on, and main is the released code.

Happy to add a toggle back if you feel strongly, but I think the ref you dispatch from already covers it.

Issue #1921, opened 2026-07-10
Automate the release train: pain points from shipping 0.34.0


mauteri:
### Describe the enhancement

Shipping 0.34.0 (2026-07-10) surfaced a list of manual steps and friction points in the release process. This is an umbrella issue to automate them where it makes sense, complementing the tooling work already in this milestone (#1827 retire gatherpress-develop, #1828 props-bot credits, #1758 beta deploys).

### Pain points observed during 0.34.0, and proposed automation

- [x] **Version-bump PRs are hand-assembled across three repos.** The train was: credits entry PR in gatherpress-develop → run `wp gatherpress develop generate_version` in a local WP install → host-side `npm i --package-lock-only` → manually branch/commit/PR in core *and* gatherpress-alpha. Once #1827 ports the generator into `.github/scripts`, wrap it in a `workflow_dispatch` workflow that takes a version input and opens the version PRs in both repos automatically.
- [x] **The develop→main release merge fights branch protection.** `main` had "Require linear history" enabled (with `enforce_admins`), which rejects the release merge commit — it had to be manually unchecked mid-release. Decide the permanent posture: either drop linear-history on `main` (release merges are the only thing that lands there), or document the toggle as an explicit release-day step in `docs/contributor/release-process.md`.
- [x] **The changelog rollup PR gets stuck and can dangle.** The auto-opened `release/X.Y.Z` → develop PR was blocked twice over: its commit is created via git CLI in the workflow, so it is unsigned while `develop` requires signed commits, and merging ultimately needed an admin override. If it is forgotten, the consumed `.github/changelog/*` entries remain on develop and the *next* release double-rolls them. Fixes: create the rollup commit via the GitHub API (API-created commits are signed automatically), enable auto-merge on the PR, and make the release workflow fail fast if a previous `release/*` PR is still open or consumed entry files are present at tag time.
- [x] **Release-branch push races with bots.** calibre/image-actions pushed optimization commits to the release branch mid-flight, rejecting a concurrent push. Consider excluding `version-*`/`release/*` branches from image-actions (optimizations belong on develop PRs), or documenting pull-before-push.
- [x] **Coverage gate re-litigates the whole cycle on release PRs.** The develop→main PR's changed-file set is the entire release diff, so the per-PR coverage check flagged a file no ordinary PR ever surfaced (fixed narrowly in #1918). Consider skipping the coverage check when the base branch is `main` — every file in that diff already passed it on its PR into develop.
[...]

Issue #1920, opened 2026-07-10
Exclude .wp-env.test.json from the distribution package


mauteri:
### Describe the bug

`.wp-env.test.json` (the CI test-environment config added during the 0.34.0 cycle) ships in the WordPress.org distribution because it is missing from `.distignore`. It was noticed in the 0.34.0 pending-release file list on wp.org.

`.distignore` already excludes its sibling `.wp-env.json` (line 56) — the test variant was simply never added. The GitHub release zip is unaffected: `wp-scripts plugin-zip` uses the `files` allowlist in `package.json`, which doesn't include it.

The file is harmless (inert local dev config, no secrets), so 0.34.0 shipped with it as-is rather than recutting the release — same fix-forward approach as #1905 (`patches/`, `jsconfig.json`).

### Steps to Reproduce

1. Look at `tags/0.34.0` in the wp.org SVN (or run the 10up deploy action's rsync locally with `.distignore`).
2. Confirm `.wp-env.test.json` is present in the distributed files.

### Fix

Add `.wp-env.test.json` to `.distignore`, next to the existing `.wp-env.json` entry.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1927, opened 2026-07-11
Deploy beta releases to wp.org trunk for translators

### LINKED ISSUE #1758: Deploy beta releases to trunk on wp.org

It is great, that GatherPress is already translated into 8 languages and 11 locales, but the current workflow makes it hard for translators to have translations in place in time.

Currently a translator can start working on a locale, earliest at the moment where some release is really published as new version on wordpress.org

The WP translation tool allows to translate each tagged plugin version AND trunk, but because GatherPress only releases tagged versions, the develop version is never deployed to SVN trunk. This aligns with the plugin repository guidelines, to not upload every development version, which was also considered as an idea at _10up/action-wordpress-plugin-deploy_, and closed as “not planned”, unfortunately. This is the same action GatherPress uses for its wp.org deployment.

To allow the polyglots teams do their great work:

 **I would suggest, that GatherPress should publishes all its named beta releases to trunk on wp.org, like the current 0.34.0-beta.1 and ….2 using a custom svn deployment.**

This would allow the translators to do their important work, before a new release is finally published AND should probably align with the guidelines, as those publications would not be every development release and as such not pollute the SVN system.

Image

- my question about best practices on that point at Make WordPress polyglots slack: https://wordpress.slack.com/archives/C02RP50LK/p1779211755433659
- https://github.com/10up/action-wordpress-plugin-deploy/issues/103 - the un-planned issue about trunk deployments

### Designs

_No response_

_No response_

### Code of Conduct

mauteri:
Fixes #1758

## Proposed changes:

* New `deploy-wporg-trunk` job in `release.yml`: on **`-beta.*` tags only**, after the GitHub Pre-Release is created, build the plugin, sync it to wp.org SVN **trunk**, and create a matching **SVN tag** (`tags/X.Y.Z-beta.N`) — so the beta appears as a *named download* in the plugin page's [Advanced view](https://wordpress.org/plugins/gatherpress/advanced/) for testers who don't use GitHub (VaultPress-style), while nothing changes for regular sites. This is what lets [translate.wordpress.org](https://translate.wordpress.org/projects/wp-plugins/gatherpress/) pick up new strings in its Development project weeks before the stable release, per @carstingaxion's polyglots research on the issue. Alphas and RCs stay excluded on purpose — the guidance is *named beta releases*, not every dev build.
* **The safety mechanism** (the part that matters): wp.org serves whatever version trunk's `Stable tag:` names, so this job treats that line as the invariant that must never change. Four layers:
    1. The preserved value must be a **strict `X.Y.Z` version** — rejecting empty values, prerelease suffixes, and the literal `trunk` (which would mean wp.org is serving trunk directly, making any sync a live deploy).
    2. That version must **exist as a real SVN tag** (`svn ls tags/X.Y.Z`) — the serve target is verified, not assumed.
    3. After the rsync overwrites readme.txt with the beta's, the Stable tag is re-pinned **and then asserted**: exactly one `Stable tag: X.Y.Z` line must exist or the job aborts *before any commit* — a silent `sed` miss cannot ship.
    4. A **dry-run switch** (`TRUNK_DEPLOY_DRY_RUN=1` repository variable) runs everything except the `svn commit`, printing the full `svn status` and final readme header — set it before the first beta of a cycle, inspect, clear it to go live.

  Net effect: this is not an update channel. Sites see no update offer (their version check compares against the pinned stable tag), and people who explicitly want the beta install the GitHub Pre-Release zip — the trunk sync exists purely so translate.wordpress.org can parse the new strings.
* Custom SVN mechanics (checkout trunk → `rsync -rc --exclude-from=.distignore --delete --delete-excluded` → `svn add`/`rm` → commit) mirror `10up/action-wordpress-plugin-deploy`'s `deploy.sh`, since the upstream action has no trunk-only mode ([10up/action-wordpress-plugin-deploy#103](https://github.com/10up/action-wordpress-plugin-deploy/issues/103), closed not-planned). Trunk's layout therefore stays identical to what stable deploys produce.
* No release-confirmation email is triggered — that flow applies to new version tags, not trunk syncs.
* `docs/contributor/release-process.md`'s pre-release verify list now includes checking the new strings appear on translate.wordpress.org.

### Other information:

- [ ] Have you written new tests for your changes, if applicable? *(Workflow change; verification below.)*
- **Verified by local simulation**: a mock trunk (stable readme + stale files) synced against a mock beta checkout — the served `Stable tag: 0.34.0` survived the rsync that overwrote readme.txt with the beta's, `--delete-excluded` cleaned both stale and `.distignore`d files, and the abort path fired when trunk's tag looked like a prerelease. YAML validates; `lint:md:docs` passes.
- Gated to `github.repository == 'GatherPress/gatherpress'` so forks never attempt it (they lack the SVN secrets anyway).
- **First live run**: the `0.35.0-beta.1` tag. Worth watching that run and then confirming (a) wp.org still serves 0.34.x and (b) translate.wordpress.org's Development project shows the new strings.

## Testing instructions:

* Review the Stable-tag preservation block — it's the load-bearing safety line.
* After merge, nothing happens until a `-beta.*` tag is pushed; alpha tags (like the upcoming `0.35.0-alpha.1`) skip the job entirely — verify it shows as skipped on that run.
* **Recommended first rollout**: set the `TRUNK_DEPLOY_DRY_RUN` repository variable to `1` before tagging `0.35.0-beta.1`; the job logs the preserved Stable tag, full `svn status`, and final readme header without committing. If it all looks right, clear the variable and re-run the job (or wait for beta.2).
* At the first live run: watch for "Preserving Stable tag: 0.34.x (verified tags/0.34.x exists)", then spot-check <https://wordpress.org/plugins/gatherpress/> (unchanged version, no update offered to a 0.34.x test site), translate.wordpress.org (new strings in Development), and the Advanced view's version dropdown (the beta listed as a named download — after its release-confirmation click, one per beta tag).
* Note the beta zips installed from wp.org behave correctly on update: a site running `0.35.0-beta.1` sees no update offer until stable `0.35.0` ships, then updates forward.

My WordPress.org username:

### Changelog entry

*Carries the "Skip Changelog" label — release tooling only.*


.github/workflows/release.yml
```diff
@@ -311,3 +312,122 @@ jobs:
...
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: .nvmrc
+          cache: npm
+
+      - name: Build production assets
+        run: |
+          npm ci
+          npm run build
+
```
carstingaxion, line 357: Shouldn't the distro be build at that point, so we only checkout what .distignore allows ?

mauteri: The build does run before the rsync (the `Build production assets` step above), so `build/` is in place — and the rsync with `--exclude-from=.distignore --delete-excluded` is the same filtering mechanism `10up/action-wordpress-plugin-deploy`'s `deploy.sh` applies on our stable deploys. That parity is the deliberate choice: trunk keeps exactly one file shape whether a beta or a stable last touched it, so the beta→stable transition is a minimal SVN diff instead of shape churn each cycle.

Deploying the distro zip's contents instead would apply the *other* filter — `package.json`'s `files` allowlist (what the GitHub zip uses) — which is a different, narrower set than stable deploys put in trunk. You're right that an allowlist is inherently safer (the `.wp-env.test.json` leak in #1920 is exactly a blocklist gap), but switching filters for betas only would make trunk oscillate between two shapes. If we want allowlist semantics on the wp.org channel, I think the right move is doing it for the stable path too in one coordinated change — happy to file that as a follow-up under #1921 if you'd like.

mauteri:
@jeffpaul would you mind taking a look at this PR for us? Thanks!

.github/workflows/release.yml
carstingaxion: No, ok. Didn't recognise the build steps before.

PR #1930, opened 2026-07-11
Feature/1528 google provider

### LINKED ISSUE #1528: Implement Google Static Maps as a venue map provider
## Background

The `map_platform` setting at `includes/core/classes/settings/class-venues.php` already lets site owners pick between OpenStreetMap and Google Maps. OSM works (it drives the Leaflet interactive map and the CartoDB-tile-composited static map). Google Maps works for the **interactive** path but the **static** path is OSM-only — `Map` just always uses CartoDB tiles regardless of the setting.

Once the Map provider refactor lands (companion issue, milestone 0.34.0), adding Google Static Maps is a single provider class.

## Goal

Implement `GatherPress\Core\Venue\Map\Provider\Google` as the second registered provider. When `map_platform === 'google'`, `Map::ensure_descriptor_for_combo()` calls into it for static map generation.

Unlike the OSM provider (XYZ tile fetch + GD compositing), Google's [Static Maps API](https://developers.google.com/maps/documentation/maps-static/start) returns a single rendered image from one HTTP call:

```
https://maps.googleapis.com/maps/api/staticmap
    ?center={lat},{lng}
    &zoom={zoom}
    &size={width}x{height}
    &scale={1|2}
    &maptype={roadmap|satellite|hybrid|terrain}
    &markers=color:red%7C{lat},{lng}
    &key={api_key}
```

So `Provider\Google::render()` collapses to: build URL, `wp_remote_get` it, decode the PNG bytes into a GD image, return.

### API key

Read the existing Google Maps API key introduced by #1494 (the interactive embed already needs it). Static Maps requires the same key — without one the API responds with `403 Forbidden`. `Provider\Google::render()` reads the key from settings; if it's empty, return null and log via `_doing_it_wrong()`. No new settings field — the key already exists, this issue just consumes it.

### Map type support

Google supports `roadmap`, `satellite`, `hybrid`, `terrain`. The `venue_map_default_type` setting already exists (and is documented as ignored by OSM). `Provider\Google::supports_map_type()` returns true for all four; OSM returns true only for `roadmap`. `Map\Manager` can use this to gate which choices appear in the editor.

### Marker

Use Google's `markers=color:red|{lat},{lng}` URL parameter. No client-side compositing — Google renders the marker server-side. Default red pin matches what OSM stamps today (close enough). Custom marker SVG support is **out of scope** for this issue; tracker as a separate enhancement if anyone asks.

### Retina

Google supports `scale=2` natively (server-renders at 2× density). Cleaner than the OSM zoom-bump trick (`zoom + log2(2)`). `Provider\Google::render( ..., int $density = 1 )` just passes `scale={$density}` through. No special-case logic.

### Attribution

Google Static Maps responses already include the "Google" watermark in the rendered PNG. `Provider\Google::attribution_html()` returns an empty string — no extra HTML attribution block needed (unlike OSM which requires the CartoDB + OpenStreetMap line).

### Error handling

- API key missing → return null from `render()`, log via `_doing_it_wrong()` with a user-visible message in the Maps settings tab.
- API returns non-200 → return null, leave existing descriptor (if any) intact. Don't overwrite a working PNG with nothing.
- API returns 200 but body isn't a valid PNG → same as above.
- Quota exceeded → same. Site stays on whatever PNGs it has; new venues will fall back to interactive until the quota resets.

## Tests

- `Test_Google::test_render_returns_image` — mock a 200 PNG response, assert a `\GdImage` comes back with the requested dimensions.
- `Test_Google::test_render_handles_403` — mock a 403, assert null and that the existing descriptor path doesn't overwrite.
- `Test_Google::test_render_passes_scale_for_retina` — assert `scale=2` lands in the URL when `$density === 2`.
- `Test_Google::test_render_passes_map_type` — assert `maptype` reflects the venue's `gatherpress_venue_map_type` meta.
- `Test_Google::test_attribution_html_is_empty` — Google attribution is in the image itself.
- `Test_Google::test_supports_map_type` — true for `roadmap|satellite|hybrid|terrain`, false otherwise.

Plus an integration test that flips `map_platform` from OSM to Google, saves a venue, and confirms a `venue-google-…png` file appears under `wp-content/uploads/gatherpress/static-maps/`.

## Things to watch for

- **API key in error messages.** Never echo the key back into a notice or log line. Mask it.
- **Cost.** Static Maps is metered ($2/1000 requests after the free tier). Prewarm on theme switch + provider switch can spike usage if a site has many venues. Document in the settings field description that prewarming will issue one request per venue per (zoom, width, height) combo. Consider a soft cap — e.g. don't prewarm more than N requests per minute — if it bites users.
- **License.** Google Static Maps has terms-of-use requirements (e.g. images must be displayed alongside a Google Map link, or in a Google Maps-branded context). The watermark in the image satisfies the visual requirement; site owners are responsible for the rest. Mention in settings copy.
- **Fallback path is already wired** by the refactor issue. Switching providers without re-saving every venue keeps the OSM PNG showing until Google's PNG lands. No additional UX work needed here.

## Out of scope

- **Custom marker icons.** Default red pin only.
- **Map styling JSON.** Google's `style=` param supports custom styling — punt to a future issue.
- **Multiple markers per map.** GatherPress always renders one venue per map.
- **Per-venue provider override.** Setting is global.

## Dependencies

Blocked on two issues — both must land before this one:

1. **Map provider refactor** (companion issue, milestone 0.34.0). This issue assumes `Map\Provider\Base` and `Map\Manager` exist and that `Map::ensure_descriptor_for_combo()` already delegates to the active provider.
2. **#1494 — optional Google Maps API key for full-featured embed.** That issue introduces the API key settings field. This issue reads the same key for Static Maps requests.

## Related

- [Google Maps Static API docs](https://developers.google.com/maps/documentation/maps-static/start).

jmarx:
Fixes #1528 

## Proposed changes:

* Add a Google Static Maps provider for venue map PNG generation, including retina `@2x` images, graceful fallback behavior, provider registration, and PHPUnit coverage.

### Other information:

## Testing instructions:
- In GatherPress → Settings → Venues → Maps, choose Google Maps and paste your API key. In [Google Cloud Console](https://console.cloud.google.com/), enable Maps Static API for that project (and Maps Embed API if you’ll use interactive maps).
`Note: If Static Maps isn’t enabled (or the key is wrong), generation fails silently today — you may still see an old OSM fallback PNG with no admin-facing explanation. A fast follow should surface Google’s error (403 body, missing key, etc.) in settings, on venue edit, and in the Regenerate map snackbar so admins don’t have to debug via uploads/CLI.`

- Edit an event with a venue that has a real address.
- In the venue map block, set Render mode: Static image and save.
- Open the event on the frontend — you should see a static map image (not an iframe; no pan/zoom).
- Right-click the map → Open image in new tab — the URL should look like .../wp-content/uploads/gatherpress/static-maps/59-christopher-st-new-york-ny-10014-google-16-600-300.png.

My WordPress.org username: jmarx75

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [x] Changed - for changes in existing functionality

#### Message

</details>

### DIFF
--- a/includes/core/classes/venue/map/class-setup.php
+++ b/includes/core/classes/venue/map/class-setup.php
@@ -19,6 +19,7 @@ namespace GatherPress\Core\Venue\Map;
 defined( 'ABSPATH' ) || exit; // @codeCoverageIgnore
 
 use GatherPress\Core\Traits\Singleton;
+use GatherPress\Core\Venue\Map\Provider\Google;
 
 /**
  * Class Setup.
@@ -44,7 +45,28 @@ class Setup {
 	 * @since 0.34.0
 	 */
 	protected function __construct() {
+		$this->setup_hooks();
 		$this->instantiate_classes();
+	}
+
+	/**
+	 * Wire companion-style provider registration before siblings boot.
+	 *
+	 * Hooks `gatherpress_register_static_map_providers` so Google registers
+	 * alongside third-party providers rather than inside
+	 * `Manager::register_core_providers()`.
+	 *
+	 * @since 0.35.0
+	 *
+	 * @return void
+	 */
+	protected function setup_hooks(): void {
+		add_action(
+			'gatherpress_register_static_map_providers',
+			static function ( Manager $registry ): void {
+				$registry->register( new Google() );
+			}
+		);
 	}
 
 	/**
--- a/includes/core/classes/venue/map/class-setup.php
+++ b/includes/core/classes/venue/map/class-setup.php
@@ -19,7 +19,6 @@ namespace GatherPress\Core\Venue\Map;
 defined( 'ABSPATH' ) || exit; // @codeCoverageIgnore
 
 use GatherPress\Core\Traits\Singleton;
-use GatherPress\Core\Venue\Map\Provider\Google;
 
 /**
  * Class Setup.
@@ -45,28 +44,7 @@ class Setup {
 	 * @since 0.34.0
 	 */
 	protected function __construct() {
-		$this->setup_hooks();
 		$this->instantiate_classes();
-	}
-
-	/**
-	 * Wire companion-style provider registration before siblings boot.
-	 *
-	 * Hooks `gatherpress_register_static_map_providers` so Google registers
-	 * alongside third-party providers rather than inside
-	 * `Manager::register_core_providers()`.
-	 *
-	 * @since 0.35.0
-	 *
-	 * @return void
-	 */
-	protected function setup_hooks(): void {
-		add_action(
-			'gatherpress_register_static_map_providers',
-			static function ( Manager $registry ): void {
-				$registry->register( new Google() );
-			}
-		);
 	}
 
 	/**
--- a/includes/core/classes/venue/map/class-manager.php
+++ b/includes/core/classes/venue/map/class-manager.php
@@ -3,10 +3,10 @@
  * Venue map provider registry.
  *
  * Singleton registry that owns the provider instances backing the static
- * map pipeline. Built-in providers (OSM today) are registered immediately
- * in the constructor so the registry is usable from any later hook;
- * companion plugins hook the `gatherpress_register_static_map_providers` action
- * (fired on `init` priority 0) to register their own providers on top.
+ * map pipeline. Built-in providers (OSM and Google) are registered
+ * immediately in the constructor so the registry is usable from any later
+ * hook; companion plugins hook the `gatherpress_register_static_map_providers`
+ * action (fired on `init` priority 0) to register their own providers on top.
  *
  * @package GatherPress\Core\Venue\Map
  * @since 0.34.0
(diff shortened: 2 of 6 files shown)


includes/core/classes/venue/map/class-setup.php
```diff
@@ -44,9 +45,30 @@ class Setup {
...
+	 *
+	 * @since 0.35.0
+	 *
+	 * @return void
+	 */
+	protected function setup_hooks(): void {
+		add_action(
+			'gatherpress_register_static_map_providers',
+			static function ( Manager $registry ): void {
+				$registry->register( new Google() );
```
mauteri: Shouldn't this just be registered on `register_core_providers` in `core/classes/venue/map/class-manager.php` along with OSM. This is a core feature, not a companion plugin. You can remove this method and just add it right after OSM like this.

```
231     /**
232      * Register the always-available core providers.
233      *
234      * Called from the constructor so the registry is populated as soon as
235      * the singleton exists. Companion plugins should NOT call this — use
236      * the `gatherpress_register_static_map_providers` action instead.
237      *
238      * @since 0.34.0
239      *
240      * @return void
241      */
242     public function register_core_providers(): void {
243         $this->register( new OSM() );
244         $this->register( new Google() );
245     }
```

mauteri (changes requested):

mauteri (approved):

PR #1934, opened 2026-07-12
Adopt the WordPress 7.0 List View block support on container blocks

### LINKED ISSUE #1815: Review all parent blocks to use the List View block support

> New for WordPress 7.0, block developers can add a "listView": true block supports declaration. This adds a List View tab to the block inspector with a dedicated List View UI for the block that allows users to easily rearrange and add inner blocks. This List View is also displayed in Patterns and is recommended for any block that acts as a container for a list of child blocks.
> 
> ```json
> {
>   "supports": {
>     "listView": true
>   }
> }
> ```
>
>https://make.wordpress.org/core/2026/03/15/pattern-editing-in-wordpress-7-0/

> ...**recommended for any block that acts as a container for a list of child blocks**

As the text states, this should be reviewed for the following of GatherPress' blocks:

- `gatherpress/modal-content`
- `gatherpress/online-event`
- `gatherpress/rsvp`
- `gatherpress/rsvp-form`
- `gatherpress/rsvp-response`
- `gatherpress/rsvp-template`
- `gatherpress/venue`

### Designs

_No response_

_No response_

### Code of Conduct

mauteri:
Fixes #1815

## Proposed changes:

* Declare `"listView": true` on all seven blocks from the issue — `modal-content`, `online-event`, `rsvp`, `rsvp-form`, `rsvp-response`, `rsvp-template`, `venue` — plus `add-to-calendar`, which the review surfaced as having the identical guarded-container profile (icon + dropdown inner blocks) and was only missing from the original list.
* **Block Guard now covers the new inspector tab.** The review's key finding: the WP 7.0 List View tab renders a guarded block's inner blocks as a standalone tree in the inspector, bypassing both the canvas `inert` overlay and the existing main-List-View handling — with Block Guard ON, clicking a row selected the inner block (verified empirically in wp-env before writing any fix). New `applyInspectorListViewGuard()` applies the same `inert` treatment as the canvas container to the inspector tree while the guarded block is selected, wired into the existing MutationObserver cycle and gated on `isSelected` so unselected instances never fight over the shared inspector DOM.
* Net behavior: while guarded, the tab stays visible as a **read-only structure map** (nice orientation aid); toggle Block Guard off and the tab becomes the full rearrange/add UI — which pairs well, since the tab is most useful exactly when intentionally editing inner layout.
* Blocks considered and left out: `modal` and `modal-manager` (structural wrappers whose children are wired to trigger behavior, not rearrangeable lists) and `dropdown` (a UI primitive; can revisit if there's demand).

### Other information:

- [x] Have you written new tests for your changes, if applicable? — Four new Jest cases for `applyInspectorListViewGuard` (tab absent, enabled, appender hidden, released); suite at 894 passing.

## Testing instructions:

* `npm run build`, create an event with the "Event with RSVP" starter pattern on WP 7.0.
* Select the RSVP block → the inspector shows a List View tab with the block's inner tree. With Block Guard ON (default), rows are non-interactive (dimmed, no selection on click, appender hidden).
* Toggle Block Guard off in the Settings tab → the List View tab rows select inner blocks and the appender returns.
* Select an unguarded container (e.g. insert an RSVP Form) → its List View tab is fully interactive with no guard involved.

### DIFF
--- a/.github/changelog/1815-list-view-block-support
+++ b/.github/changelog/1815-list-view-block-support
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Container blocks (Add to Calendar, Modal Content, Online Event, RSVP, RSVP Form, RSVP Response, RSVP Template, Venue) now declare the WordPress 7.0 List View block support, and Block Guard extends to the new inspector List View tab so guarded blocks stay protected there.
--- a/src/blocks/add-to-calendar/block.json
+++ b/src/blocks/add-to-calendar/block.json
@@ -8,13 +8,18 @@
 	"icon": "calendar",
 	"example": {},
 	"description": "Allows a member to add an event to their preferred calendar.",
-	"usesContext": [ "postId", "postType", "queryId" ],
+	"usesContext": [
+		"postId",
+		"postType",
+		"queryId"
+	],
 	"supports": {
 		"gatherpress": {
 			"blockGuard": true,
 			"postIdOverride": true
 		},
-		"html": false
+		"html": false,
+		"listView": true
 	},
 	"textdomain": "gatherpress",
 	"editorScript": "file:./index.js"
--- a/src/blocks/modal-content/block.json
+++ b/src/blocks/modal-content/block.json
@@ -4,7 +4,9 @@
 	"name": "gatherpress/modal-content",
 	"version": "1.0.0",
 	"title": "Modal Content",
-	"parent": ["gatherpress/modal"],
+	"parent": [
+		"gatherpress/modal"
+	],
 	"category": "gatherpress",
 	"icon": "external",
 	"example": {},
@@ -48,7 +50,8 @@
 				"style": true,
 				"width": true
 			}
-		}
+		},
+		"listView": true
 	},
 	"textdomain": "gatherpress",
 	"editorScript": "file:./index.js",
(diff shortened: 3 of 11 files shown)


mauteri:
Follow-up per discussion: the guarded inspector tree now explains itself. A notice above the tree reads "Block Guard is enabled." with an **Unprotect** button that flips the same shared state as the sidebar toggle (both stay in sync, query-loop siblings included). The notice sits outside the inert subtree so it stays clickable, and it disappears the moment the guard is off. Verified the full loop in wp-env: notice renders → Unprotect click releases the tree, removes the notice, keeps the block selected → sidebar toggle reads off → re-enabling the toggle brings the notice back. Suite at 897 passing with three new cases (single injection, button wiring, removal on disable).

mauteri:
Reworked per feedback: the custom notice is gone — the List View tab now renders the actual Block Guard ToggleControl (same label, help text, and shared state as the Settings tab control), repeated UI as discussed. Both toggles now float to the top of their tabs via a small injected stylesheet, so the protection state is the first thing you see in either tab. Also fixed the inspector tree's appender: it uses its own list-view-appender class, so it's now hidden while guarded just like the canvas plus button, instead of being visible-but-dead. Verified in wp-env with DOM measurements: guard panel above the tree in the List View tab, topmost panel in Settings, tree inert + appender hidden when guarded, all interactive when unguarded, both toggles in sync. Suite at 896.

Issue #1816, opened 2026-06-18
Ensure attributes that represent a block’s content have "role": "content" set in block.json


carstingaxion:
### Is your enhancement related to a problem? Please describe.

I created this as an _Enhancement_ issue, just because I haven't found a related _bug_ yet. But surely, this WP core change has the potential to become buggy on the GatherPress site, sooner or later. 

Related to #1815 

> ### What developers need to do
> 
> #### Block authors
> 
> If your block is nested in a contentOnly pattern and should be editable, ensure attributes that represent a block’s content have `"role": "content"` set in `block.json`. This is unchanged from WordPress 6.7, but is now more important as `contentOnly` mode is applied more broadly by default.
> 
> ```json
> {
>   "attributes": {
>     "url": {
>       "type": "string",
>       "role": "content"
>     },
>     "label": {
>       "type": "string",
>       "role": "content"
>     }
>   }
> }
> ```
> 
> **Blocks without any `"role": "content"` attributes will be hidden from List View and non-selectable inside a `contentOnly` container.**
> 
> At times a block may not have an appropriate attribute to which to apply `"role": "content"`. A `"contentRole": true` property can be added to the block supports declaration, and this has the same effect as `"role": "content"`.
> 
> ```json
> {
>   "supports": {
>     "contentRole": true
>   }
> }
> ```
> 
> Developers should prefer `"role": "content"` where possible.
> 
> #### Parent / child contentOnly blocks
> 
> Many blocks are considered ‘content’, but consist of both parent and child block types. Some examples of Core blocks are:
> 
> - List and List Item
> - Gallery and Image
> - Buttons and Button
> 
> Whenever both a parent and child block have a `"role": "content" `attribute or `"contentRole": true` block supports, `contentOnly` mode allows insertion of child blocks. This behavior has been present since WordPress 6.9, but is now more prominent.
> 
> Block developers can take advantage of this behavior.
> 
> #### List View block support
> 
> New for WordPress 7.0, block developers can add a `"listView": true` block supports declaration. This adds a List View tab to the block inspector with a dedicated List View UI for the block that allows users to easily rearrange and add inner blocks. This List View is also displayed in Patterns and is recommended for any block that acts as a container for a list of child blocks.
> 
> ```json
> {
>   "supports": {
>     "listView": true
>   }
> }
> ```
> 
>https://make.wordpress.org/core/2026/03/15/pattern-editing-in-wordpress-7-0/

### Code of Conduct

[...]

PR #1935, opened 2026-07-12
Add the isLink attribute to the event-date block

### LINKED ISSUE #1726: Introduce isLink attribute to event-date block

Sometimes it may be totally ok to only show the dates and times of some upcoming events. The post title can potentially omitted, when the context is clear.

This could be the case on a (shadow-source connected) Workshop posttype, with an event query that lists all occurrences of that workshop. There would be no need to re-render “Learning to cook” 10 times, when just the dates matter in such a list.

Now to the point: The `gatherpress/event-date` block should get a new attribute `isLink`, that when set, renders the date as link to the event (supporting post type).

This would align with the core/post-date block, that uses the same attribute.

### Designs

_No response_

_No response_

### Code of Conduct

mauteri:
Fixes #1726

## Proposed changes:

* Add an `isLink` attribute to `gatherpress/event-date`, mirroring `core/post-date`: `boolean`, `default: false`, `"role": "content"` (matching core's attribute shape exactly, which also serves #1816's role-annotation effort for this block).
* **Frontend:** when set, `render.php` wraps the formatted datetime in `<a href="{event permalink}">`, resolved through the block's existing post-id handling (so post-id overrides and query-loop context both link to the right event).
* **Editor:** a "Link to event" toggle in the Display Settings panel, and the same pseudo-link treatment core uses (`href="#event-date-pseudo-link"` with `preventDefault`) so the date looks like a link without navigating the editor.
* The block already declares `color.link` support, so link color controls apply to the new anchor out of the box.
* Use case from the issue: shadow-source connected post types (e.g. a Workshop listing all its occurrences) can show bare linked dates without repeating the post title ten times.

### Other information:

- [x] Have you written new tests for your changes, if applicable? — Two PHP render tests (anchor present with `isLink`, absent by default) and five Jest cases for the Edit component (pseudo-link render both ways, preventDefault, toggle on/off). Suites: 1526 PHP-side in the block's class run green, 901 JS.

## Testing instructions:

* Insert an Event Date block in an event (or an Event Query loop), open the block's Display Settings, and enable "Link to event".
* In the editor the date renders as a link that doesn't navigate; on the frontend the datetime is wrapped in a link to the event's permalink.
* Toggle it off — the date renders as plain text again.

### DIFF
--- a/.github/changelog/1726-event-date-islink
+++ b/.github/changelog/1726-event-date-islink
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+The Event Date block supports a new isLink attribute, matching the core Post Date block: toggle "Link to event" in the block settings to render the date as a link to the event.
--- a/src/blocks/event-date/block.json
+++ b/src/blocks/event-date/block.json
@@ -22,6 +22,11 @@
 			"type": "string",
 			"default": "both"
 		},
+		"isLink": {
+			"type": "boolean",
+			"default": false,
+			"role": "content"
+		},
 		"startDateFormat": {
 			"type": "string",
 			"default": ""
--- a/src/blocks/event-date/edit.js
+++ b/src/blocks/event-date/edit.js
@@ -292,7 +292,7 @@ const Edit = ( { attributes, setAttributes, context } ) => {
 			</BlockControls>
 			{ isLink ? (
 				<a
-					href="#event-date-pseudo-link"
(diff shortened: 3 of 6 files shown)


carstingaxion:
@mauteri Could you file a small addition to this? I'd like to suggest including the new attribute into the block transform, especially because the already supported core/post-date block has the same attribute, that can be easily inherited.

PR #1938, opened 2026-07-12
Stand Block Guard down where core already restricts editing

### LINKED ISSUE #1817: Review blocks about their "Block Guard" vs contentOnly behavior

Especially because of the existing "Block Guard" (#1137 ) feature, but not strictly related, GatherPress should review all blocks that support it, because `Patterns that previously relied on unrestricted editing of their inner blocks will now be presented to users in contentOnly mode by default.`

> ### Review your registered patterns and consider:
> 
> 1. Testing that the content users are expected to change is accessible in contentOnly mode.
> 2. Auditing patterns containing Buttons, List, Social Icons, and Navigation blocks specifically — these have had targeted contentOnly improvements and may behave differently than before.
> 3. Restrict the allowed blocks if users shouldn’t be able to insert blocks in a specific area of a pattern. If assembling a pattern in a block editor, this can be done using the ‘Manage allowed blocks’ feature in the Advanced section of the block inspector for any blocks that have "allowedBlocks": true block support. Through code, the "allowedBlocks":[] attribute can be added to prevent insertion of inner blocks.
> 
> https://make.wordpress.org/core/2026/03/15/pattern-editing-in-wordpress-7-0/

### Designs

_No response_

_No response_

### Code of Conduct

mauteri:
Fixes #1817

## Proposed changes:

This is the actionable outcome of the #1817 review. Full audit findings:

* **The conflict (confirmed empirically):** inside a `templateLock: "contentOnly"` container, core assigns a guarded GatherPress block the `disabled` editing mode and its inner text blocks `contentOnly` — core's promise is that the text stays editable. Block Guard's `inert` overlay broke that promise, and because core removes both selection routes to a disabled block (canvas clicks select nothing; the block is hidden from List View entirely), the Block Guard toggle was unreachable. Net effect: content inside guarded blocks in locked containers was permanently uneditable with no UI escape.
* **The fix:** the guard stands down whenever core already restricts the block's context — any ancestor with a contentOnly template lock, or a synced pattern instance (`core/block`) ancestor. Core's protection is stronger and more granular there; stacking ours on top only subtracted. The guard resumes automatically in normal contexts.
* **Implementation note (the tricky part):** the restriction check runs imperatively inside each guard application pass instead of as a hook. Core freezes rendering of blocks it resolves to a non-default editing mode, so both `getBlockEditingMode` and `useBlockEditingMode` report a stale `default` forever from inside the frozen component — verified with render instrumentation. The guard's MutationObserver passes keep firing on DOM churn regardless of React, and read the live registry each time, so the guard releases as soon as the lock resolves.

Other review findings, for the record: synced pattern instances are fully `disabled` by core (guard now stands down there too — consistent, and pattern overrides are core-blocks-only anyway); our registered starter patterns detach to `default` mode on insertion, unaffected; `allowedBlocks` coverage already exists where it matters (modal in block.json, dropdown and rsvp-form in JS); and the `role: "content"` annotation gaps surfaced by this audit are #1816's worklist.

### Other information:

- [x] Have you written new tests for your changes, if applicable? — Four unit tests for `isCoreRestrictedBlock` (no ancestors, unrestricted ancestors, contentOnly lock, synced pattern); block-guard suite at 53, full JS suite at 905.

## Testing instructions:

* Create an event, add a Group block, set its template lock to contentOnly via code or `templateLock` attribute, and place an RSVP block inside plus a plain paragraph as control.
* Before this fix: the paragraph edits, but everything inside the RSVP block is dead — no clicks, absent from List View, toggle unreachable.
* After: text inside the RSVP block (button labels, status text) behaves exactly like the control paragraph — editable per contentOnly, structure protected by core.
* Outside any lock, Block Guard behaves exactly as before (default on, toggle in both inspector tabs).

### DIFF
--- a/.github/changelog/1817-guard-defers-to-core-restrictions
+++ b/.github/changelog/1817-guard-defers-to-core-restrictions
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Block Guard now stands down inside contentOnly template locks and synced pattern instances, where core already protects structure but deliberately keeps text editable — previously the guard made that content uneditable with no way to reach the toggle.
--- a/src/supports/block-guard.js
+++ b/src/supports/block-guard.js
@@ -5,6 +5,7 @@ import { getBlockType } from '@wordpress/blocks';
 import { createHigherOrderComponent } from '@wordpress/compose';
 import { InspectorControls } from '@wordpress/block-editor';
 import { PanelBody, ToggleControl } from '@wordpress/components';
+import { select } from '@wordpress/data';
 import { __ } from '@wordpress/i18n';
 import { addFilter, hasFilter } from '@wordpress/hooks';
 import { useState, useEffect, useCallback } from '@wordpress/element';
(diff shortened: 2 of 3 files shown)


mauteri:
Closing this for now. The fix works as designed, but I'm not comfortable with the resulting UX: the guarded block ends up unlocked in the event view (inside the locked container) but locked in the pattern editor. Full reasoning in my comment on #1817. Keeping the branch for reference in case we revisit.

PR #1941, opened 2026-07-12
Venue map: core dimensions support; venue: let contents expand with wide/full alignments

### LINKED ISSUE #1818: Use core dimensions support, instead of width & height attributes for venue-map block

The `gatherpress/venue-map` block uses custom block attributes `width` and `height`, but could better use WP core dimensions support for that. 

> WordPress 7.0 expands the Dimensions block supports system with three significant improvements: width and height are now available as standard block supports under dimensions, and themes can now define dimension size presets to give users a consistent set of size options across their site.
> 
> https://make.wordpress.org/core/2026/03/15/dimensions-support-enhancements-in-wordpress-7-0/

### Designs

I suggest updating the `block.json` with:

```json
{
    "supports": {
        "dimensions": {
            "height": true,
            "width": true,
            "__experimentalSkipSerialization": true,
            "__experimentalDefaultControls": {
                "height": false,
                "width": false
            }
        }
    }
}
```

_No response_

### Code of Conduct

mauteri:
Closes #1818

Two related changes, folded into one PR so a single branch/build covers testing.

## Venue Map — core dimensions support, height-only (#1818)

The sizing model: **the map always fills its container** — width comes from the column and the block's alignment (wide/full for larger layouts), never from a stored value. Same model as core's Cover block.

- **Height is the only stored dimension**, via `supports.dimensions` with `__experimentalSkipSerialization` — a CSS string in `style.dimensions.height`, and the block owns its own style output on both sides (editor wrapper + `render.php`).
- **Resizing is the single bottom drag handle** (or the core Dimensions panel's Height control, opt-in via the panel's options menu). Dragging commits an explicit height, which takes over from the aspect ratio until reset. Aspect ratio (and Scale, in static mode) join the same panel via `InspectorControls group="dimensions"`.
- **Height precedence**: block value → Settings → Venues default height → aspect ratio shapes the block. The Settings default reaches rendering through the dimension-resolution paths on both sides (no more attribute-default stamping). The **Default Width setting is removed** — it has no meaning in this model.
- **Any CSS unit works for display**; only px-expressible heights feed the static-map PNG pipeline, which derives its width from height × ratio as always. `render.php` mirrors the editor's sizing rules and runs style values through `safecss_filter_attr()`.
- **No legacy fallbacks in core**: the old numeric `width`/`height` attributes are gone entirely. GatherPress/gatherpress-alpha#52 migrates saved content (heights carried, widths dropped); unmigrated maps render at the default size until the fix runs — the standard alpha contract.

## Venue — contents expand with wide/full alignments

Wide/full venue alignment looked broken: the venue expanded but its contents stayed clamped at content width, and the frontend disagreed with the editor.

- **Layout default switches from `constrained` to flow** (`"layout": true`). Contents track the venue's own width in every alignment — matching what the frontend has always rendered (see below), so existing sites see no change. The layout panel's content-width toggle re-enables constrained per block.
- **The context fix moves into `render.php`** — the old `render_block_gatherpress/venue` filter re-rendered the inner blocks *after* core had decorated the wrapper, then rebuilt a bare div and silently dropped everything block supports had added (layout classes, global padding, container ids) — which is why the old constrained default never applied on the frontend. Now the render callback resolves the shadow-source post, re-renders the inner blocks with it passed as explicit available context, and emits the one real wrapper for core's block-supports pipeline to decorate normally. No filter, no wrapper surgery, no `core/null`/`render_block_context` gymnastics — and nested context providers inside a venue (e.g. a query loop) override correctly again instead of being clobbered by the old `PHP_INT_MAX` context filter.
- Stored `constrained` layout attributes on existing venue blocks are migrated away by GatherPress/gatherpress-alpha#52.

## Testing

- PHP: `Test_Map_Dimensions` covers the `Venue\Map\Dimensions` helpers branch-by-branch plus `do_blocks` render assertions for every sizing rule (explicit height, Settings default, block-beats-default precedence, non-px units, safecss rejection, ratio-only, hydration payload heights — and that width is never stamped). Venue wrapper-preservation tests in `Test_Venue`.
- Jest: bottom-only handle, auto-vs-explicit box sizing, height commit on resize, unstyled block wrapper, style-only dimension reads.
- Full suites green (1552 PHP + 315 multisite, 918 JS, lint/PHPStan clean). Verified live in wp-env and on a Lando site through several rounds of feel-testing.


carstingaxion:
That is a lot of code to opt into core supports @mauteri .
Are you really, really sure we need to rewrite context handling to support dimension attributes? This seems overkill.

mauteri:
@carstingaxion there was a lot of issues and some things discovered that needed to be fixed. I spent a lot of time testing and adjusting the UI to make this work as intended.

PR #1942, opened 2026-07-13
Let venue block contents expand with wide and full alignments

mauteri:
Fixes the venue block behaving oddly with wide/full alignment: the block could be set wide or full, but the contents inside never expanded with it.

## Root cause (two stacked bugs)

1. **The venue block declared a constrained inner layout** (`supports.layout.default.type = "constrained"`, since 42e8f1a6 / 0.34.0 dev). Constrained layout caps children at the theme's content width — so in the **editor**, a wide/full venue expanded while every child (detail groups, map) stayed clamped at `contentSize` and centered.
2. **The frontend never applied that layout at all.** `Blocks\Venue::render_with_source_context()` re-renders the inner blocks with the venue's post context and then rebuilt the wrapper `<div>` by hand with only the base class + align + className. Since the block-specific `render_block_gatherpress/venue` filter runs *after* the generic `render_block` filters, everything core's block-supports pipeline had injected into the real wrapper (`is-layout-*`, `wp-container-*`, `has-global-padding`) was silently thrown away. Editor and frontend disagreed with each other.

## Fix

- **Layout default switches from `constrained` to flow** (`"layout": true`). Contents now track the venue's own width in every alignment — which is exactly what the frontend has always rendered, so existing sites see no frontend change. Unaligned venues are visually identical (the constraint only ever manifested when the venue was wider than `contentSize`). The layout panel's "Inner blocks use content width" toggle remains available for anyone who wants constrained behavior back — and with the second fix it now actually works on the frontend.
- **`render_with_source_context()` reuses the first-pass wrapper tag verbatim** and swaps only the inner markup, so alignment, custom classes, and core's layout/block-supports output survive the context-corrected re-render. A minimal rebuilt wrapper remains as a fallback when the original markup doesn't open with a wrapper tag.

## Testing

- Updated the wrapper-shape tests to model the real first-pass markup (align, className, and layout classes preserved; first-pass inner markup replaced by the context-corrected render), added a fallback-branch test, and fixed stale `@covers ::render_with_venue_context` annotations pointing at the renamed method.
- Suites green: 1538 PHP + 315 multisite, 901 JS, lint/PHPStan clean.
- Verified live in wp-env: frontend venue wrapper now carries `is-layout-flow wp-block-gatherpress-venue-is-layout-flow` (previously bare `wp-block-gatherpress-venue alignfull`), editor children show `max-width: none` instead of the 645px clamp, and an unaligned venue renders pixel-identical to before.

### DIFF
--- a/.github/changelog/venue-wide-full-layout
+++ b/.github/changelog/venue-wide-full-layout
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Venue blocks set to wide or full width now let their inner content expand with them: the block's inner layout no longer defaults to content-width constraint, and the frontend wrapper keeps the layout and block-supports classes WordPress generates instead of rebuilding a bare div.
--- a/includes/core/classes/blocks/class-venue.php
+++ b/includes/core/classes/blocks/class-venue.php
@@ -93,7 +93,7 @@ class Venue {
 		}
 
 		// Has source post — render with source context.
-		return $this->render_with_source_context( $source_post, $instance );
+		return $this->render_with_source_context( $source_post, $instance, $block_content );
 	}
 
 	/**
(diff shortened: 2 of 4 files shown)


mauteri:
Folded into #1941 — the two changes are related (the venue layout fix is what makes wide/full venues give the resized map room) and testing them from one branch is much easier. The commit landed there verbatim as 56f447e9.

PR #1944, opened 2026-07-13
Fix: stop force-loading wp-date/moment.js on every frontend page

lhero-org:
## What

`Assets::enqueue_timezone_shim()` is hooked on both `admin_enqueue_scripts` and `wp_enqueue_scripts`. On the frontend it guards with:

```php
if ( ! wp_script_is( 'wp-date', 'registered' ) ) {
    return;
}
```

`wp-date` is registered by WordPress core on every request regardless of whether the page actually uses it, so this guard never blocks anything on the frontend. The result is that `wp_enqueue_script( 'wp-date' )` fires unconditionally on **every** frontend pageview — including pages with no GatherPress block at all — and since core's `wp-date` bundle depends on `moment`, that drags `moment.min.js` (plus `hooks.min.js` and `deprecated.min.js`) onto every page of a GatherPress-enabled site.

We found this while auditing frontend script weight on a community site (~400 plugins, GatherPress network-activated) and noticed `wp-date`/`moment` loading on a plain blog homepage with zero event/RSVP content.

## Fix

- On the frontend, check `wp_script_is( 'wp-date', 'enqueued' )` instead of `'registered'`, and never call `wp_enqueue_script( 'wp-date' )` ourselves outside `is_admin()`. This only patches the timezone setting when something else (a GatherPress block, another plugin) has already enqueued `wp-date` for the request — it no longer forces the load.
- Moved the `wp_enqueue_scripts` registration to priority `100` so the `'enqueued'` check reflects enqueues made earlier in the request by blocks/plugins running at default priority.
- Admin/block-editor behavior (the original Moment Timezone console-warning fix) is unchanged — `wp-date` is still registered on essentially every admin screen there, so the existing `'registered'` check and unconditional enqueue stay as-is for `is_admin()`.

## Testing

Manually verified against a live Multisite install:
- Homepage with no GatherPress blocks: `wp-date`/`moment`/`hooks`/`deprecated` no longer present in the page's enqueued scripts.
- Event/RSVP pages: `wp-date` still enqueued (by the relevant GatherPress block) and the timezone shim still applies — no change in editor console-warning behavior.

Happy to add a unit test around the `is_admin()` branch if that's wanted — let me know the preferred approach given the existing test setup for this class.


includes/core/classes/class-assets.php
```diff
@@ -101,7 +101,10 @@ protected function setup_hooks(): void {
 		// Set priority to 11 to not conflict with media modal.
 		add_action( 'admin_footer', array( $this, 'event_communication_modal' ), 11 );
 		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_timezone_shim' ) );
-		add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_timezone_shim' ) );
+		// Late priority (100) so that on the frontend this runs after any
+		// block/script that might enqueue wp-date for this request — see
+		// the is_admin()/enqueued-vs-registered branch in the callback.
+		add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_timezone_shim' ), 100 );
```
mauteri: If it should come after anything (and never before) best to use `PHP_INT_MAX` so it's always last.

includes/core/classes/class-assets.php
```diff
@@ -129,7 +147,16 @@ protected function setup_hooks(): void {
...
+				return;
+			}
+
+			wp_enqueue_script( 'wp-date' );
+		} elseif ( ! wp_script_is( 'wp-date', 'enqueued' ) ) {
+			// Frontend: never force-load wp-date (and moment.js) on pages
+			// that don't otherwise need it. Only patch the timezone
+			// setting if another block/script already enqueued wp-date.
 			return;
 		}
```
mauteri, line 158: Maybe something like this to avoid nested if statements.
```
$is_admin = is_admin();

if ( ! wp_script_is( 'wp-date', $is_admin ? 'registered' : 'enqueued' ) ) {
	return;
}

if ( $is_admin ) {
	wp_enqueue_script( 'wp-date' );
}
```

mauteri:
@lhero-org thanks for putting in this fix! When you have a moment, please link your WordPress profile to GitHub. We have a new workflow that automated give credit within the plugin and we'd like you to get that. Thanks!

Read how to do this here: https://make.wordpress.org/core/2020/03/19/associating-github-accounts-with-wordpress-org-profiles/

lhero-org:
Thanks for the review, @mauteri! Will get my WordPress.org profile linked to this account.

mauteri:
@lhero-org if you need help with the tests passing and test coverage let me know.

mauteri:
Thanks for finding, reporting, and fixing this @lhero-org!

Sharp diagnosis. Core registers wp-date on every request, so the registered check could never bail on the front end, and that quietly put moment.js on every page of every GatherPress site. Nice catch out of a real audit.

I pushed one commit with your review feedback applied (PHP_INT_MAX for the hook priority, flattened conditional), plus updates to the two tests that were still asserting the old behavior and coverage for the new branches. The front-end one is a proper regression test: it fails against the old registered check. Also added the changelog entry the gate wanted.

That commit shows under my name since that's how GitHub attributes maintainer pushes to a fork. Your authorship stands and props are already picking you up as lheroorg.

CI is green now, so this is ready to merge.

PR #1945, opened 2026-07-14
Fix event REST API URL to use rest_url() instead of a hardcoded wp-json path

bor0:
## Summary
- Without pretty permalinks, WordPress serves the REST API via `?rest_route=` instead of `/wp-json/`. `Assets::add_interactivity_state()` built `eventApiUrl` with `home_url( 'wp-json/' . $slug )`, a hardcoded path that only resolves with pretty permalinks, so requests like `/wp-json/gatherpress/v1/event/nonce` 404 without them.
- Switched to `rest_url( $slug )`, WP core's function for building REST URLs, which adapts to either permalink structure automatically.
- This state value is also consumed for the RSVP, RSVP form, and RSVP status endpoints (all built via string concatenation on `eventApiUrl` in the interactivity view scripts), so the fix covers those too.
- Updated the existing PHPUnit coverage to assert against `rest_url()` output rather than a hardcoded `wp-json/...` string, and added a new test that disables pretty permalinks and confirms the URL still resolves via the `?rest_route=` form.

| Before | After |
| ------ | ----- |
| Screenshot 2026-07-14 at 17 25 53 | Screenshot 2026-07-14 at 17 27 37 |

## Test plan
- [ ] `npm run test:unit:php -- --filter Test_Assets`
- [ ] Manually confirm on a site with plain permalinks (`Settings > Permalinks > Plain`) that the RSVP block's nonce fetch (`/wp-json/gatherpress/v1/event/nonce` equivalent) no longer 404s and RSVP submission works.
- [ ] Confirm no regression on a site with pretty permalinks enabled.

### DIFF
--- a/.github/changelog/1938-fix-nonce-plain-permalinks
+++ b/.github/changelog/1938-fix-nonce-plain-permalinks
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed the RSVP and nonce REST endpoints returning a 404 when pretty permalinks are disabled, by building the event REST API URL with rest_url() instead of a hardcoded /wp-json/ path.
--- a/includes/core/classes/class-assets.php
+++ b/includes/core/classes/class-assets.php
@@ -176,7 +176,7 @@ class Assets {
 		wp_interactivity_state(
 			'gatherpress',
 			array(
-				'eventApiUrl' => home_url( 'wp-json/' . $event_rest_api_slug ),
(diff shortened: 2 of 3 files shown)


bor0:
cc @mauteri for review 🙇‍♂️

mauteri (approved):

mauteri:
@bor0 looks good and thanks for contributing to the project!

PR #1947, opened 2026-07-15
Fix venue Save button not disabling during creation, allowing duplicate venues

bor0:
## Summary
- The venue creation "Save" button's `disabled` state was wired to core-data's `isSavingEntityRecord` selector, but venue creation POSTs via a raw `apiFetch()` call that never touches `saveEntityRecord`, so the selector never reflected the in-flight request.
- Rapid/repeated clicks on Save could fire multiple duplicate venue-creation requests before the UI ever disabled the button.
- Adds a `useRef`-based re-entrancy guard in `CreateVenueForm` (state alone is not sufficient here since React batches state updates — clicks fired synchronously in the same tick all read the same stale `isCreating` closure) plus a `useState` flag wired into the existing `isSaving` prop, so the button now disables immediately and only one create request is ever sent per save action.

Before

https://github.com/user-attachments/assets/2ad0cdf6-3853-4394-84bc-91ccccb2842f

After

https://github.com/user-attachments/assets/1e723845-a36c-4acd-8f46-9c5602843e68

## Test plan
- [x] Added `test/unit/js/src/components/VenueForm.test.js` covering: (1) rapid repeated clicks only trigger one `apiFetch` call and the button disables, (2) the button re-enables and a subsequent click works normally after the request settles.
- [x] `npm run test:unit:js` — full suite (903 tests) passes.
- [x] `npm run lint:js` — clean on changed files.
- [x] Manually verified in the block editor: clicking Save repeatedly on the venue creation form no longer creates duplicate venue posts.

### DIFF
--- a/.github/changelog/1947-venue-save-double-click
+++ b/.github/changelog/1947-venue-save-double-click
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed the venue creation Save button not disabling while the request is in flight, which allowed repeated clicks to create multiple duplicate venue posts.
--- a/src/components/VenueForm.js
+++ b/src/components/VenueForm.js
@@ -10,7 +10,7 @@ import {
 	__experimentalHStack as HStack,
 	useNavigator,
 } from '@wordpress/components';
-import { useState } from '@wordpress/element';
+import { useRef, useState } from '@wordpress/element';
 import { useSelect, useDispatch } from '@wordpress/data';
 import apiFetch from '@wordpress/api-fetch';
 import { store as coreDataStore } from '@wordpress/core-data';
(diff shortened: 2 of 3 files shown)


bor0:
cc @mauteri for another review :) thanks!

carstingaxion:
Thanks for your contribution @bor0 !

This kind of fixes is great, I really appreciate enhancing the existing UX in ways like that. And the case makes total sense.

Looking into the changed files, everything looked good to me. Unfortunately I wasn't able to test the Playground on my current mobile device, but will do later today.

carstingaxion:
@mauteri @bor0 Tested in Playground a multiple times and can confirm it works as described. And as it should!

Thanks @bor0 !

carstingaxion (approved):

Issue #1961, opened 2026-07-18
Adopt an explicit class finality policy and modernize for the PHP 8.1 minimum


mauteri:
Follow-up from #1509.

## 1. Make class finality an explicit, codebase-wide policy

The generic RSVP subsystem introduced in #1509 declares most of its classes `final` (`Repository`, the `Response` value objects, the `User`/`Email` providers, `Provider_Registry`, `Serializer`). No other class in the codebase uses `final`, so we currently have two implicit conventions side by side.

Decide the policy and apply it consistently in one pass:

- **Option A — final by default**: classes not designed as extension points are `final`; extensibility flows through hooks, interfaces, and abstract bases (like `Rsvp\Response\Provider\Base` and `Venue\Map\Provider\Base`). Existing classes that nothing extends get the keyword.
- **Option B — no final**: drop the keyword from the #1509 classes and keep the codebase's status quo.

Whichever wins, record it in `AGENTS.md` so future PRs don't reintroduce the split.

## 2. PHP 8.1 modernization sweep

The minimum PHP moved from 7.4 to 8.1, but most of the codebase still reads as 7.4-era. Candidates for a hardening/modernization pass, best done as focused per-subsystem PRs rather than one big diff:

- [ ] **Native backed enums** for constant-bag value sets. #1509 establishes the pattern (`Status`, `Identity_Type`, `Visibility`); candidates elsewhere include RSVP status strings used outside the new subsystem, venue-map render modes / map types / scale options, and similar closed sets currently expressed as class constants plus allow-list arrays.
- [ ] **Constructor property promotion + `readonly` properties**. `Rsvp\Repository` already does `protected readonly int $post_id`; the instance classes (`Event`, `Rsvp`, `Venue`, etc.) and small value-style constructors can follow.
- [ ] **`match` expressions** where a `switch` exists only to assign one result (the "switch dispatch with a single trailing return" shape our guidelines currently recommend). `match` is exhaustive by default, which also retires the SonarCloud `default:`-arm boilerplate for those cases.
- [ ] **Native parameter/return/property types** where we only have docblock types today — lets PHPStan verify instead of trust.
- [ ] **Misc 8.x niceties** where they clarify: nullsafe `?->` over isset-chains, `array_is_list()`, `never` for terminating helpers, new-in-initializers for default value objects.

Cautions for the sweep:

[...]

PR #1981, opened 2026-07-20
Add Admin\Notice + Admin\Notifications, warn on upcoming 0.35.0 requirements (#1963)

### LINKED ISSUE #1963: Notify sites below PHP 8.1 / WP 7.0 that 0.35.0 raises the requirements
## What

GatherPress 0.35.0 raises the minimum requirements to **PHP 8.1** and **WordPress 7.0** (develop's readme already declares both). Ship a 0.34.1 patch that warns affected sites ahead of time, so nobody discovers the new floors by failing to update.

Two **separate** admin notices, each shown **only** when its own condition holds (both can appear at once):

1. **PHP notice** — shown when the site runs PHP older than 8.1: GatherPress 0.35.0 will require PHP 8.1; update PHP (or ask your host) to keep receiving updates.
2. **WordPress notice** — shown when the site runs WordPress older than 7.0: GatherPress 0.35.0 will require WordPress 7.0; update WordPress to keep receiving updates.

Sites meeting both floors see nothing.

## Notifications manager class

Rather than two ad-hoc `admin_notices` callbacks, introduce a small class for managing conditional notifications of this kind — this won't be the last lifecycle notice we ship (future requirement bumps, migration prompts, deprecation warnings). Sketch:

- A `Notifications`-style singleton (house pattern) hooked on `admin_notices`, holding registered notices.
- Each notice: an id/slug, a condition callback (evaluated at render time), a message, and a capability gate (something like `update_plugins`/`manage_options` — site admins are who can act on these).
- Registration for these two notices lives with the class; the mechanism stays generic so future notices are one registration each.
- Open to implementation: per-user dismissibility (with the notice returning while the condition still holds seems right for requirement warnings).

## Implementation notes

- **0.34.1 is a patch release**: per the branch model, the work is born on `develop` and cherry-picked to the `version-0.34.1` branch off `main` (`git cherry-pick -x`).
- **The class must be PHP 7.4-compatible.** 0.34.x still supports 7.4 — that's the whole point of the notice — so no 8.1-only syntax (enums, readonly, promotion) in this code path, even though develop's new code is moving the other way (#1961).
- Version floors should not be hardcoded twice — compare against named constants so the 0.35.0 requirements live in one place.

## Acceptance

- On PHP < 8.1: the PHP notice renders in wp-admin for capable users; on PHP ≥ 8.1 it never renders.
- On WP < 7.0: the WordPress notice renders; on WP ≥ 7.0 it never renders.
- Both render together when both conditions hold; neither renders on a compliant site.
- Tests cover the condition gating both ways for each notice and the capability gate.

mauteri:
Fixes #1963. Milestoned 0.34.1; bound for the patch branch by `git cherry-pick -x`.

## What

0.35.0 raises the floors to **PHP 8.1** and **WordPress 7.0**. This ships the warning in 0.34.1 so nobody discovers them by quietly stopping receiving updates.

Two separate notices, each gated on its own condition. A site failing both floors sees both; a compliant site sees nothing.

## The registry

Rather than two ad-hoc `admin_notices` callbacks, this adds a small `Notifications` singleton. Lifecycle notices all have one shape (show while a condition about the site holds, disappear on their own once it stops), so future requirement bumps, migration prompts, and deprecation warnings are one `register()` call each.

It is deliberately **separate from `requirements-check.php`**, which enforces the floors in effect *now* and blocks the plugin from loading. These notices are advisory: the site works fine today, but something is coming.

## Three decisions worth review

**Condition and message are both callables, evaluated at render time.** This matters for the message specifically: building it eagerly would call the translation functions before `init`, which WordPress 6.7 flags as loading a textdomain too early. Registration happens in the constructor, so an eager `esc_html__()` would have tripped that.

**The version comparisons take the version as an argument** (`is_below_upcoming_php( string $php_version )`) rather than reading `PHP_VERSION` internally. That makes both sides of each gate testable without the suite having to run on an old PHP, which also sidesteps the fact that a `final` class can't be subclassed to stub the version source.

**Notices are dismissible**, which in WordPress means dismissed for the current page view only. For a requirement warning that's the behavior you want: it can be waved away while reading a screen, but returns on the next load and keeps returning until the site is actually updated. No user meta, no storage.

The capability gate is `update_plugins`, not `manage_options` — updating PHP or WordPress isn't something an editor can act on, nor a site administrator on multisite.

## PHP 7.4

0.34.x still supports 7.4, and those are exactly the sites that need to see this, so the code path avoids enums, `readonly`, promotion, `match`, and union types even though develop is moving the other way under #1961.

Verified with tooling rather than by inspection. The repo ruleset pins `testVersion` to `8.1-`, so checking this needs an explicit override:

```
vendor/bin/phpcs --standard=PHPCompatibilityWP --runtime-set testVersion 7.4- \
  includes/core/classes/class-notifications.php \
  includes/core/classes/traits/class-singleton.php
```

Both clean, including the `Singleton` trait it depends on.

## Follow-up before 0.35.0

The **two requirement notices** should be unregistered from develop before 0.35.0 ships; a 0.35.0 notice announcing 0.35.0's own floors is nonsense. The **registry itself stays** — that's the reusable part. I'll open a ticket for the removal once this merges.

## Verification

| Gate | Result |
|---|---|
| PHPUnit single-site | 1624 tests, 6844 assertions — OK |
| PHPUnit multisite | 323 tests, 856 assertions — OK |
| PHPCS | clean |
| PHPStan | `[OK] No errors` |
| PHPCompatibilityWP @ 7.4- | clean |

One thing the suite caught that's worth flagging: the singleton persists across tests, so my first version of `test_register` reset `notices` to an empty array and broke a later test through ordering. Every test that mutates the registry now snapshots and restores it.


mauteri:
Reworked per feedback. The flat `Core\Notifications` is gone, replaced by a namespaced pair under `Core\Admin` with a real base class.

## Shape

**`Admin\Notice`** describes a notice as data rather than a bespoke callback: `slug`, `message`, `type` (error / warning / info / success), `dismissible`, `persistent`, `capability`, `condition`.

The two kinds of dismissal are now distinct, which they weren't before:

| | Behavior |
|---|---|
| `dismissible` | Dismissed for the current page view only — all WordPress does natively |
| `persistent` | Records the slug in the new `gatherpress_admin_notifications` option and stays gone |

**`Admin\Notifications`** is the registry: holds `Notice` objects, renders the ones that apply, handles dismissal requests.

**`requirements-check.php`** now builds its two notices through `Notice` instead of hand-rolling `wp_admin_notice()` calls.

## The constraint that shaped it

`requirements-check.php` runs at [gatherpress.php:39](gatherpress.php:39) — **before the autoloader is registered** at line 47, and before we know anything about the site's PHP. So `Notice` is `require_once`'d explicitly, and written in deliberately old syntax: no return types, no scalar parameter types, no typed properties, no null coalescing.

Getting that wrong is worse than not doing it at all. A parse error there replaces the "please upgrade PHP" notice with a fatal, so the file whose whole job is explaining the problem becomes a bigger one — and it takes down the entire site, not just the plugin.

**So it's enforced, not just commented.** `npm run lint:php:early` runs PHPCompatibilityWP at `testVersion 7.2` over the three pre-autoloader files. 7.2 is WordPress 6.7's own PHP floor, so it's the oldest PHP that can run a WordPress capable of running this plugin — derived, not guessed.

It earned its keep immediately: it flagged a `: void` return type in `duplicate-check.php:19` (PHP 7.1+). Not actually reachable, since WP 6.7 already requires 7.2, but that file loads at line 25 — even earlier than the requirements check — and nothing was checking it before.

## Two bugs found while testing

**`check_admin_referer()` calls `wp_die()` on a bad nonce**, so a stale or forged dismiss link would have replaced the admin screen with an error page. Now uses `wp_verify_nonce()` and ignores bad requests quietly. There's a test for the bad-nonce path.

**`wp_admin_notice()` is WordPress 6.4+**, and this class is reachable from the pre-requirements path where the running WordPress may predate it. Calling it there would fatal on an undefined function. `Notice::render()` now falls back to plain markup. That branch is `@codeCoverageIgnore`d per the AGENTS.md rule for `function_exists` guards the test bootstrap always satisfies.

## One decision worth overruling if you disagree

The two requirement notices are **dismissible but not persistent**. A warning that 0.35.0 raises the floors should keep coming back until the site is actually updated, rather than being silenceable forever by the person who most needs to see it. The persistent option exists and is fully tested — it's just not what these two use. Easy to flip if you'd rather they stick.

## Verification

| Gate | Result |
|---|---|
| PHPUnit single-site | 1644 tests, 6892 assertions — OK |
| PHPUnit multisite | 323 tests, 856 assertions — OK |
| PHPCS | clean |
| PHPStan | `[OK] No errors` |
| `lint:php:early` @ 7.2 | clean |
| PR coverage gate | `class-notice.php` 100% (67/67), `class-notifications.php` 100% (77/77), `class-setup.php` 100% (195/195) |

PHPStan also caught that `handle_dismissal` returned `bool` while hooked as an action callback; it's `void` now, with the outcome asserted through the option instead of a return value.

Issue #1982, opened 2026-07-20
Site fatals when GatherPress fails its requirements check while GatherPress Alpha is active


mauteri:
## Describe the bug

When GatherPress fails its own requirements check — most importantly when the site's PHP is below the minimum — it is supposed to degrade gracefully: show an admin notice explaining the problem and not load. Instead, on any site that also has **GatherPress Alpha** active, the whole site fatals with a white screen.

```
Fatal error: Uncaught Error: Class "GatherPress_Alpha\Setup" not found
in /plugins/gatherpress-alpha/gatherpress-alpha.php:76
Stack trace:
#0 /app/wp-includes/class-wp-hook.php(341): {closure}('')
...
#3 /app/wp-settings.php(622): do_action('plugins_loaded')
```

The notice that exists specifically to tell someone their PHP is too old is replaced by the least actionable possible outcome, and it takes down the front end too — not just wp-admin, and not just the plugin.

## Root cause

An ordering problem in `gatherpress.php`, plus Alpha reading a constant as a signal it was never meant to be:

1. `GATHERPRESS_VERSION` is defined at [`gatherpress.php:37`](gatherpress.php#L37) — **two lines before** the requirements gate at line 39.
2. When the gate fails, the file `return`s. The autoloader at lines 47–48 never registers.
3. Alpha registers its namespace with *GatherPress's* autoloader via the `gatherpress_autoloader` filter, so Alpha's classes cannot load without it.
4. Alpha's `plugins_loaded` callback gates on `defined( 'GATHERPRESS_VERSION' )` and a version match. Both pass, because the constant was defined before the bail.
5. Alpha calls `GatherPress_Alpha\Setup::get_instance()`. No autoloader, no class, fatal.

So the constants mean *"GatherPress's main file began executing"*, and Alpha is treating them as *"GatherPress finished bootstrapping."*

### Why the obvious fix is wrong

Moving the `GATHERPRESS_VERSION` define below the gate would break the duplicate-plugin guard: [`duplicate-check.php`](includes/core/duplicate-check.php) detects a second active copy precisely by testing whether `GATHERPRESS_VERSION` is already defined. Move the define later and two copies stop being detected.

## Fix

GatherPress already has the right signal — the `gatherpress_loaded` action, wrapped in `plugins_loaded` so listeners register regardless of plugin order. If the requirements gate fails, it never fires.

[...]

PR #1984, opened 2026-07-20
Correct the gatherpress_loaded @since to 0.34.1 (#1982)

mauteri:
One line. The `gatherpress_loaded` action is documented `@since 0.35.0`, but it now ships first in **0.34.1** as the fix for #1982 — the site-wide fatal when GatherPress fails its requirements check while GatherPress Alpha is active.

Per the docblock conventions in `AGENTS.md`, `@since` tracks the release a symbol first ships in, so the earlier release wins.

## Context

The action exists on develop but is absent from both `main` and the `0.34.0` tag, so it's being added to the 0.34.1 patch branch in #1983. Alpha then boots from it instead of from `defined( 'GATHERPRESS_VERSION' )` ([gatherpress-alpha#54](https://github.com/GatherPress/gatherpress-alpha/pull/54)), which is what stops the fatal.

Docs-only; the generated hook docs under `docs/developer/hooks/` are left for CI to regenerate on merge, per the AGENTS.md rule.

### DIFF
--- a/.github/changelog/1982-gatherpress-loaded-since
+++ b/.github/changelog/1982-gatherpress-loaded-since
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Correct the gatherpress_loaded action's @since to 0.34.1, the release it now first ships in.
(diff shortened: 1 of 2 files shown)


mauteri:
Closing — not worth a standalone PR. The `version-0.34.1` branch already carries the docblock with `@since 0.34.1`, and version/credits/@since stamping is handled by the release tooling on the version PR rather than hand-edited here.

PR #2041, opened 2026-07-27
fix(setup): wipe GatherPress transients on deactivation

### LINKED ISSUE #680: Remove transients on deactivation

> On deactivation, plugins can run a routine to remove temporary data such as cache and temp files and directories.
>
> https://developer.wordpress.org/plugins/plugin-basics/activation-deactivation-hooks/

And they should. 

In the case of GatherPress it's:

- the transients, that should be deleted on deactivation,

### Designs

_No response_

_No response_

### Code of Conduct

faisalahammad:
Fixes #680

Cleans up plugin-owned transient rows on uninstall so the options table no longer accumulates orphans after the plugin is deleted and reinstalled. Multisite-aware: a network-wide uninstall wipes every subsite. Settings options are preserved, only transients are touched — the full settings + table cleanup will land in the #681 follow-up.

## Proposed changes:
* New `uninstall.php` at the plugin root runs the wipe when the user selects "Delete" on the plugin row in wp-admin. It declares a private `gatherpress_uninstall_wipe_transients()` function and calls it once on single-site installs or once per subsite via `get_sites()` + `switch_to_blog()` on multisite.
* `Setup::deactivate_gatherpress_plugin()` is restored to its pre-PR shape — the wipe moved off the deactivation hook so deactivation stays a low-cost affair (settings preserved, no DB churn) and a hook is reserved for the follow-up uninstall work in #681.
* New test file `tests/core/classes/class-test-uninstall.php` covers single-site wipe, settings + unrelated-transient preservation, and a `@group multisite` per-subsite flow.
* Changelog entry at `.github/changelog/680-remove-transients-on-deactivation` (Significance: minor, Type: changed).

### Other information:

## Testing instructions:
Manual smoke on a fresh WP 6.5+ install with PHP 8.1+:

1. Install and activate `gatherpress.zip` from this branch.
2. Trigger something that writes a transient: open the venue block in the editor and type some search characters (writes `gatherpress_photon_search_*`), save a venue with an address (`gatherpress_photon_geocode_*`), or load any page that lists RSVPs (`gatherpress_all_comment_types`).
3. Confirm rows exist: `wp transient list --search='gatherpress*'`.
4. Confirm settings survive the test before uninstall: `wp option get gatherpress_version`.
5. Uninstall the plugin (Delete, not just Deactivate) from the plugins screen, or run `wp plugin uninstall gatherpress`.
6. Re-run `wp transient list --search='gatherpress*'` — expect empty (the fix).
7. Re-run `wp option get gatherpress_version` — still present (settings preserved; #681 will add opt-in cleanup).
8. Reinstall from this branch: search triggers a new transient write (`gatherpress_photon_search_<hash>`), confirming the plugin still functions after an uninstall cycle.

Multisite variant (only required if you want to validate the network-wide path):
1. Activate network-wide on a multisite install: `wp plugin activate --network gatherpress`.
2. On the primary site and at least one subsite, repeat step 2 from above so each site's `wp_options` has its own `gatherpress_*` transient rows.
3. Network uninstall: `wp plugin uninstall --network gatherpress`.
4. On each site: `wp transient list --search='gatherpress*'` — expect empty.
5. Confirm events table still exists on each subsite: uninstall is a cache wipe, not a table drop.

Also covers an unrelated transient (`_transient_other_plugin`) in the test seeding — confirms it survives (we only wipe `_transient_gatherpress_*` prefix).

### Changelog entry

Changelog file already in the branch at `.github/changelog/680-remove-transients-on-deactivation` (Significance: minor, Type: changed).

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

-   [x] Changed - for changes in existing functionality

#### Message

Flush GatherPress cache transients (geocode cache, RSVP caches, rate-limit buckets) on plugin uninstall. Multisite-aware: network uninstall wipes each subsite's rows. Plugin settings are preserved.

</details>


mauteri:
Thanks for contributing @faisalahammad! First, let's make the change `0.36.0` as that's where this will eventually land. Second, I'm thinking we can add this to a new `uninstall.php` file, which will be extended with this [ticket](https://github.com/GatherPress/gatherpress/issues/681).

faisalahammad:
Thanks @mauteri! Both shipped in 25146c1:

- New `uninstall.php` at the plugin root declares a private `gatherpress_uninstall_wipe_transients()` and handles multisite via `get_sites()` + `switch_to_blog()`. Settings, custom tables, and other data stay preserved until #681 lands.
- `Setup::deactivate_gatherpress_plugin()` is reverted to its pre-PR shape (no params, `flush_rewrite_rules()` only) and the private `delete_gatherpress_transients` helper is gone.
- New `tests/core/classes/class-test-uninstall.php` covers single-site wipe, unrelated-transient survival, settings survival, and a `@group multisite` per-subsite flow.
- All new symbols bumped to `@since 0.36.0`.
- `package.json` `files` array includes `uninstall.php` so the `wp-scripts plugin-zip` output ships it.

uninstall.php
```diff
@@ -0,0 +1,122 @@
...
+	);
+
+	// Pattern A: data rows (`_transient_gatherpress_*`).
+	// Pattern B: timeout rows (`_transient_timeout_gatherpress_*`).
+	// Patterns do not overlap — after the `_transient_` prefix the next
+	// char differs (`g` vs `t`).
+	// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching -- Bulk delete on uninstall; not a read path.
+	$wpdb->query(
+		$wpdb->prepare(
+			"DELETE FROM {$wpdb->options} WHERE option_name LIKE %s OR option_name LIKE %s",
```
mauteri: Let's use `%i` instead of `{$wpdb->options}`

uninstall.php
```diff
@@ -0,0 +1,122 @@
...
+	global $wpdb;
+
+	// Collect the plugin-owned transient keys before deletion so we can
+	// invalidate the persistent object cache. The cache stores values
+	// under the unsuffixed key inside the `transient` cache group, so
+	// we need the key names, not just the option names.
+	// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching -- Cache invalidation pre-delete; not a read path.
+	$rows = $wpdb->get_results(
+		$wpdb->prepare(
+			"SELECT option_name FROM {$wpdb->options}"
```
mauteri: Let's use `%i` instead of `{$wpdb->options}`

mauteri:
This is looking good @faisalahammad, and thanks for picking up a second one unprompted.

**First, the easy out:** it's completely fine to keep this PR scoped to the transient wipe exactly as it stands and let me handle what follows separately. You've already done what #680 asked for. Just say the word and we'll land it.

What follows is an offer, not a requirement, because I think there's something more interesting here if you want it.

## The bigger version

#681 is going to need to remove a lot more than transients: settings options, the `gatherpress_events` custom table, event and venue posts with their meta, three taxonomies' terms, RSVP comments and comment meta, and scheduled cron jobs. Each of those will have its own opt-in setting, since some are destructive enough that we don't want to surprise anyone.

If all of that lands as procedural functions in `uninstall.php`, that file becomes unmaintainable fast. So I'd like to follow the same pattern we use for admin notices in `includes/core/classes/admin/notices/`: an abstract base, one small class per concern, and a registry that runs them.

Your transient wipe would be the first one, and it would establish the pattern for everything after it.

### Shape

New namespace `GatherPress\Core\Uninstall`, at `includes/core/classes/uninstall/`.

**1. `uninstall.php` bootstraps the autoloader.** Your current docblock is right that classes aren't available by default, since WordPress includes this file without loading the plugin. Opting back in is two lines:

```php
require_once __DIR__ . '/includes/core/classes/class-autoloader.php';
GatherPress\Core\Autoloader::register();
```

Note `__DIR__` rather than `GATHERPRESS_CORE_PATH`, since that constant is defined in `gatherpress.php` and won't exist here. Worth checking whether `includes/core/register-class-aliases.php` needs requiring first, since `gatherpress.php` loads it immediately before the autoloader.

**2. An abstract `Base`,** loosely mirroring `Admin\Notices\Base`:

- `applies(): bool` — should this run? For now just return `true`. #681 will add the settings check here.
- `uninstall_site(): void` — per-site cleanup.
- `uninstall_network(): void` — network-level cleanup, with a no-op default.

**3. The base owns multisite, so no subclass ever calls `is_multisite()`.** That's the main thing this buys us. The base calls `uninstall_site()` once on single-site, or loops `get_sites()` with `switch_to_blog()` / `restore_current_blog()` on multisite, then calls `uninstall_network()` once. Your existing loop is exactly this logic, it just moves up a level where every future task inherits it.

**4. A small registry** in the same namespace holding the list of tasks, which `uninstall.php` calls. `Admin\Notices\Setup` is the model.

**5. Your transient wipe becomes the first subclass,** implementing `uninstall_site()`.

### Worth keeping as-is

Your reasoning for going at `$wpdb` directly rather than `delete_transient()`, because some object-cache backends no-op the API against the options row, is correct. Please carry that comment across unchanged.

### One thing that should get easier

Your test currently works around `require_once` short-circuiting by calling `gatherpress_uninstall_wipe_transients()` directly, with a comment explaining why. Once it's a class, tests can just instantiate and call the method, so that workaround goes away.

### Changelog

If you take this on, the entry is worth rewording. It currently reads as a transient cleanup, but establishing the uninstall pattern is the more notable half.

---

Again, no obligation. If you'd rather ship the focused version, that's a completely good outcome and I'll build the base on top of it.

mauteri (changes requested):

uninstall.php
faisalahammad: Replaced with `%i` placeholder and passed `$wpdb->options` as argument. uninstall.php L53-L55.

uninstall.php
faisalahammad: Replaced with `%i` placeholder and passed `$wpdb->options` as argument. uninstall.php L68-L70.

faisalahammad:
Addressed feedback: replaced `$wpdb->options` interpolation with `%i` identifier placeholder in both SELECT and DELETE queries inside gatherpress_uninstall_wipe_transients() (uninstall.php L53-L55, L68-L70). Matches established pattern from class-query.php and class-admin-list.php.

Ready for re-review.

mauteri:
> Addressed feedback: replaced `$wpdb->options` interpolation with `%i` identifier placeholder in both SELECT and DELETE queries inside gatherpress_uninstall_wipe_transients() (uninstall.php L53-L55, L68-L70). Matches established pattern from class-query.php and class-admin-list.php.
> 
> Ready for re-review.

Thanks @faisalahammad! Did you see my other comment above about the larger setup I'd like to see here? If you're not interested in that that's okay and we can merge this into 0.36.0 as is and then rework the code with #681, but if you'd like to lay down the bones for how we uninstall all things that'd be cool. Let me know. Thx!

faisalahammad:
Re-requesting review from @mauteri.

Answer to your question on the bigger Uninstall Base/registry pattern: shipping the focused transient-wipe version on this PR as you offered. Designing the abstract pattern against a single wipe target would add scaffolding for one of N future subclasses (settings, custom tables, posts/taxs, RSVPs, cron); it lands cleaner as part of #681 when the full target set is visible. Nothing more to action here until #681.

Other items: `%i` placeholder shipped in `a4832097`. CodeRabbit on the 4 changed files reports 0 findings. `npm run lint:php` clean on the PR branch. Plugin-zip bundles `uninstall.php` correctly.

Coverage gate (`pr-coverage.yml` shows PHP check did not run): the wp-env `coverage.xml` artifact at `/var/www/html/` is not flushing before the post-test `docker cp` step. Same shape seen on prior PRs. Infrastructure, not code. Suggest a workflow re-run once wp-env stabilizes.

mauteri (approved):

PR #2042, opened 2026-07-27
fix(email): make event-update subject editable per send

### LINKED ISSUE #827: Make email subject editable per email

Based on feedback (https://github.com/GatherPress/gatherpress/pull/752#pullrequestreview-2260237488), I’d like to suggest making the email subject customisable for each email.

### Designs

_No response_

_No response_

### Code of Conduct

faisalahammad:
Fixes #827

## Proposed changes:

* Add an editable Subject field to the "Send event update via email" modal, pre-populated with the existing `📅 {event title}` default so behavior is unchanged when organizers leave it alone.
* Thread an optional `subject` argument through the email send chain: new `subject` arg on the `POST /email` REST route, passed as a fourth cron argument via `wp_schedule_single_event()`, then through `handle_email_send_action()` → `send_emails()` → `send_event_email_to_recipient()` → `wp_mail()`. All new parameters are optional with a `''` default, so events scheduled before this change still execute, and the cron action accepted-args count is bumped from 3 to 4.
* Add a `gatherpress_email_subject` filter so developers can customize the subject template. It applies to both the default and a supplied subject.
* The RSVP confirmation email is untouched since it is not composed by a human at send time.

### Other information:

## Testing instructions:

* Edit an upcoming event, save/publish it, and open the "Compose Message" modal from the success notice.
* The modal now shows a Subject input above the message textarea, pre-populated with `📅 {event title}`.
* Change the subject (e.g. `Reminder: doors at 7`), check a recipient box, confirm the prompt, and send. The received email carries the custom subject.
* Clear the Subject field and send again. The email falls back to the default `📅 {event title}` subject.
* Developer check: add `add_filter( 'gatherpress_email_subject', function ( $subject ) { return '[My Community] ' . $subject; } );` and the prefix appears on both default and custom subjects.
* PHPUnit: four new tests capture the outgoing subject via the `pre_wp_mail` filter (custom subject passthrough, default fallback, filter on default, filter on custom subject); existing email tests updated for the new parameter.

My WordPress.org username:

### Changelog entry

A changelog entry is already committed in this branch at `.github/changelog/827-editable-email-subject`, so the automated creation below is left unchecked.

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Event organizers can now set a custom subject line when sending the event-update email; the existing `📅 {title}` default stays in place when the field is left untouched, and a new `gatherpress_email_subject` filter lets developers customize the default template.

</details>

## Screen recording

https://github.com/user-attachments/assets/54fcc140-15b2-4a0c-8c6e-a7d005664fd4

### DIFF
--- a/src/modals/email-communication/index.js
+++ b/src/modals/email-communication/index.js
@@ -132,6 +132,7 @@ const EventCommunicationModal = () => {
 						label={ __( 'Subject', 'gatherpress' ) }
 						value={ subject }
 						onChange={ ( value ) => setSubject( value ) }
+						style={ { marginBottom: '16px' } }
 					/>
 					<TextareaControl
 						label={ __( 'Optional message', 'gatherpress' ) }
--- a/src/modals/email-communication/index.js
+++ b/src/modals/email-communication/index.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { __, _x } from '@wordpress/i18n';
+import { __, _x, sprintf } from '@wordpress/i18n';
 import domReady from '@wordpress/dom-ready';
 import { createRoot, useState, useEffect, useRef } from '@wordpress/element';
 import {
@@ -10,6 +10,7 @@ import {
 	Flex,
 	FlexItem,
 	Modal,
+	TextControl,
 	TextareaControl,
 } from '@wordpress/components';
 import apiFetch from '@wordpress/api-fetch';
@@ -42,6 +43,7 @@ const EventCommunicationModal = () => {
 	const [ isWaitingListChecked, setWaitingListChecked ] = useState( false );
 	const [ isNotAttendingChecked, setNotAttendingChecked ] = useState( false );
 	const [ buttonDisabled, setButtonDisabled ] = useState( false );
+	const [ subject, setSubject ] = useState( '' );
 	const [ message, setMessage ] = useState( '' );
 	const textareaRef = useRef( null );
 	const sendMessage = () => {
@@ -54,6 +56,7 @@ const EventCommunicationModal = () => {
 				method: 'POST',
 				data: {
 					post_id: select( 'core/editor' ).getCurrentPostId(),
+					subject,
 					message,
 					send: {
 						all: isAllChecked,
@@ -65,6 +68,7 @@ const EventCommunicationModal = () => {
 			} ).then( ( res ) => {
 				if ( res.success ) {
 					closeModal();
+					setSubject( '' );
 					setMessage( '' );
 					setAllChecked( false );
 					setAttendingChecked( false );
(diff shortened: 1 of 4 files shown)


mauteri:
Thanks for this one too. Let's change this to `0.36.0` where this will eventually land. Thx!

mauteri:
@faisalahammad one thing I noticed in your screencast is that the subject input needs some bottom margin so it's not right up against the Optional Message label. Thx!

faisalahammad:
Thanks @mauteri! Both points fixed in 5c30659:

- All four `@since 0.35.0` docblocks in `class-rest-api.php` bumped to `0.36.0` (handle_email_send_action, send_emails, send_event_email_to_recipient, and the new apply_filters comment).
- Subject TextControl now has 16px bottom margin so it sits apart from the Optional message label.

src/modals/email-communication/index.js
```diff
@@ -109,6 +128,12 @@ const EventCommunicationModal = () => {
 					shouldCloseOnClickOutside={ false }
 					style={ { maxWidth: '550px' } }
 				>
+					<TextControl
+						label={ __( 'Subject', 'gatherpress' ) }
+						value={ subject }
+						onChange={ ( value ) => setSubject( value ) }
+						style={ { marginBottom: '16px' } }
```
mauteri: Sorry, one small change. Let's make this 1rem instead of 16px (code uses mostly rems in these cases).

mauteri (changes requested):

src/modals/email-communication/index.js
faisalahammad: Fixed in next push — marginBottom switched to '1rem' on src/modals/email-communication/index.js:135.

wppoland:
@faisalahammad nice feature, this is one I would use. Two things I hit while testing it, both fixed in a PR into your branch so it stays your change: faisalahammad/gatherpress#1.

**The pre-filled subject costs per-recipient locale.** `send_event_email_to_recipient()` builds the default subject after `switch_to_user_locale( $recipient['user_id'] )`, so each member currently gets `📅 {title}` in their own language. Because the modal pre-fills the field and posts it every time, the `'' === $subject` branch stops being reachable and everyone receives the sender's rendering. Making the default a `placeholder` rather than a value keeps both behaviours: untouched means the server composes it per recipient, typed means your custom subject.

**`'validate_callback' => 'sanitize_text_field'` on the new `subject` arg does neither.** A validate callback's return value is only read as pass/fail and a string is never `false`, so the param arrives unchecked and unsanitized. That already bites on the existing `message` arg on `develop`: the request answers success, the cron event is scheduled, and the cron callback then fatals on the typed parameter, so nothing is sent and nobody is told.

```
REST status: 200 data: {"success":true}
cron scheduled with the array message: true
CRON CALLBACK FATAL: TypeError: Rest_Api::handle_email_send_action(): Argument #3 ($message) must be of type string, array given
```

With `'type' => 'string'` plus `rest_validate_request_arg` and `sanitize_text_field`, the same request answers `400 subject is not of type string` and a normal subject still passes. I fixed `message` in the same array since it is the identical one-liner.

Two tests come along: both params declaring the schema trio, and an unsupplied subject being scheduled as an empty string so the server-side default stays reachable.

mauteri (approved):

PR #2048, opened 2026-07-28
Defer venue-save-triggered static-map generation to WP-Cron (opt-in)

### LINKED ISSUE #1823: Groups: venue create/update blocks synchronously on static-map tile fetch + composite
ℹ️ Fix in [GatherPress core](https://github.com/GatherPress/gatherpress/)

Found during @raicem led testing in https://docs.google.com/document/d/1qH37rfsgAF0f5bA16PTezfslvS57_4OKc-u6kFIxHRk/edit?tab=t.0, confirmed as real issue.

## Reproduction

Noted while REST-testing the venue endpoints (create/update on `/wp/v2/gatherpress_venues/{id}`): updating a venue's address is slow, with the request visibly taking several seconds to return.

## Root cause

GatherPress's `Map::maybe_generate()` (`plugins/gatherpress/includes/core/classes/venue/map/class-map.php:301`, hooked on `wp_after_insert_post`) runs synchronously on every venue save — not async. It calls `Provider::render()` (`.../provider/class-osm.php:109`), which fetches OSM/CartoDB map tiles over HTTP and composites a PNG with GD, once for the 1x image and again for the 2x retina image. Each render is bounded by a 10-second wall-clock budget (`Osm::COMPOSITE_TIME_BUDGET`), so a single venue save/update can block the request for up to ~20 seconds in the worst case (two renders back-to-back).

## Impact

Every venue create/update through our custom REST flow (`wporg-groups-frontend/inc/rest.php::resolve_venue_id()` / `assign_venue_to_event()`) pays this cost inline, which is especially painful during bulk data migration (creating many venues) and makes the front-end event form feel unresponsive whenever a new venue address is entered.

## Fix direction

This is vendor (GatherPress core) behavior, so the fix lives in our own code rather than patching the plugin:
- Lower `gatherpress_static_map_composite_time_budget` and/or disable the retina render via GatherPress's `gatherpress_static_map_generate_retina` filter in `gatherpress-groups-tweaks.php`, at least for bulk/migration contexts.
- Consider deferring venue-save-triggered map generation to `wp_schedule_single_event()` (fire-and-forget) instead of relying on the synchronous `wp_after_insert_post` hook, if GatherPress doesn't offer an async path itself.
- File the synchronous-generation behavior upstream with GatherPress if it isn't already a known issue.

## How to verify a fix

Time a venue create/update (REST or front-end form) with a real, geocodable address before and after, and confirm the request no longer scales with tile-fetch latency.

bor0:
Fixes https://github.com/WordPress/wordcamp.org/issues/1823

## Summary

- `Map::maybe_generate()` runs synchronously on every venue save via `wp_after_insert_post`, fetching OSM tiles and compositing PNGs inline — up to ~20s worst case (1× + 2× retina) when the tile host is slow. This blocks venue create/update requests, which is especially painful for REST integrations or importers that save many venues in a tight loop.
- Adds an opt-in `gatherpress_static_map_generate_async` filter (default `false`, fully backward compatible) that defers the actual render to a WP-Cron job instead of blocking the save, mirroring the existing async-geocode pattern (`gatherpress_geocode_on_save_enabled` / `gatherpress_async_geocode_pre_enqueue_job`) already used elsewhere in the venue subsystem. A matching `gatherpress_static_map_generate_pre_enqueue_job` short-circuit filter lets a companion plugin route through Action Scheduler instead of WP-Cron.

Fixes WordPress/wordcamp.org#1823

## Test plan

- [x] `npm run lint:php` — clean
- [x] `npm run lint:phpstan` — clean
- [x] `npm run test:unit:php` / `npm run test:unit:php:multisite` — full suite passes; added 7 new tests covering the async scheduling path, dedup, the pre-enqueue short-circuit filter, and cron-handler edge cases (missing coordinates, unsupported post type)

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in option to defer venue static-map generation to WP-Cron.
  - Static maps can now be generated asynchronously during bulk imports and REST saves.
  - Duplicate scheduled jobs are prevented, and queued jobs revalidate venue details before processing.
  - Invalid or deleted venues are skipped safely; existing maps for trashed venues remain unchanged.

- **Documentation**
  - Added a changelog entry describing asynchronous static-map generation.


bor0:
cc @mauteri for visibility and review :)

mauteri:
@bor0 I fixed the `Test Coverage Report` which wasn't showing what was failing in a comment (see above). Make sure to pull in changes I made. Thx!

includes/core/classes/venue/map/class-map.php
```diff
@@ -272,6 +272,17 @@ final class Map {
 	 */
 	const UPLOADS_SUBDIR = 'gatherpress/static-maps';
 
+	/**
+	 * Cron action fired to generate a single (venue, combo) static map when
+	 * {@see self::maybe_generate()} defers to WP-Cron instead of rendering
+	 * inline. Handler signature: `( int $post_id, int $zoom, int $width,
+	 * int $height, string $map_type )`.
+	 *
+	 * @since 0.35.0
```
mauteri: We'll need to update all to `@since 0.36.0`

wppoland:
Read this at `e11a8ec1`. Not much to push back on — the shape matches `Prewarm` deliberately, which is the right call: same `pre_*` short-circuit convention, same `wp_next_scheduled()` dedup, same handler-args contract, so there is one pattern to learn rather than two.

Two things I specifically checked because they are where deferred work usually goes wrong, and both are handled:

- **Stale args.** `process_generate_job()` re-reads `get_information()` at run time rather than carrying coordinates through the job payload, so a venue re-geocoded between the save and the tick renders from current data. The re-validation of post status, post type support and coordinate parseability covers the trashed/unsupported/un-geocodable cases the docblock claims.
- **Default unchanged.** `should_generate_async()` defaults false, so every existing site keeps the synchronous behaviour and nobody gets a surprise "map appears a few seconds later" from an upgrade.

### `@since 0.35.0` needs to become `0.36.0`

Correct when you opened this on 28 July, stale now that 0.35.0 shipped on 10 August. Affects `GENERATE_CRON_ACTION`, `should_generate_async()`, `schedule_combo_generation()`, `process_generate_job()` and both filter docblocks.

@mauteri asked for the same bump on my #2067 ([comment](https://github.com/GatherPress/gatherpress/pull/2067#issuecomment-3159657468)), so it will be raised anyway. Same note is on #2112, which has it too — mentioning that only so it is clear this is a release-timing sweep and not a comment on your patch.

### Pre-existing, not caused by this PR: nothing clears the map crons

`deactivate_gatherpress_plugin()` only calls `flush_rewrite_rules()`. Searching the repo, the only production caller of `wp_clear_scheduled_hook()` is `rsvp/class-cleanup.php` — the remaining hits are tests. So `Prewarm::CRON_ACTION` and `Prewarm::FULL_SWEEP_ACTION` already survive deactivation, and this adds a third action with the same property.

Being honest about severity: it is low. Orphaned events fire once against no listener and drop out, and your run-time re-validation means even a deactivate/reactivate cycle cannot act on stale data. So I would not hold this PR for it.

Worth its own ticket though, and if you would rather absorb it here it is two lines in the deactivation handler alongside the prewarm ones. Happy to file the ticket separately so it does not widen your diff — say which you prefer.

Props motylanogha

wppoland (commented):
The async design reads well: it mirrors the existing geocode pattern faithfully, the `pre_enqueue_job` short-circuit follows the core `pre_*` convention including the null-vs-falsy distinction, and `process_generate_job()` re-validating post status, post type support and coordinates at run time is exactly right for a job that can fire after the venue has moved on.

One thing to fix before this lands: the diff rewrites three existing `@since 0.35.0` docblocks to `@since 0.36.0`.

- `ensure_descriptor_for_combo()`
- `parse_coord()`, including its `@since 0.35.0 Visibility changed to public so the REST layer applies the same coercion` line
- `normalize_map_type()`

All three shipped in 0.35.0:

```
$ git show 0.35.0:includes/core/classes/venue/map/class-map.php | grep -nE "function (ensure_descriptor_for_combo|parse_coord|normalize_map_type)"
1107:	protected function ensure_descriptor_for_combo(
1474:	public function parse_coord( $raw ): ?float {
1553:	public function normalize_map_type( string $map_type ): string {
```

`parse_coord()` is already public there, so the visibility note is dated 0.35.0 correctly too. Moving these to 0.36.0 tells an extender that public API is new in the upcoming release when it has been available since 0.35.0, and AGENTS.md is explicit that a later change does not move `@since`:

> **Signature changes after introduction don't move `@since`.** If a filter shipped in 0.30.0 and grew a third `@param` in 0.31.0, the docblock stays `@since 0.30.0`.

The `@since 0.36.0` tags on the genuinely new symbols (`GENERATE_CRON_ACTION`, `should_generate_async()`, `schedule_combo_generation()`, `process_generate_job()`, and both new filters) are all correct as written.

Separately, the red CI here is not yours. Every PHPUnit job in this run died at `wp-env start` with `fatal: couldn't find remote ref 7.0.4` — WordPress.org announced 7.0.4 before the git mirror published the tag, so env setup failed for every PR in that window. The tag exists now (`16a7ab2f48969f10b1390b36866abed0d40f0f73`), so a re-run or a rebase onto current develop clears it.

carstingaxion (commented):
This looks all very good to me! 👍 

And the one oncovered line is propably an easy fix for @bor0 

> ❌ includes/core/classes/venue/map/class-map.php
   Uncovered line: 679

mauteri (approved):

PR #2049, opened 2026-07-28
feat: add Event::is_online()/set_online() and pre-resolved online-event term ID for editor JS

### LINKED ISSUE #2047: Add a public API for online events (setter, term helpers, REST parity)

**Proposed:**

| File | Line |
| --- | --- |

**Proposed:**

```php
	'show_in_rest'      => true,
	'single'            => true,
	'default'           => '',
),
```

### Designs

### Credits

faisalahammad:
Fixes #2047

## Proposed changes:

* Add `Event::ONLINE_EVENT_TERM_SLUG` constant and `Event::is_online()` / `Event::set_online()` methods — the public API for owning the online-event sentinel term and `gatherpress_online_event_link` meta as a single coordinated write.
* Add `Venue\Setup::is_online_event_term_slug()` and `Venue\Setup::get_online_event_term_id()` — direct helpers so callers never hardcode the `'online-event'` string or run `get_term_by()` on their own.
* Add `Venue\Setup::taxonomy_for_event_post_type()` — resolves the venue taxonomy for any event-related post type via `post_type_supports`, replacing inline `$this->get_taxonomy( ... )` patterns that assumed a single venue post type.
* Expose pre-resolved sentinel term IDs in `gatherpress.config.onlineEventTermIds` through the editor settings filter, keyed by venue post type.
* Migrate all JS components (OnlineEvent, VenueNavigator, VenueTermsCombobox, venue blocks, venue helpers) from taxonomy REST lookups to `getOnlineEventTermId()` helper that reads the pre-resolved map.
* Comprehensive PHP + JS test coverage: new `Event::is_online()`/`set_online()` tests (9 methods), `Venue\Setup::get_online_event_term_id()` tests, and JS fallback branch coverage for `getOnlineEventTermId()` and `calculateMode()` edge cases.

Deliverable 3 (REST `is_online` write parity) is deferred per issue discussion — settled representation first.

### Other information:

## Testing instructions:

1. Open an existing or new event post in the block editor.
2. Verify the online-event toggle appears and the venue combobox populates correctly.
3. Inspect `wp.data.select('core/editor').getEditorSettings().gatherpress.config.onlineEventTermIds` — should show numeric term IDs, not an empty object.
4. Check the Network panel — taxonomy REST lookups for `slug=online-event` should be absent.
5. Toggle the Online Event switch on with a URL. Save. Confirm the `online-event` term appears in `_gatherpress_venue` and `gatherpress_online_event_link` meta is set.
6. Toggle off. Confirm both are removed. Attach a physical venue first and repeat — hybrid events preserve venue terms.
7. Run `npm run test:unit:js` (997 passing) and `npm run test:unit:php` (1656 passing, 6900 assertions).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Add `Event::is_online()` and `Event::set_online()` for coordinated online-event term and meta writes, plus `Venue\Setup` helpers and pre-resolved editor JS config for the sentinel term ID.

</details>

## Summary by CodeRabbit

* **New Features**
  * Added support for marking events as online and saving an optional online link.
  * Online and hybrid events now preserve existing venue selections.
  * Added reliable online-event detection across event and venue editing tools.
  * Editor settings now provide pre-resolved online-event information for faster, more consistent controls.

* **Bug Fixes**
  * Improved handling when online-event settings or terms are unavailable.
  * Online-event detection now works even when the term’s slug differs from the default.


includes/core/classes/event/class-event.php
```diff
@@ -810,4 +825,105 @@ public function maybe_get_online_event_link(): string {
...
+			return false;
+		}
+
+		foreach ( $terms as $term ) {
+			if ( $term instanceof WP_Term && self::ONLINE_EVENT_TERM_SLUG === $term->slug ) {
+				return true;
+			}
+		}
+
+		return false;
```
mauteri: Cant simplify this all to `has_term`.

includes/core/classes/event/class-event.php
```diff
@@ -810,4 +825,105 @@ public function maybe_get_online_event_link(): string {
...
+	/**
+	 * Whether this event is marked online.
+	 *
+	 * Unconditional online-status check: returns true whenever the event
+	 * carries the `online-event` sentinel term in its venue taxonomy, no
+	 * matter the current user's RSVP status, the event's time, or the admin
+	 * context. Distinct from {@see self::maybe_get_online_event_link()}, which
+	 * gates link disclosure on attendance and time.
+	 *
+	 * @since 0.35.0
```
mauteri: `0.36.0` here and everywhere

wppoland:
I am closing my #2066 and #2068 in favour of this one. Yours predates them by a couple of days and covers the same API plus the JS half, so there is no sense in three open PRs against #2047. Apologies for the duplicated effort, I should have checked the open PRs before starting.

Two observations from my version that might be worth folding in. Both are design calls rather than defects, so take or leave them.

**1. Where `ONLINE_EVENT_TERM_SLUG` lives.**

You define it on `Event`, and `Venue\Setup` reads it back as `Core_Event::ONLINE_EVENT_TERM_SLUG` (`venue/class-setup.php:610`). That gives the venue subsystem a dependency on `Event` for a term that only exists inside the venue taxonomy: the sentinel is a term in `_gatherpress_venue`, and it is seeded and resolved entirely by `Venue\Setup`.

I had put the constant on `Venue\Setup` instead, so ownership matches the taxonomy and the dependency arrow points event to venue only. Not a correctness problem in either direction, but it is cheap to settle now and awkward to reverse once other code imports it.

**2. `set_online()` returning `void` hides a failed write.**

There are three paths where it returns having done nothing:

- `$this->event` is falsy
- `$term_id` is null and `$is_online` is false
- `$term_id` is still null after `add_online_event_term()` ran

The third is the interesting one. The seeding attempt failed, and the caller is told nothing and has no way to find out short of re-querying the terms itself. I had this return `bool` so a caller can distinguish "marked online" from "silently did nothing".

Given the point of #2047 is a public API that other plugins call, that distinction seems worth having.

Happy to send either of these as a PR into your branch rather than leaving you to apply them by hand. Just say which, if any.

faisalahammad:
Thanks for the review, @wppoland. Both suggestions adopted:

**1. ONLINE_EVENT_TERM_SLUG moved to Venue\Setup** — constant is now defined on `Venue\Setup` (as `self::ONLINE_EVENT_TERM_SLUG`), `Event` reads it back via `Venue_Setup::ONLINE_EVENT_TERM_SLUG`. Dependency arrow is event→venue only, matching the taxonomy ownership.

**2. set_online() now returns bool** — signature is `public function set_online( bool $is_online, string $link = '' ): bool`. Returns `true` on successful toggle-on/off, `false` when the event has no bound post, or when the term seeding fails (or no-op toggle-off when term doesn't exist).

Both changes are in commit 0eddc138 on this branch.

No need to open a PR into my branch — these were straightforward to apply. Appreciate the thorough review.

wppoland:
Sorry for the slow follow-up. I checked both changes at `51667ce2` rather than taking the summary on trust, and both are right.

**1. Constant ownership.** `ONLINE_EVENT_TERM_SLUG` is on `Venue\Setup` (class-setup.php:58) and `Event` reads it back. The arrow points event→venue only now, which is what I was after — the sentinel lives in the venue taxonomy, so the venue subsystem should own its name.

**2. `set_online(): bool`.** The paths are all there: `false` when there is no bound post, `false` when the sentinel cannot be seeded (both the "toggle off with no term" case and the "seeding failed" case after `add_online_event_term()`), `true` on a real toggle. That is the part that mattered for a public API — a caller can now tell a failed seeding from a successful no-op, which it could not before.

### One thing my suggestion caused, worth a one-line fix

Moving the constant increased how often `Event` reaches into the venue subsystem, and that surfaced a naming collision I should have anticipated when I proposed it.

`class-event.php` is in `namespace GatherPress\Core\Event`, and a local `Event\Setup` exists (`final class Setup`, event/class-setup.php:37). The file imports the venue one unaliased:

```php
use GatherPress\Core\Setup as Core_Setup;
use GatherPress\Core\Rsvp\Setup as Rsvp_Setup;
use GatherPress\Core\Venue\Setup;          // <- unaliased
```

So the four bare `Setup::` usages in the file (lines 559, 833, 841, 872) resolve to `Venue\Setup`, while a reader sitting in the `Event` namespace would reasonably assume `Event\Setup`. Line 884 makes the hazard visible: `Core_Setup::get_instance()` sits four lines below `Setup::get_instance()`, and they are different classes.

`AGENTS.md` has this as an explicit rule, with this exact pair as the bad example:

> **Cross-subsystem use aliases**: When importing a class from another subsystem that shares a name with a local one (every subsystem has its own `Setup`, for example), alias it.
> ❌ Bad (silently shadows the local `Event\Setup`): `use GatherPress\Core\Venue\Setup;` → `Setup::get_instance()`

Suggested:

```php
use GatherPress\Core\Venue\Setup as Venue_Setup;
```

and the four call sites become `Venue_Setup::`. You already alias the other two Setups in the same file, so this just makes the third consistent. Nothing functional changes.

Happy to send it as a PR into your branch if that is easier than doing it yourself — say the word.

Props motylanogha

faisalahammad:
Alias applied in `d65e39c5`. `use GatherPress\Core\Venue\Setup as Venue_Setup;` now lines up with the `Core_Setup` and `Rsvp_Setup` already aliased in the same file. All four call sites (`get_venue_information()`, `is_online()`, `set_online()`) now read `Venue_Setup::`. Nothing functional changed, this is the AGENTS.md naming rule: the bare `Setup::` no longer shadows the local `Event\Setup`.

Kept `Core_Setup::get_instance()` on line 884 untouched.

I also resolved the merge conflict with `develop` in the test file: the import block had diverged. The test needs both `GatherPress\Core\Event\Setup as Event_Setup` (for `resolve_pending_datetimes()`) and `GatherPress\Core\Setup` (for `add_online_event_term()`), so both are kept; `phpcs` passes on the resolved file and there are no stray conflict markers.

Lint and static checks pass: `phpcs` clean on both edited files, `phpstan` no errors, JS suite 1003/1003. The PHP unit suite could not run locally this session because wp-env failed to fetch its PHP image (`fatal: couldn't find remote ref 7.0.4`) — that is an environment issue, not related to this change; CI runs the PHP suite.

Thanks for catching the collision. Please re-review when you get a chance.

includes/core/classes/event/class-event.php
```diff
@@ -810,4 +811,106 @@ public function maybe_get_online_event_link(): string {
...
+		// Toggle off: drop the sentinel from the term list (preserve venue terms)
+		// and clear the link meta so re-enabling starts blank.
+		$remaining = array_values( array_filter( $existing, static fn ( int $id ): bool => $id !== $term_id ) );
+
+		if ( count( $remaining ) !== count( $existing ) ) {
+			wp_set_post_terms( $this->event->ID, $remaining, $taxonomy );
+		}
+
+		delete_post_meta( $this->event->ID, 'gatherpress_online_event_link' );
+		return true;
```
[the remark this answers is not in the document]
faisalahammad: Fixed in `18d2ce7b`. `set_online()` now guards the `wp_get_post_terms()` read and both `wp_set_post_terms()` writes with `is_wp_error()` — a failing taxonomy operation returns `false` and never reaches the `gatherpress_online_event_link` meta write (no orphaned meta; the toggle-off path only deletes the meta after the term removal succeeds). Preserves the existing early-return semantics. New branch coverage added for all three failure paths.

includes/core/classes/venue/class-setup.php
```diff
@@ -467,11 +492,14 @@ public function get_venue_meta( int $post_id, string $post_type ): array {
...
-			$event       = new Event( $post_id );
+			$event       = new Core_Event( $post_id );
 			$venue_terms = get_the_terms( $post_id, $this->taxonomy_for_event_post_type( $post_type ) );
 			$venue_slug  = ( is_array( $venue_terms ) && ! empty( $venue_terms ) ) ? $venue_terms[0]->slug : null;
 
-			$venue_meta['isOnlineEventTerm'] = ( 'online-event' === $venue_slug );
+			$venue_meta['isOnlineEventTerm'] = (
+			null !== $venue_slug
+			&& $this->is_online_event_term_slug( $venue_slug )
+			);
```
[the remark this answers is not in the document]
faisalahammad: Fixed in `18d2ce7b`. `get_venue_meta()` now derives `isOnlineEventTerm` from `Core_Event::is_online()`, which iterates every assigned venue term instead of reading only `$venue_terms[0]`. A hybrid event with a physical venue first and `online-event` later now correctly reports the online state. `$venue_terms`/`$venue_slug` locals dropped. A hybrid-order coverage test was added in `class-test-setup.php`.

test/unit/php/includes/tests/core/classes/event/class-test-event.php
```diff
@@ -955,6 +957,272 @@ public function test_maybe_get_online_event_link(): void {
...
+		$event = new Event( 0 );
+
+		// Should not error, should not create a term.
+		$event->set_online( true, 'https://example.com/meet' );
+
+		$this->assertFalse(
+			get_term_by( 'slug', Venue\Setup::ONLINE_EVENT_TERM_SLUG, Venue::TAXONOMY ) instanceof WP_Term,
+			'set_online with no bound event should not seed any term.'
+		);
+	}
```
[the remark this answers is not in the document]
faisalahammad: Fixed in `18d2ce7b`. `test_set_online_bails_when_no_event` now captures and asserts the `bool` return: `$this->assertFalse( $event->set_online( true, 'https://example.com/meet' ) );`, preserving the existing no-term-created assertion.

faisalahammad:
All three CodeRabbit findings are fixed and their threads resolved in `18d2ce7b`:

- **`set_online()` taxonomy failures (Major):** `wp_get_post_terms()` read + both `wp_set_post_terms()` writes now guarded with `is_wp_error()` — a failed taxonomy op returns `false` and never reaches the `gatherpress_online_event_link` meta write. New branch coverage for all three failure paths.
- **`get_venue_meta()` sentinel detection (Major):** `isOnlineEventTerm` now derives from `Core_Event::is_online()`, which iterates every assigned venue term instead of only `$venue_terms[0]` — correct for hybrid events with a physical venue first. Hybrid-order coverage test added.
- **`test_set_online_bails_when_no_event` (Minor):** now captures and asserts the `bool` return (`assertFalse`), preserving the no-term-created assertion.

Local verification: PHPCS clean, PHPStan clean, PHP single-site 1701 OK, multisite 327 OK, JS 1003 pass, coverage gate 5/5 files @ 100%, `plugin-zip` builds.

I don't have the permission to request a formal re-review from you as a reviewer (`requested_reviewers` needs collaborator rights), so pinging here — @wppoland, could you re-review when convenient? 🎉

wppoland (commented):
Good shape overall, and the test coverage on the term/meta pairing is thorough. Three things worth addressing.

**1. `set_online( true )` erases a stored link when called without one.**

```php
update_post_meta( $this->event->ID, 'gatherpress_online_event_link', esc_url_raw( $link ) );
```

That write is unconditional, and `$link` defaults to `''`. So an event that already has a link loses it on any call that only means to assert the status:

```php
$event = new Event( $post_id );
$event->set_online( true, 'https://meet.example/abc' );
$event->set_online( true );   // link meta is now ''
```

That is a plausible call for an importer re-syncing rows, or for a caller that toggles status separately from editing the URL. The docblock calls `$link` optional, which reads as "leave it alone", not "clear it". `test_set_online_is_idempotent_on` doesn't catch it because it passes the same link both times. Suggest writing the meta only when `'' !== $link`, and saying so in the docblock, so that clearing a link stays an explicit act (`set_online( false )`, which already deletes it).

**2. `@since 0.35.0` is stale on every new symbol.**

0.35.0 and 0.35.1 have both shipped and none of these exist in either:

```
$ git show 0.35.1:includes/core/classes/event/class-event.php | grep -cE "function (is_online|set_online)"
0
$ git show 0.35.1:includes/core/classes/venue/class-setup.php | grep -cE "function (is_online_event_term_slug|get_online_event_term_id)"
0
```

`is_online()`, `set_online()`, `ONLINE_EVENT_TERM_SLUG`, `is_online_event_term_slug()`, `get_online_event_term_id()` and the JS helpers all want `@since 0.36.0`. This is just PR age, not an oversight.

**3. The coverage-ignore blocks cover more than the unreachable branch.**

```php
// @codeCoverageIgnoreStart
$terms_to_set = array_values( array_unique( $existing ) );

if ( is_wp_error( wp_set_post_terms( $this->event->ID, $terms_to_set, $taxonomy ) ) ) {
	return false;
}
// @codeCoverageIgnoreEnd
```

The comment is right that only a DB error reaches `return false`, but the markers also swallow the `wp_set_post_terms()` call, which is the load-bearing line of the whole method. Same in the toggle-off branch. Hoisting the call out keeps the write measured and ignores only what is genuinely unreachable:

```php
$result = wp_set_post_terms( $this->event->ID, array_values( array_unique( $existing ) ), $taxonomy );

// @codeCoverageIgnoreStart
if ( is_wp_error( $result ) ) {
	return false;
}
// @codeCoverageIgnoreEnd
```

Also, the red CI here is not yours: every PHPUnit job failed at `wp-env start` with `fatal: couldn't find remote ref 7.0.4`, because WordPress.org announced 7.0.4 before the git mirror published that tag. The tag exists now, so a re-run or a rebase onto current develop clears it.

test/unit/php/includes/tests/core/classes/event/class-test-event.php
```diff
@@ -955,6 +957,386 @@ public function test_maybe_get_online_event_link(): void {
...
+		$existing = term_exists( Venue\Setup::ONLINE_EVENT_TERM_SLUG, Venue::TAXONOMY );
+
+		if ( $existing ) {
+			wp_delete_term( intval( $existing['term_id'] ), Venue::TAXONOMY );
+		}
+
+		$filter = static function ( $term, $taxonomy, $args ) {
+			if ( Venue\Setup::ONLINE_EVENT_TERM_SLUG === $args['slug'] ) {
+				return new \WP_Error( 'seed_failed', 'Forced seeding failure.' );
+			}
```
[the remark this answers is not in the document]
faisalahammad: Fixed in `cb5c1d1f`: imported `WP_Error` and changed the test callback to instantiate `new WP_Error(...)`, matching project import conventions.

faisalahammad (commented):
All review threads addressed in `cb5c1d1f`.

- Taxonomy-only sentinel filtering, post/term numeric ID separation, and `useMemo` dependency fix.
- Link preservation on `set_online( true )` without a link; toggle-off still clears term + meta.
- Coverage-ignore narrowing — term writes stay measured; only DB-error branches ignored.
- `@since 0.36.0` on all new API symbols.
- `WP_Error` import + additional regression tests (JS + PHP).

Verification: PHP lint, PHPStan, JS unit tests, and package zip all pass; manual package testing confirmed. PHP/multisite unit tests and JS lint remain blocked on local infra (Docker/OrbStack daemon unavailable; inherited ESLint `wordpress` config missing).

Requesting re-review.

wppoland (commented):
Checked all three at `cb5c1d1f` by reading the file at that ref rather than the summary. All three are resolved, and resolved the way that keeps the behaviour honest rather than the way that just silences the note.

**1. Link preservation.** `update_post_meta()` now sits behind `'' !== $link`, so `set_online( true )` without a link asserts the status and leaves existing meta alone, while `set_online( false )` still deletes it. That makes clearing a link an explicit act, which is what the docblock promised.

**2. `@since`.** 0.36.0 on `is_online()`, `set_online()`, `ONLINE_EVENT_TERM_SLUG`, `is_online_event_term_slug()`, `get_online_event_term_id()` and the JS side. The `postTypes` line moved from 0.35.0 to 0.36.0 too, which I had not flagged and which is correct since 0.35.1 shipped without it.

**3. Coverage markers.** `wp_set_post_terms()` is hoisted out in both the toggle-on and toggle-off branches, so the load-bearing write is measured and only the unreachable `is_wp_error( $result )` guard is ignored. The `wp_get_post_terms()` read guard keeps its marker, which is right: `get_online_event_term_id()` has already proven the taxonomy exists by that point.

CI is green across the board at this head, including the PHPUnit matrix that was previously dying on the `7.0.4` mirror tag.

Nothing further from me. This reads as ready.

faisalahammad:
Fixed 5 failing checks (4 PHP unit on 8.1/8.2/8.4/8.5 + coverage).

Test `test_set_online_off_bails_when_no_term_seeded` still expected `false` after `3bdf4860` changed `set_online(false)` to clear stale link meta and return `true` when no sentinel exists. Updated to `test_set_online_off_clears_meta_when_no_term_seeded` to match (asserts `true` and meta cleared). Coverage failure was downstream.

Local: lint:php pass, phpstan 0 errors, JS 1004 pass. Commit `f272fb60`.

mauteri (changes requested):

includes/core/classes/event/class-event.php
faisalahammad: Simplified to WP core `has_term()`. It matches term IDs, names, and slugs, so the loop over `get_the_terms()` is gone. The taxonomy is still resolved per post type, and a missing post still returns `false`. One small difference: `has_term()` also matches a term literally named `online-event`. The sentinel term name is "Online event" with a space, and venue shadow slugs start with `_`, so this cannot collide here. All existing tests pass unchanged.

includes/core/classes/event/class-event.php
faisalahammad: Done. Every new symbol uses `@since 0.36.0`: `is_online()`, `set_online()`, `ONLINE_EVENT_TERM_SLUG`, `is_online_event_term_slug()`, `get_online_event_term_id()`, and the JS helper `getOnlineEventTermId()`.

includes/core/classes/venue/class-setup.php
```diff
@@ -363,7 +388,7 @@ public function register_starter_pattern(): void {
 		 * is in scope.
 		 *
 		 * @since 0.27.0
-		 * @since 0.35.0 Definitions may include a `postTypes` key to
+		 * @since 0.36.0 Definitions may include a `postTypes` key to
```
mauteri: why'd this change?

mauteri (changes requested):

includes/core/classes/venue/class-setup.php
faisalahammad: Good catch, that was wrong. Reverted in `de6f0ab9`.

I applied an earlier note about moving new symbols to `@since 0.36.0` too broadly and swept up this line, which is not a new symbol. The `postTypes` key shipped in 0.35.0 and is in the 0.35.0, 0.35.1 and 0.35.2 tags, so `@since 0.35.0` was already correct.

The six remaining `@since 0.36.0` tags in this PR are all new symbols.

faisalahammad:
Pushed as `0a993541`.

- Reverted the `@since` bump on the `postTypes` line, details in the thread above.
- Reverted an unrelated `Event` import rename in `class-settings.php`. The BC alias already resolves the old name, and it was the only conflict with develop. That file is out of the diff now.
- Merged current develop, no longer conflicting.

No behavior changed. CI is green.

@mauteri I cannot request a formal re-review without collaborator rights, so pinging here. Could you take another look?

mauteri:
@faisalahammad heads up, #2270 just merged and it renames the post type supports from string literals to class constants:

| Literal | Constant |
| --- | --- |

This branch has one line using the literal form, a `get_post_types_by_support()` call with `gatherpress-venue-information`. Could you merge `develop` in and swap it for `Venue::SUPPORT`? The value is unchanged, so it is just the constant and a `use` statement if one is not already there.

You have the same thing on #2205, where there is rather more of it.

faisalahammad:
Merged develop in and made both changes:

- `get_post_types_by_support( Venue::SUPPORT )` in `add_editor_settings()` now uses the constant from #2270. Value unchanged.
- `use GatherPress\Core\Venue\Setup as Venue_Setup;` in `event/class-event.php`, per the AGENTS.md cross-subsystem alias rule wppoland flagged. Four call sites renamed. No functional change.

Also fixed two CodeRabbit findings on the re-run: the venue combobox now falls back to slug-filtering the `online-event` sentinel when `onlineEventTermIds` has no entry (new Jest test added), and `test_set_online_bails_when_no_event` asserts the sentinel term state is unchanged instead of requiring it absent.

CI-relevant suites pass locally: PHPUnit 2010 tests / 7680 assertions, Jest 64 suites, PHPStan 0 errors, PHPCS clean on changed files. CodeRabbit clean.

@mauteri ready for re-review when you have a moment.

includes/core/classes/event/class-event.php
```diff
@@ -1144,4 +1145,124 @@ public function maybe_get_online_event_link(): string {
...
+	 * @since 0.36.0
+	 *
+	 * @param bool   $is_online True to mark online, false to mark offline.
+	 * @param string $link      Optional URL for the `gatherpress_online_event_link` meta when online. An empty
+	 *                          value preserves an existing link.
+	 *
+	 * @return bool True when the online status was saved, false otherwise.
+	 */
+	public function set_online( bool $is_online, string $link = '' ): bool {
+		if ( ! $this->post ) {
```
mattcowan: Add a check for `Venue::ONLINE_SUPPORT` here. Without it, `set_online( true )` on a post type with only `gatherpress-event-date` writes the term and returns true, but `Online_Event::has_online_event_term()` returns false for the same post.

```suggestion
		if ( ! $this->post || ! post_type_supports( $this->post->post_type, Venue::ONLINE_SUPPORT ) ) {
			return false;
		}
```

includes/core/classes/event/class-event.php
```diff
@@ -1144,4 +1145,124 @@ public function maybe_get_online_event_link(): string {
...
+				// @codeCoverageIgnoreStart
+				// The write only fails on a DB error this harness cannot force.
+				if ( is_wp_error( $result ) ) {
+					return false;
+				}
+				// @codeCoverageIgnoreEnd
+			}
+
+			if ( '' !== $link ) {
+				update_post_meta( $this->post->ID, 'gatherpress_online_event_link', esc_url_raw( $link ) );
```
mattcowan: A non-empty link that esc_url_raw() rejects overwrites a saved link with ''. Skip the write when the escaped value is empty.

includes/core/classes/venue/class-setup.php
```diff
@@ -486,11 +511,13 @@ public function get_venue_meta( int $post_id, string $post_type ): array {
...
 
 		if ( post_type_supports( $post_type, Venue::ASSIGNMENT_SUPPORT ) ) {
-			$event       = new Event( $post_id );
-			$venue_terms = get_the_terms( $post_id, $this->taxonomy_for_event_post_type( $post_type ) );
-			$venue_slug  = ( is_array( $venue_terms ) && ! empty( $venue_terms ) ) ? $venue_terms[0]->slug : null;
+			$event = new Core_Event( $post_id );
+
+			// Derive the online state from the event object so the sentinel
+			// is detected across every assigned term, not just the first.
+			$is_online = $event->is_online();
```
mattcowan, line 518: `Event::is_online()` returns false when the post type has no `gatherpress-event-date` support, because the `Event` constructor does not bind the post. A post type with only `gatherpress-venue` support now gets `isOnlineEventTerm => false` for a post that has the term. The old `get_the_terms()` check returned true. Is this intended?

mattcowan (commented):
Reviewed at a6d21631. PHPUnit (2010 tests) and Jest (1143 tests) pass. I confirmed the toggle-on, toggle-off, and hybrid paths at runtime.

Three small inline comments, all on the new PHP API. I will approve when the Venue::ONLINE_SUPPORT check is submitted. 

@mauteri looks like Faisal addressed your three threads and the Venue::SUPPORT swap from #2270.

includes/core/classes/event/class-event.php
faisalahammad: Added in ae0d7fcf. `set_online()` now bails with the support check you suggested, so it cannot write a term that `Online_Event::has_online_event_term()` would then reject. Added `test_set_online_returns_false_without_online_support` for it.

includes/core/classes/event/class-event.php
faisalahammad: Fixed in ae0d7fcf. `set_online()` now skips the meta write when `esc_url_raw()` returns an empty string, so a rejected link leaves the saved one in place. Added `test_set_online_ignores_link_rejected_by_esc_url_raw` to cover it.

includes/core/classes/venue/class-setup.php
faisalahammad: Yes, intended. A post type with only `gatherpress-venue` support could not render the online-event block on develop either, because `Online_Event::has_online_event_term()` already checks `Venue::ONLINE_SUPPORT` and returns false. So the old true from `get_the_terms()` was the outlier.

In ae0d7fcf I gated `is_online()` on the same support check too, next to the existing null post bail. That way the check is explicit instead of an accident of the constructor not binding the post. `is_online()`, `set_online()`, and the block now all agree on one definition of supports online events.

faisalahammad:
Pushed ae0d7fcf to address all three open comments from @mattcowan.

1. `set_online()` now checks `Venue::ONLINE_SUPPORT` and returns false without writing anything when the post type does not support online events.
2. `is_online()` got the same support check, next to the existing null post bail, so the check is explicit instead of depending on the constructor not binding the post. `is_online()`, `set_online()`, and `Online_Event::has_online_event_term()` now all agree on one definition of supports online events.
3. A link that `esc_url_raw()` rejects no longer overwrites a saved link with an empty string. The rejected value works like an empty link and the previous link stays.

Tests added for all three paths: `test_is_online_returns_false_without_online_support`, `test_set_online_returns_false_without_online_support`, and `test_set_online_ignores_link_rejected_by_esc_url_raw`.

@mattcowan @mauteri this is ready for another look when you have time.

mattcowan (approved):
Approving at `ae0d7fcf`. All three items I flagged in last review are in: I ran the `is_online`/`set_online` tests and re-checked the unsupported post type, the rejected link, and the toggle on/off paths at runtime.

Issue #2056, opened 2026-07-30
No way to cancel or postpone an event


wppoland:
There is currently no way to say "this event is not happening". An organizer can delete the event, unpublish it, or edit the title to add "CANCELLED", and each of those loses something: deleting throws away the RSVP list and the URL people shared, unpublishing 404s a page attendees are still looking at, and a title edit tells nobody. Meanwhile the ICS entry keeps telling every subscribed calendar that the event is on.

I could not find any existing state for this in 0.35.0-beta.1 (no `register_post_status`, no event status meta, nothing in the VEVENT), and I did not find a prior issue or a rejected proposal for it.

### What I would expect

1. A status on the event: scheduled (default), cancelled, postponed. Postponed matters separately because the date is unknown but the event still exists, which is exactly when people need the page to stay up.
2. The single event page saying so, prominently, without the author hand-editing content.
3. Upcoming/past queries treating it correctly, so a cancelled event stops occupying the top of "upcoming".
4. Calendars finding out: `STATUS:CANCELLED` in the VEVENT so subscribed clients grey out or drop the entry.
5. Optionally notifying people who RSVP'd, since they are the reason to keep the page alive at all.

### Where it would hook in

* Storage: a `gatherpress_event_status` post meta, registered in `Event\Meta`, exposed over REST like the other event meta. A custom post status is the other option but fights the block editor's status UI and every query that filters on `publish`.
* Editor: a control in the existing Event settings panel, next to date and duration.
* Frontend: a state on the existing `gatherpress/event-date` block (or a small sibling block) rather than a new template part, so themes get it for free.
* Queries: `Event\Query` already rewrites upcoming/past ordering, so the filter has a home. Whether cancelled events are excluded or shown with a badge should probably be a setting, since a public group usually wants the cancellation visible.
* iCal: `Calendar::build_vevent()` (`includes/core/classes/calendar/class-calendar.php:254`) currently emits URL, DTSTART, DTEND, DTSTAMP, SUMMARY, DESCRIPTION, LOCATION, UID. Adding `STATUS` there is one line, but for clients to actually pick the change up the VEVENT also needs a `SEQUENCE` that increments, which it does not have today. I filed that separately as a prerequisite.
* Notification: the `/gatherpress/v1/event/email` route and the "Send an event update via email" panel already exist, so "notify attendees" can reuse both instead of inventing a second mail path.

[...]

PR #2059, opened 2026-07-30
Add SEQUENCE and LAST-MODIFIED to iCal VEVENTs

### LINKED ISSUE #2058: iCal feeds emit no SEQUENCE, so edited events may never refresh in subscribed calendars
`Calendar::build_vevent()` (`includes/core/classes/calendar/class-calendar.php:254`) emits URL, DTSTART, DTEND, DTSTAMP, SUMMARY, DESCRIPTION, LOCATION and UID. There is no `SEQUENCE` and no `LAST-MODIFIED`.

Per RFC 5545 §3.8.7.4, `SEQUENCE` is the revision number of a calendar component, and clients use it to decide whether an already-imported VEVENT with the same `UID` has changed. GatherPress emits a stable UID (`gatherpress_<ID>`, good) with no revision number, so an event whose date or venue changes can keep showing the original time in a subscribed calendar until the user removes and re-adds the feed. `DTSTAMP` does move with the modified date here, but DTSTAMP is not what the update rules key on.

This also blocks a proper cancellation flow: `STATUS:CANCELLED` on an unchanged `SEQUENCE` is exactly the case clients are allowed to ignore.

### Proposal

Add to each VEVENT:

* `SEQUENCE:` a non-negative integer that increases whenever the event is edited. Rather than a stored counter or a revision count (revisions get pruned or switched off), the seconds between `post_date_gmt` and `post_modified_gmt` work: 0 for a never-edited event, monotonically increasing per event afterwards, and small enough to stay well inside the RFC's integer range.
* `LAST-MODIFIED:` `post_modified_gmt` in UTC form, which is what a human reading the ICS expects to see next to DTSTAMP.

Both are one-liners in the existing `$args` array and affect every feed that goes through `build_vevent()`, including the single-event, taxonomy, and sitewide feeds.

Sending a PR with this and a test.

wppoland:
Fixes #2058

## Proposed changes:

* Every VEVENT now carries `SEQUENCE` and `LAST-MODIFIED` alongside the existing `DTSTAMP`.

`SEQUENCE` is what calendar clients key their update rules on (RFC 5545 §3.8.7.4): a VEVENT repeating a `UID` they already hold is only treated as a revision when the sequence advances. GatherPress emits a stable UID with no sequence, so an event whose date or venue changed could keep showing the original time in a subscribed calendar. This is also the prerequisite for a future `STATUS:CANCELLED` (#2056) being honored.

The value is the seconds between `post_date_gmt` and `post_modified_gmt`: zero until someone edits the event, monotonically increasing for that event afterwards, and clamped to the RFC's integer ceiling. Two alternatives I passed on, both noted in the docblock: post revisions get pruned and can be disabled entirely, which would freeze the sequence at zero; a post meta counter costs a write on every save and has nothing to fall back on for events that predate it.

`LAST-MODIFIED` reports `post_modified_gmt`, which is what a human reading the ICS expects to see next to DTSTAMP.

Both properties flow through `Calendar::get_ical_event_string()`, so the single-event, taxonomy, and sitewide feeds all pick them up.

### Other information:

Four tests added to `Test_Calendar`, covering the VEVENT output and every branch of the new helper (unedited, edited, modification before creation, unparsable dates on either side, and the clamp). The guard branches are exercised through `Utility::invoke_hidden_method` per the extracted-helper coverage note in AGENTS.md.

Local runs, all green: `Test_Calendar` 22 tests / 48 assertions, PHPCS against `phpcs.ruleset.xml`, and PHPStan on the changed class.

## Testing instructions:

1. Publish an event and subscribe to (or download) its iCal endpoint, `/event/<slug>/ical/`.
2. Confirm the VEVENT now contains `SEQUENCE:0` and a `LAST-MODIFIED` matching the event's GMT modification time.
3. Edit the event (any change, for example the excerpt) and fetch the feed again. `SEQUENCE` is now a higher number and `LAST-MODIFIED` has moved. On a real site this is what makes a subscribed client refresh the entry rather than keep the old date.
4. Same check on a taxonomy or sitewide feed, since all of them go through `get_ical_event_string()`.

Verified by hand on WP 7.0.2 / PHP 8.2.29 with GatherPress 0.35.0-beta.1: `SEQUENCE:0` before an edit, `SEQUENCE:3797` after, no notices with `WP_DEBUG_LOG` on.

### DIFF
--- a/.github/changelog/2058-ical-sequence
+++ b/.github/changelog/2058-ical-sequence
@@ -1,4 +1,4 @@
 Significance: minor
-Type: added
+Type: fixed
 
 Added `SEQUENCE` and `LAST-MODIFIED` to every VEVENT in the iCal feeds, so calendar clients recognize an edited event as a revision of the entry they already hold instead of keeping the original date.
--- a/.github/changelog/2058-ical-sequence
+++ b/.github/changelog/2058-ical-sequence
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Added `SEQUENCE` and `LAST-MODIFIED` to every VEVENT in the iCal feeds, so calendar clients recognize an edited event as a revision of the entry they already hold instead of keeping the original date.
--- a/includes/core/classes/calendar/class-calendar.php
+++ b/includes/core/classes/calendar/class-calendar.php
@@ -236,10 +236,9 @@ final class Calendar {
 		$time_end       = $this->event->get_formatted_datetime( 'His', 'end', false );
 		$datetime_start = sprintf( '%sT%sZ', $date_start, $time_start );
 		$datetime_end   = sprintf( '%sT%sZ', $date_end, $time_end );
-		$modified_date  = strtotime( $this->event->event->post_modified );
-		$datetime_stamp = sprintf( '%sT%sZ', gmdate( 'Ymd', $modified_date ), gmdate( 'His', $modified_date ) );
 		$modified_gmt   = strtotime( $this->event->event->post_modified_gmt );
-		$last_modified  = sprintf( '%sT%sZ', gmdate( 'Ymd', $modified_gmt ), gmdate( 'His', $modified_gmt ) );
+		$datetime_stamp = sprintf( '%sT%sZ', gmdate( 'Ymd', $modified_gmt ), gmdate( 'His', $modified_gmt ) );
+		$last_modified  = $datetime_stamp;
 		$sequence       = $this->get_sequence();
 		$venue          = $this->event->get_venue_information();
 		$location       = $venue['name'];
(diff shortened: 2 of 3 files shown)


includes/core/classes/calendar/class-calendar.php
```diff
@@ -266,6 +271,42 @@ public function get_ical_event_string(): string {
...
+	 * the value only saturates for an event edited a lifetime after it was
+	 * created. It is clamped rather than left to overflow.
+	 *
+	 * @since 0.35.0
+	 *
+	 * @return int Non-negative revision number for this event.
+	 */
+	private function get_sequence(): int {
+		$created  = strtotime( (string) $this->event->event->post_date_gmt );
+		$modified = strtotime( (string) $this->event->event->post_modified_gmt );
```
mauteri, line 315: This is the concern from the issue and I don't think the guard above covers it. `$modified <= $created` catches modified-before-created, but the problem is `post_date_gmt` moving *forward* while staying below modified.

Concretely: event created Jan 1, edited Jan 10, we publish `SEQUENCE:777600`. A later edit that also corrects the publish date to Jan 8 emits `SEQUENCE:172800`. That's lower than the value subscribers already hold, so per RFC 5545 clients are entitled to ignore that revision, and every subsequent one, until the delta climbs back past the old high-water mark. It's the bug this PR fixes, reappearing intermittently and much harder to diagnose.

The docblock argues against a post-meta counter and against revision counts, and I agree with both. But `post_modified_gmt` as a plain Unix timestamp avoids all three problems: strictly monotonic per event because `post_modified` only moves forward, no second field to depend on, no write on save, and nothing to backfill. At roughly 1.7 billion it's comfortably inside the INTEGER range you've already documented.

The only thing it gives up is `SEQUENCE:0` for a never-edited event, which nothing needs.

includes/core/classes/calendar/class-calendar.php
```diff
@@ -238,6 +238,9 @@ public function get_ical_event_string(): string {
 		$datetime_end   = sprintf( '%sT%sZ', $date_end, $time_end );
 		$modified_date  = strtotime( $this->event->event->post_modified );
```
mauteri: Worth fixing in this PR since you're adding the correct field right below it.

This is `post_modified`, the site-local time, then formatted with `gmdate()` and stamped `Z`. So on any site not running UTC, DTSTAMP has always been off by the UTC offset.

Line 241 gets it right for LAST-MODIFIED. That means after this PR the same VEVENT carries two timestamps that disagree by the offset, on adjacent lines, which is a worse state than the single quiet bug we had before. Switching this to `post_modified_gmt` fixes it and lets both values share one derivation.

A test with a non-UTC timezone in setup is what catches the regression.

.github/changelog/2058-ical-sequence
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: added
```
mauteri: Minor: `Type: added` puts this under Added, but #2058 is labeled a bug and the user-visible effect is that edited events now refresh in subscribed calendars. `Type: fixed` reads truer for someone scanning the changelog.

mauteri (changes requested):
Thanks, the shape of this is right and the docblock reasoning is good. Two of the notes from the issue are still open though, and the first is a correctness problem rather than a preference, so flagging rather than approving. Details inline.

wppoland:
Thanks @mauteri, all three addressed in the latest push:

1. **SEQUENCE correctness.** Switched to `post_modified_gmt` as a Unix timestamp — strictly monotonic per event, so the publish-date-correction regression can't happen. Dropped the created→modified delta and rewrote the docblock to record why (and why the post-meta counter / revision-count options were also rejected). Gives up `SEQUENCE:0` for a never-edited event, which nothing needs.
2. **DTSTAMP.** Now derived from `post_modified_gmt`, sharing one derivation with LAST-MODIFIED, so the two no longer disagree by the UTC offset on non-UTC sites. Added `test_get_ical_event_string_dtstamp_uses_gmt_on_non_utc_site` (sets `America/New_York`, asserts the GMT value and not the site-local one) as the regression guard.
3. **Changelog.** Retyped `added` → `fixed`.

Sequence tests updated for the timestamp semantics. Full calendar suite green locally (7 tests, 18 assertions), PHPCS clean.

mauteri (approved):

PR #2060, opened 2026-07-30
Add a Duplicate row action for events

### LINKED ISSUE #2057: Add a Duplicate action for events
Most groups run the same event over and over: same description, same venue, same RSVP limits, new date. Today that means opening last month's event, copying the blocks, creating a new event, pasting, re-picking the venue, re-setting the RSVP limit, re-adding topics.

Recurring events (#80) is the full answer to this and it is a large feature. A duplicate action is the small one, and it covers a good share of the same need without committing to a recurrence model.

I did not find this in 0.35.0-beta.1 or in any open or closed issue.

### Proposal

A "Duplicate" row action on the Events list screen (and optionally in the editor's more menu) that creates a draft copy carrying over:

* post content, excerpt, featured image, topics, and the venue term
* the datetime meta, with the start shifted forward so the copy is not in the past, and the same duration and timezone
* RSVP settings meta (max attendance, anonymous RSVP, and friends)

and deliberately not carrying over: RSVPs themselves, comments, the slug, or the published status. The copy lands as a draft so nothing goes public by accident.

### Notes

* `Event\Meta` already knows which keys are event meta, so the copy list can come from there instead of a hardcoded array that rots.
* The datetimes live in the custom table via `Event::save_datetimes()`, so the duplicate has to go through that rather than copying rows.
* The shift default is worth bikeshedding: same weekday next week, or same time tomorrow, or leave the default datetime the editor would use for a fresh event. My preference is the last one, because it is the least surprising and the author is going to set the real date anyway.
* Whether this belongs in core or in a companion plugin is a fair question. It is small enough that a filter to opt out feels like enough.

I am happy to send a PR for this one if you want it in core.

wppoland:
Fixes #2057

## Proposed changes:

* Adds a **Duplicate** row action to the events list table, which copies an event into a fresh draft and opens it in the editor.

What travels: content, excerpt, title, taxonomy terms (topics and the venue, which is a hidden shadow term), the featured image, and every `gatherpress_`-prefixed meta key stored on the source.

What deliberately does not: RSVPs, comments, the slug, and the published status. Those describe the occurrence being copied from.

The datetimes move forward by the smallest whole number of weeks that lands in the future. That keeps the weekday, the time of day, the timezone, and the duration of the original, which is what a group repeating a meetup expects, and it means duplicating a future event still steps a week out instead of stacking two events on the same slot. `gatherpress_datetime` is written for the editor and `Event::save_datetimes()` for the custom table and the derived meta, so the copy is correct in the list table and queries before its first save.

Recurring events (#80) remains the full answer to repetition; this is the small slice that covers most of the day-to-day need without committing to a recurrence model.

Two extension points, both documented in the docblocks: `gatherpress_duplicate_event_postarr` reshapes the copy (renaming it "(copy)" is a one-liner if that is preferred as the default), and `gatherpress_duplicate_event_datetime` places it somewhere other than the weekly shift. `gatherpress_event_duplicated` fires afterwards with both post IDs.

Also in this PR: `Event\Meta::READONLY_DATETIME_KEYS`, so the derived datetime keys have a single definition shared by the REST strip in `filter_readonly_meta()` and the duplicate pass, instead of a literal array in each. Same shape as `Venue\Meta::STRUCTURED_ADDRESS_FIELDS`.

Two calls worth a maintainer opinion:

1. The title is copied verbatim rather than suffixed. A monthly meetup reuses its name, and the draft status distinguishes the copy in the list. Easy to change via the filter or by default if you would rather see "(copy)".
2. `copy_meta()` reads the meta stored on the source instead of `get_registered_meta_keys()`, so meta added later travels with no change here. It skips `gatherpress_datetime` and the read-only derived keys.

### Other information:

`Test_Duplicate` adds 12 tests: hooks, the row action for events and its two refusal paths, a full copy assertion, the weekly shift for past and future events, the no-datetimes bail, all three `duplicate()` guards, both filters, and the action. `handle_request()` is covered by filtering `wp_redirect` to false so the redirect is observable without the exit.

Local runs: full PHP suite 1659 tests with the same 27 pre-existing failures my ad-hoc single-site environment produces on a clean `develop` checkout (missing `gatherpress_events` table plus `wp_`-prefix expectations), so no change in that count; PHPCS and PHPStan clean on all four changed files.

## Testing instructions:

1. Publish an event with a venue, a topic, a featured image, and an RSVP or two. Set an RSVP limit so there is meta to check.
2. Go to **Events** in the admin, hover the event's row, click **Duplicate**.
3. You land in the editor on a new draft. Confirm the content, venue, topic, featured image, and RSVP settings match the original, and that the date is the same weekday and time of day, a week or more later.
4. Confirm the RSVPs did not come along, and that the original is untouched.
5. Duplicate an event that is already in the future: the copy lands exactly one week after it.
6. As a subscriber (or any user who cannot edit the event), confirm no Duplicate link appears, and that hitting the URL directly is refused.

Verified by hand on WP 7.0.2 / PHP 8.2.29 with GatherPress 0.35.0-beta.1: duplicating a July 31 18:00 UTC event produced a draft on August 7 18:00 UTC with a matching 2 hour duration, the same content and excerpt, zero RSVPs on the copy, the `gatherpress_events` row written, and nothing in `debug.log`.

### DIFF
--- a/.github/changelog/2057-duplicate-event
+++ b/.github/changelog/2057-duplicate-event
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Added a Duplicate row action for events. It copies the content, taxonomy terms including the venue, the featured image, and the event meta into a fresh draft, moving the datetimes forward by whole weeks so the copy keeps the original weekday, time of day, and duration without landing in the past. RSVPs, the slug, and the published status stay with the original.
--- a/includes/core/classes/event/class-duplicate.php
+++ b/includes/core/classes/event/class-duplicate.php
@@ -1,0 +1,419 @@
+<?php
+/**
+ * Duplicates an event into a fresh draft.
+ *
+ * Most groups run the same event repeatedly: same description, same venue,
+ * same RSVP limits, new date. This file defines the `Duplicate` class, which
+ * copies everything about an event except the parts that belong to the
+ * original occurrence, so an organizer sets the date and publishes rather
+ * than rebuilding the event from scratch.
+ *
+ * @package GatherPress\Core\Event
+ * @since 0.35.0
+ */
+
+namespace GatherPress\Core\Event;
+
+// Exit if accessed directly.
+defined( 'ABSPATH' ) || exit; // @codeCoverageIgnore
+
+use DateInterval;
+use DateTimeImmutable;
+use DateTimeZone;
+use Exception;
+use GatherPress\Core\Event;
+use GatherPress\Core\Traits\Singleton;
+use GatherPress\Core\Utility;
+use WP_Error;
+use WP_Post;
+
+/**
+ * Duplicates an event into a fresh draft.
+ *
+ * Copies content, taxonomy terms (including the venue), the featured image,
+ * and every author-writable event meta key, then moves the datetimes forward
+ * so the copy is not in the past. RSVPs, comments, the slug, and the
+ * published status deliberately do not come along: those describe the
+ * occurrence that already happened.
+ *
+ * @since 0.35.0
+ */
+final class Duplicate {
+
+	/**
+	 * Enforces a single instance of this class.
+	 */
+	use Singleton;
+
+	/**
+	 * Query action that triggers a duplication request.
+	 *
+	 * @since 0.35.0
+	 *
+	 * @var string
+	 */
+	const ACTION = 'gatherpress_duplicate_event';
+
+	/**
+	 * Class constructor.
+	 *
+	 * This method initializes the object and sets up necessary hooks.
+	 *
+	 * @since 0.35.0
+	 */
+	protected function __construct() {
+		$this->setup_hooks();
+	}
+
+	/**
+	 * Set up hooks for various purposes.
+	 *
+	 * This method adds hooks for different purposes as needed.
+	 *
+	 * @since 0.35.0
+	 *
+	 * @return void
+	 */
+	protected function setup_hooks(): void {
+		add_filter( 'post_row_actions', array( $this, 'row_action' ), 10, 2 );
+		add_action( sprintf( 'admin_action_%s', self::ACTION ), array( $this, 'handle_request' ) );
+	}
+
+	/**
+	 * Add a Duplicate link to an event's row on the posts list table.
+	 *
+	 * Hooked on the generic `post_row_actions` rather than per post type:
+	 * the support check below covers every post type acting as an event,
+	 * including ones companion plugins register after this class loads.
+	 *
+	 * @since 0.35.0
+	 *
+	 * @param array   $actions Row action links keyed by action name.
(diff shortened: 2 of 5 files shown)


mauteri:
@wppoland @carstingaxion I need to sit with this one and read through ticket, PR description, and code to see advantage of building duplicate functionality into the core plugin rather than using a plugin dedicated to duplicating posts like [Yoast Duplicate Post](https://wordpress.org/plugins/duplicate-post/). If taking advantages of hooks available to the Yoast Duplicate Plugin gets us what we need instead of recreating the same/similar functionality, I'd rather do that and advise using the plugin. Similarly, we didn't recreate the functionality of [AQL](https://wordpress.org/plugins/advanced-query-loop/), we just extended it for Events and advised people install AQL along with GatherPress to get more advanced query features.

I'm going to put this in 0.36.0 to review and discuss and see if this is the right direction. Thanks to you both!

carstingaxion:
> @wppoland @carstingaxion I need to sit with this one and read through ticket, PR description, and code to see advantage of building duplicate functionality into the core plugin rather than using a plugin dedicated to duplicating posts like [Yoast Duplicate Post](https://wordpress.org/plugins/duplicate-post/). If taking advantages of hooks available to the Yoast Duplicate Plugin gets us what we need instead of recreating the same/similar functionality, I'd rather do that and advise using the plugin. Similarly, we didn't recreate the functionality of [AQL](https://wordpress.org/plugins/advanced-query-loop/), we just extended it for Events and advised people install AQL along with GatherPress to get more advanced query features.
> 
> 
> 
> I'm going to put this in 0.36.0 to review and discuss and see if this is the right direction. Thanks to you both!

I'm wondering, why this PR exists at all. Wasn't the plan to not change anything on this side, given @mauteri|s comment on the issue.

mauteri:
@carstingaxion I think the PR predated my comment. Ticket was put in, PR was put in, my comment was made.

wppoland:
@carstingaxion right, and @mauteri has it exactly: the issue, the PR and the decision landed in that order. The PR was open before option 3 was chosen, not after, so this is not me relitigating the call.

Closing it. Option 3 is the right shape, and I do not want an eighth unreviewed PR sitting on the queue for a feature core has decided not to own.

One piece is worth carrying over rather than losing, and it is the only genuinely event-specific part of the whole patch: a generic post-and-meta copy leaves no row in `gatherpress_events`, so the duplicate is invisible to `Event\Query` until something saves the post again. That is not something a general duplicator can be expected to know about. I will take it to `duplicate-as` as a small GatherPress-aware fix-up rather than reopening the question here.

The other two open questions from the PR description (verbatim title versus "(copy)", and reading stored meta instead of `get_registered_meta_keys()`) become `duplicate-as` questions too, so nothing is left hanging on this side.

Thanks both for the straight answer on direction.

PR #2063, opened 2026-07-30
Fix bulk actions on the RSVPs screen being silently refused

### LINKED ISSUE #2062: Bulk actions on the RSVPs screen silently do nothing (duplicate _wpnonce field)
Every bulk action on the RSVPs screen (Events > RSVPs) silently does nothing when triggered from a browser. Select rows, pick Approve, Unapprove, Mark as Spam, Not Spam or Delete, hit Apply: the page reloads, no notice, nothing changed.

Reproduced on 0.35.0-beta.1, WP 7.0.2, PHP 8.2.29, default theme, no other plugins, as an administrator.

### Cause

`List_Table::display()` emits a nonce field before delegating to the parent:

```php
public function display(): void {
	wp_nonce_field( Rsvp::COMMENT_TYPE );
	wp_nonce_field( 'gatherpress_rsvp_action', '_gatherpress_rsvp_action_nonce' );

	parent::display();
}
```

`WP_List_Table::display()` then emits its own `wp_nonce_field( 'bulk-' . $this->_args['plural'] )`, which also uses the default field name `_wpnonce`. The rendered form therefore carries two inputs named `_wpnonce`: GatherPress's `gatherpress_rsvp` nonce first, core's `bulk-rsvps` nonce second.

A browser submits both, in document order, and PHP keeps the last value. So `$_REQUEST['_wpnonce']` is core's `bulk-rsvps` nonce, while `process_bulk_action()` only accepts:

```php
wp_verify_nonce( $nonce, Rsvp::COMMENT_TYPE )
|| ( 'delete' === $this->current_action() && wp_verify_nonce( $nonce, 'gatherpress_rsvp_action' ) )
```

Verification fails, the method returns early, and because it returns silently there is no feedback that anything was refused.

### Evidence

Replaying the exact POST the screen produces, changing only which of the two `_wpnonce` values is sent, against two approved RSVPs:

```
BEFORE:                        rsvp 2 approved=1   rsvp 3 approved=1
POST unapprove, first nonce:   rsvp 2 approved=0   rsvp 3 approved=0
POST approve,  first nonce:    rsvp 2 approved=1   rsvp 3 approved=1
POST unapprove, second nonce:  rsvp 2 approved=1   rsvp 3 approved=1   <- no change
```

The second nonce is the one a browser actually sends, which is why the screen appears inert.

Row actions are unaffected: those are GET links carrying a single `_wpnonce` in the query string.

### Fix

Two parts, and I think both are worth doing:

1. Accept core's bulk nonce in `process_bulk_action()`, which is the nonce WordPress itself put on the form: `wp_verify_nonce( $nonce, 'bulk-' . $this->_args['plural'] )`. Smallest change, restores the screen.
2. Stop shadowing `_wpnonce`. The `wp_nonce_field( Rsvp::COMMENT_TYPE )` call in `display()` could use its own field name so the two never collide again, the way the sibling `_gatherpress_rsvp_action_nonce` already does. That is the part that keeps this from coming back.

PR follows with 1, and with 2 if you agree it belongs in the same change.

wppoland:
Fixes #2062

## Proposed changes:

* `process_bulk_action()` accepts core's `bulk-<plural>` nonce, which is the nonce WordPress puts on the bulk form and therefore the one a browser actually sends.
* `display()` emits the comment-type nonce under `_gatherpress_rsvp_nonce` instead of the `_wpnonce` default, so it can no longer shadow core's field. The sibling `_gatherpress_rsvp_action_nonce` already worked this way.

Before this, `display()` wrote a `_wpnonce` for `gatherpress_rsvp` and then `parent::display()` wrote another `_wpnonce` for `bulk-rsvps`. A browser submits both in document order, PHP keeps the last, so the handler received core's nonce, matched neither of the two actions it accepted, and returned early. Approve, Unapprove, Mark as Spam, Not Spam and Delete all did nothing from the UI, and because the refusal is silent there was no notice to explain it.

The nonce reading is now a small loop over the two candidate field names against the accepted actions, so a valid pairing from any of the three entry points (bulk form, view links, row-action delete link) authorizes the action, and `gatherpress_rsvp_action` still only counts for `delete`. The capability check is untouched.

### Other information:

Three tests added to `Test_List_Table`: the core bulk nonce authorizes a bulk action, the dedicated GatherPress field authorizes one, and an unrelated nonce is still refused. `Test_List_Table` runs 74 tests / 113 assertions green locally, PHPCS and PHPStan clean.

## Testing instructions:

1. On an event with a couple of RSVPs, go to **Events > RSVPs**.
2. Tick both rows, choose **Unapprove**, click **Apply**.
3. On trunk: the page reloads and both rows still say Approved. With this branch: both move to Pending and the Approved / Pending counts in the view links update.
4. Repeat with **Approve**, **Mark as Spam** and **Not Spam** to confirm the rest of the set works.
5. Confirm the single-row **Delete** row action still works, since it carries its own nonce.

Verified by hand on WP 7.0.2 / PHP 8.2.29 with 0.35.0-beta.1. I also isolated the cause by replaying the screen's own POST with each of the two `_wpnonce` values against two approved RSVPs:

```
BEFORE:                        rsvp 2 approved=1   rsvp 3 approved=1
POST unapprove, first nonce:   rsvp 2 approved=0   rsvp 3 approved=0
POST approve,  first nonce:    rsvp 2 approved=1   rsvp 3 approved=1
POST unapprove, second nonce:  rsvp 2 approved=1   rsvp 3 approved=1   <- no change
```

The second value is the one a browser sends, which is what made the screen look inert.

### DIFF
--- a/includes/core/classes/rsvp/class-list-table.php
+++ b/includes/core/classes/rsvp/class-list-table.php
@@ -739,17 +739,41 @@ final class List_Table extends WP_List_Table {
 	 * @return void
 	 */
 	public function process_bulk_action(): void {
-		$nonce = isset( $_REQUEST['_wpnonce'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['_wpnonce'] ) ) : '';
+		// Requests reach this method carrying different nonces depending on how
+		// they were made: core's `bulk-<plural>` nonce from the bulk form, the
+		// comment-type nonce from the view links and from `display()`'s own
+		// field, and the dedicated `gatherpress_rsvp_action` nonce from the
+		// row-action delete link, which is why that one only counts for
+		// `delete`. Any single valid pairing is enough; the capability check
+		// below is what authorizes the action.
+		$nonces = array();
 
-		// Accept either the standard comment-type nonce, or — only for the
-		// `delete` action — the dedicated `gatherpress_rsvp_action` nonce
-		// the row-action emits. Cap check folds in for a single guard.
-		$valid_nonce = $nonce && (
-			wp_verify_nonce( $nonce, Rsvp::COMMENT_TYPE )
-			|| ( 'delete' === $this->current_action()
-				&& wp_verify_nonce( $nonce, 'gatherpress_rsvp_action' )
-			)
+		foreach ( array( '_wpnonce', '_gatherpress_rsvp_nonce' ) as $field ) {
+			if ( isset( $_REQUEST[ $field ] ) ) {
+				$nonces[] = sanitize_text_field( wp_unslash( $_REQUEST[ $field ] ) );
+			}
+		}
+
+		$nonce_actions = array(
+			sprintf( 'bulk-%s', $this->_args['plural'] ),
+			Rsvp::COMMENT_TYPE,
 		);
+
+		if ( 'delete' === $this->current_action() ) {
+			$nonce_actions[] = 'gatherpress_rsvp_action';
+		}
+
+		$valid_nonce = false;
+
+		foreach ( $nonces as $nonce ) {
+			foreach ( $nonce_actions as $nonce_action ) {
+				if ( wp_verify_nonce( $nonce, $nonce_action ) ) {
+					$valid_nonce = true;
+
+					break 2;
+				}
+			}
+		}
 
 		if ( ! $valid_nonce || ! current_user_can( Rsvp::CAPABILITY ) ) {
 			return;
@@ -957,7 +981,11 @@ final class List_Table extends WP_List_Table {
 	 * @return void
 	 */
 	public function display(): void {
-		wp_nonce_field( Rsvp::COMMENT_TYPE );
+		// Own field name rather than the `_wpnonce` default: `parent::display()`
+		// emits core's `bulk-<plural>` nonce under `_wpnonce` too, and a browser
+		// sending both leaves PHP with the last one, which used to mean every
+		// bulk action failed verification and did nothing (#2062).
+		wp_nonce_field( Rsvp::COMMENT_TYPE, '_gatherpress_rsvp_nonce' );
 		wp_nonce_field( 'gatherpress_rsvp_action', '_gatherpress_rsvp_action_nonce' );
 
 		parent::display();
--- a/includes/core/classes/rsvp/class-list-table.php
+++ b/includes/core/classes/rsvp/class-list-table.php
@@ -739,19 +739,17 @@ final class List_Table extends WP_List_Table {
 	 * @return void
 	 */
 	public function process_bulk_action(): void {
-		// Requests reach this method carrying different nonces depending on how
-		// they were made: core's `bulk-<plural>` nonce from the bulk form, the
-		// comment-type nonce from the view links and from `display()`'s own
-		// field, and the dedicated `gatherpress_rsvp_action` nonce from the
-		// row-action delete link, which is why that one only counts for
-		// `delete`. Any single valid pairing is enough; the capability check
-		// below is what authorizes the action.
+		// Requests reach this method carrying different nonces under `_wpnonce`
+		// depending on how they were made: core's `bulk-<plural>` nonce from the
+		// bulk form, the comment-type nonce from the view links, and the
+		// dedicated `gatherpress_rsvp_action` nonce from the row-action delete
+		// link, which is why that one only counts for `delete`. Any single valid
+		// pairing is enough; the capability check below is what authorizes the
+		// action.
 		$nonces = array();
 
-		foreach ( array( '_wpnonce', '_gatherpress_rsvp_nonce' ) as $field ) {
-			if ( isset( $_REQUEST[ $field ] ) ) {
-				$nonces[] = sanitize_text_field( wp_unslash( $_REQUEST[ $field ] ) );
-			}
+		if ( isset( $_REQUEST['_wpnonce'] ) ) {
+			$nonces[] = sanitize_text_field( wp_unslash( $_REQUEST['_wpnonce'] ) );
 		}
 
 		$nonce_actions = array(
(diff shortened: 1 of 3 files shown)


includes/core/classes/rsvp/class-list-table.php
```diff
@@ -957,7 +981,11 @@ public function get_views(): array {
 	 * @return void
 	 */
 	public function display(): void {
-		wp_nonce_field( Rsvp::COMMENT_TYPE );
+		// Own field name rather than the `_wpnonce` default: `parent::display()`
+		// emits core's `bulk-<plural>` nonce under `_wpnonce` too, and a browser
+		// sending both leaves PHP with the last one, which used to mean every
+		// bulk action failed verification and did nothing (#2062).
+		wp_nonce_field( Rsvp::COMMENT_TYPE, '_gatherpress_rsvp_nonce' );
```
mauteri: Worth checking whether this field is still needed at all, rather than renamed.

After the fix, a bulk submit carries two nonces that both authorize it: core's `bulk-rsvps` under `_wpnonce`, and this one under `_gatherpress_rsvp_nonce`. Since `process_bulk_action()` now accepts core's, this one is redundant on that form.

And the row actions don't read the form. They carry their own nonce in the query string, minted at line 580 and placed at line 620, with delete overriding to `gatherpress_rsvp_action`.

If nothing actually submits either of these two fields, deleting both lines is a better ending than renaming one: it removes the shadowing risk by removing the field rather than by stepping around the collision. Could you check what, if anything, reads them?

includes/core/classes/rsvp/class-list-table.php
```diff
@@ -739,18 +739,42 @@ public function single_row( $item ): void {
...
+		// below is what authorizes the action.
+		$nonces = array();
+
+		foreach ( array( '_wpnonce', '_gatherpress_rsvp_nonce' ) as $field ) {
+			if ( isset( $_REQUEST[ $field ] ) ) {
+				$nonces[] = sanitize_text_field( wp_unslash( $_REQUEST[ $field ] ) );
+			}
+		}
+
+		$nonce_actions = array(
```
mauteri, line 755: Not a blocker, and not a regression, but this is looser than it needs to be. Every collected nonce is tried against every action, so a comment-type nonce would authorize a bulk delete, and core's bulk nonce would authorize a row action. Neither is intended.

The old code had the same property for the comment-type nonce, so this isn't new. But since you're already in here, binding each source to its own action would be tighter and probably shorter than the nested loop: core's `bulk-<plural>` for the bulk form, the comment-type nonce for row actions, `gatherpress_rsvp_action` for row delete.

Capability checks still gate everything either way, so this is about keeping the nonce meaning what it says rather than about a live hole.

test/unit/php/includes/tests/core/classes/rsvp/class-test-list-table.php
```diff
@@ -783,6 +783,98 @@ public function test_process_bulk_action_empty_ids(): void {
...
+	 * @return void
+	 */
+	public function test_process_bulk_action_accepts_the_core_bulk_nonce(): void {
+		wp_set_current_user( $this->factory->user->create( array( 'role' => 'administrator' ) ) );
+
+		$rsvp_id = (int) $this->rsvp['comment_ID'];
+
+		wp_set_comment_status( $rsvp_id, 'approve' );
+
+		$_REQUEST['_wpnonce']            = wp_create_nonce( 'bulk-rsvps' );
```
mauteri: Small thing: `'bulk-rsvps'` is hardcoded here while the code derives it from `$this->_args['plural']`. If that arg ever changes, this test keeps passing while testing a nonce the screen no longer emits.

Deriving it the same way the code does would keep the test honest.

mauteri (commented):
Nice work, and the test is exactly the assertion that would have caught this. Three notes inline. Only the first one might change the shape of the PR; the other two are fine as follow-ups if you'd rather.

mauteri (changes requested):
Requesting changes on the first of the three inline notes above, the one on `display()`.

The fix itself is right and the test is the correct assertion, so this isn't a rework. I'd just like an answer on whether those two `wp_nonce_field()` calls are still read by anything before we settle on renaming one of them. If nothing submits them, deleting both is the smaller and more durable change, and I'd rather land that than a rename we revisit later.

The other two notes, the nonce-to-action cross-product and the hardcoded `bulk-rsvps` in the test, are fine to address here or leave as follow-ups. Your call.

wppoland:
Thanks @mauteri — checked what reads each field, repo-wide:

- `_gatherpress_rsvp_action_nonce` was read by nothing.
- `_gatherpress_rsvp_nonce` was only collected in `process_bulk_action()`, and on the bulk form it duplicates core's `_wpnonce`/`bulk-<plural>`, which that method already accepts.

So I deleted both fields rather than renaming one. With them gone, `display()` was a pure `parent::display()` passthrough, so I removed the whole override too — that's what closes the #2062 shadowing risk at the source (nothing emits a competing `_wpnonce` anymore) instead of stepping around it. The nonce collection now reads only `_wpnonce`, which every path already carries: the bulk form (core's `bulk-<plural>`), the view links (comment-type), and the row-action delete link (`gatherpress_rsvp_action`).

Tests: dropped the obsolete gatherpress-nonce-field and `display()` tests, and the core-bulk-nonce test now derives `bulk-<plural>` from the table's args instead of hardcoding `bulk-rsvps`. Full `List_Table` suite green locally (74 tests, 116 assertions), PHPCS clean.

On the cross-product note (every collected nonce tried against every action): I've left that as a follow-up rather than fold it in here, since it's a separate, non-blocking behavior change and — with the collection now down to a single `_wpnonce` source — it's cleaner to reason about on its own. Happy to open it separately if you'd like the per-source binding.

mauteri (approved):

Issue #1240, opened 2025-11-27
Automatically Make The Event Creator An Attendee


kofimokome:
### Is your enhancement related to a problem? Please describe.

The person who schedules an event should automatically be an attendee just like on [meetup.com](http://meetup.com/)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #665, opened 2024-05-31
Export & Import RSVPs


carstingaxion:
### Is your enhancement related to a problem? Please describe.

Part 2 of https://github.com/GatherPress/gatherpress/pull/655

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

carstingaxion:
Nice side effect of that could be (with little additional effort), that GatherPress can have _real_ attendees added to the events of the [demo-data](#649) set. 

We, contributors to GatherPress, could either create PRs against the [demo-data-repo](https://github.com/carstingaxion/gatherpress-demo-data) and add our attended events into the import file directly OR we automate this process with names & avatars from the [latest.php](includes/data/credits/latest.php) and let _us_ randomly attend events.

In the end every WordPress Playground created with the demo-data, will show the well-known faces from the Credits screen.

carstingaxion:
If GatherPress defines RSVPs as custom comment-type, like explored in #687, export & import would just work.

mauteri:
@carstingaxion can this be closed since we made RSVP's a comment type now? Maybe we need to test to make sure it works as expected.

carstingaxion:
You might be right and this could eventually be closed, but only after we did some intense testing; like you said @mauteri .

I also haven't done this yet, but I think this should be at least easy to test. Let's see what the results are.

mauteri:
@carstingaxion sounds good, thx!

mauteri:
@carstingaxion we need to check if this can be closed. I'll put in 0.36.0 so we don't forget about it.

PR #2064, opened 2026-07-30
Add check-in for RSVPs

### LINKED ISSUE #2061: No record of who actually attended an event (check-in)
An RSVP records who intended to come. Nothing records who actually turned up, so on the day the organizer counts heads on paper and the site keeps showing an attendee list that includes the people who did not make it. Attendance is the number groups care about afterwards: it is what tells you whether a venue is the right size, what your real no-show rate is, and who to follow up with.

I could not find this in 0.35.0-beta.1 (no check-in state, nothing in the RSVP admin screen) or in any open or closed issue.

### Proposal

Check-in is a small addition on top of what already exists, because RSVPs are comments:

* Storage: a `gatherpress_checked_in` comment meta holding the GMT timestamp of the check-in. Absent means not checked in, so nothing needs backfilling and nothing changes for existing RSVPs.
* Admin: a Checked in column on the existing RSVPs screen, plus Check in and Clear check-in bulk actions, gated on the same `Rsvp::CAPABILITY` the other bulk actions use. That gives an organizer a workable check-in flow on a phone at the door with no new UI surface to design.
* A small public API (`check_in()`, `clear()`, `is_checked_in()`, `count_checked_in()`) so the number is available to blocks, companion plugins, and reporting without anyone reading the meta key directly.

Deliberately out of scope for a first pass, as separate follow-ups if the direction is right: a frontend or editor block showing the checked-in count, a WP-CLI command, and per-attendee check-in from the event screen rather than the RSVP list.

Two things worth deciding before I build:

1. Whether check-in should be restricted to RSVPs whose response is Attending. My inclination is no: someone who said "not attending" and then showed up is exactly the case worth recording, and a gate makes the flow fail confusingly at the door.
2. Whether the timestamp matters or a boolean is enough. I would store the timestamp, since "when did people arrive" is free to keep and impossible to recover later.

Happy to send the PR for the storage, the API, and the admin column and bulk actions.

wppoland:
Fixes #2061

Depends on #2063, and is branched on top of it: the check-in bulk actions are unreachable from a browser until that nonce fix lands. Happy to rebase onto `develop` once it does, or to fold the two together if you would rather review one change.

## Proposed changes:

* New `Rsvp\Check_In` singleton recording which RSVPs actually turned up, in a `gatherpress_checked_in` comment meta holding the GMT arrival time. Absent means not checked in, so no existing RSVP needs backfilling and nothing changes for sites that never use it.
* The RSVPs screen gains a **Checked in** column and **Check in** / **Clear check-in** bulk actions, gated on the same `Rsvp::CAPABILITY` as the moderation actions.

API, so nothing has to read the meta key directly: `check_in()`, `clear()`, `is_checked_in()`, `get_check_in_time()`, `count_checked_in()`. Two actions, `gatherpress_rsvp_checked_in` and `gatherpress_rsvp_check_in_cleared`, both documented in the docblocks.

Behaviour decisions, both raised in the issue:

* **No status gate.** Someone who answered Not Attending and then showed up is exactly the case worth recording, and a gate would make the flow fail confusingly at the door.
* **Timestamp, not a boolean.** Arrival times are free to keep and impossible to reconstruct later. A repeated check-in keeps the first time rather than moving it, because a second scan at the door is a duplicate, not a later arrival.
* `count_checked_in()` counts approved RSVPs only, matching what the attendee list shows.

Deliberately out of scope here, as follow-ups if you want them: a block showing the checked-in count, a WP-CLI command, and per-attendee check-in from the event screen rather than the list.

### Other information:

`Test_Check_In` adds 8 tests covering hooks, recording an arrival, the repeated check-in case, clearing, both writers refusing non-RSVP comments and IDs that do not exist, the approved-only count, the delete sweep, and both actions. `Test_List_Table` gains 3 more for the column in both states and the two bulk actions. Locally: 85 tests / 141 assertions green across those two classes, full suite unchanged at the 27 failures my single-site environment already produces on a clean checkout, PHPCS and PHPStan clean.

## Testing instructions:

1. On an event with two or more RSVPs, go to **Events > RSVPs**. The new **Checked in** column shows a dash for each.
2. Tick both rows, choose **Check in**, click **Apply**. Both rows now show an arrival time in the site's timezone.
3. Apply **Check in** again to the same rows and confirm the times do not move.
4. Tick one row, choose **Clear check-in**, and confirm only that row goes back to a dash.
5. As a user without `moderate_comments`, confirm the bulk actions are not offered.

Verified by hand on WP 7.0.2 / PHP 8.2.29 with 0.35.0-beta.1: bulk check-in from the browser wrote `2026-07-30 10:05:48` to both RSVPs, `count_checked_in()` returned 2, clearing removed both, and `debug.log` stayed empty throughout.

## Summary by CodeRabbit

- **New Features**
  - Added RSVP check-in tracking for event attendance.
  - Approved RSVPs now include check-in status and arrival time in the RSVP admin list.
  - Added individual and bulk actions to check in or clear attendance records.
  - Added checked-in RSVP counts for events.
  - Check-in records are automatically removed when the associated RSVP is deleted.


includes/core/classes/rsvp/class-check-in.php
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * Records which RSVPs actually turned up.
+ *
+ * This file defines the `Check_In` class. An RSVP says who intended to come;
+ * check-in says who arrived. The two are separate on purpose: the gap between
+ * them is the no-show rate, which is what tells an organizer whether a venue
+ * is the right size.
+ *
+ * @package GatherPress\Core\Rsvp
+ * @since 0.35.0
```
mauteri: Need to update to `0.36.0` throughout.

includes/core/classes/rsvp/class-list-table.php
```diff
@@ -483,6 +484,18 @@ public function column_default( $item, $column_name ): string {
...
+
+				if ( '' === $check_in_time ) {
+					return '-';
+				}
+
+				// The stored value is GMT; render it in the site's timezone so
+				// the column reads like the Date column next to it.
+				return esc_html(
+					get_date_from_gmt( $check_in_time, 'Y/m/d \a\t g:i a' )
+				);
```
mauteri: Don't escape here, should escape where it renders.

mauteri (changes requested):

includes/core/classes/rsvp/class-check-in.php
wppoland: Done in 9df02ca. Every `@since` this PR introduces is now `0.36.0`, in `class-check-in.php` and in the new test class. I left the pre-existing `0.35.0` tags in `class-list-table.php` alone since they belong to code that shipped already.

includes/core/classes/rsvp/class-list-table.php
wppoland: Done in 9df02ca. Dropped the `esc_html()` wrapper so the column returns the raw formatted date and escaping happens at the render layer, same as the `date` case right below it. The existing test already asserted the unescaped value, so it stayed green.

wppoland:
@mauteri both of your points were addressed in 9df02ca and this has been green since. Ready for another look when you have time.

- `@since` reads `0.36.0` on all fifteen tags this PR introduces, in `class-check-in.php` and the new test class. The pre-existing `0.35.0` tags on untouched code are left alone.
- The `esc_html()` wrapper is gone from the `checked_in` column, so it returns the raw formatted date and escaping happens at the render layer, matching how the `date` case already worked.

No conflicts with `develop`.

mauteri:
Three things after testing this locally. I merged develop into the branch and swapped the constructor test's hand-rolled reflection for the suite's PMC `Utility` while I was in there; the rest is yours.

## 1. Store check-in state in a taxonomy, not meta

RSVP state already lives in hidden comment taxonomies (`_gatherpress_rsvp_status`, `_gatherpress_rsvp_provider`, set via `wp_set_object_terms` in `Storage`), and `Rsvp\Query::taxonomy_query()` already grafts `tax_query` support into `WP_Comment_Query`. Check-in should follow that pattern instead of introducing a meta `EXISTS` query: a `_gatherpress_rsvp_checkin` taxonomy with a single `checked-in` term, registered in `Rsvp\Setup::register_taxonomy()` alongside the other two.

The mapping is nearly one-to-one:

| Today (meta) | Taxonomy version |
| --- | --- |
| `update_comment_meta( $id, META_KEY, $timestamp )` | `wp_set_object_terms( $id, 'checked-in', TAXONOMY )` |
| `delete_comment_meta( ... )` | `wp_remove_object_terms( ... )` |
| `'' !== get_comment_meta( ... )` | `true === is_object_in_term( $id, TAXONOMY, 'checked-in' )` |
| `meta_query` + `phpcs:ignore SlowDBQuery` | `tax_query` through the existing bridge |

Keep the GMT timestamp in comment meta, but display-only: the term answers every query, the meta only feeds the list table column. Two notes from checking core behavior: `wp_delete_comment()` clears comment meta but never comment term relationships, so your `deleted_comment` hook stays and additionally removes the term. And the timestamp storage is already right, `current_time( 'mysql', true )` in, `get_date_from_gmt()` out.

## 2. Label consistency

Line 780 vs 781: "Check in" and "Clear check-in" (normalize naming). One has the hyphen, the other does not. Pick one form and use it everywhere the feature is named.

## 3. Row actions, not just bulk

Right now checking someone in requires the bulk dropdown. Each row already carries Unapprove | Spam | Delete quick links; add "Check in" / "Clear check-in" there too (showing whichever applies to the row's current state), so working the door is one click per person.

wppoland:
Updated in 8c9ff8d8d:

- Merged latest `develop` to resolve the upstream PHPStan CI failure.
- Registered `_gatherpress_rsvp_checkin` taxonomy on comment objects in `Setup::register_taxonomy()`.
- Switched `Check_In::check_in()` and `clear()` to assign / remove the `checked-in` taxonomy term on the RSVP comment, while retaining GMT timestamps in meta for list table column formatting.
- Updated `Check_In::is_checked_in()` to check taxonomy terms (`is_object_in_term()`).
- Updated `Check_In::count_checked_in()` to use `tax_query` instead of comment meta `EXISTS`.
- Added 1-click `Check in` and `Clear check-in` row actions to the RSVP list table alongside existing row actions for approved attendees.
- Updated PHPCS and PHPUnit tests. All PHPStan level 8 checks and coding standards passing cleanly.

includes/core/classes/rsvp/class-check-in.php
```diff
@@ -0,0 +1,259 @@
...
+// Exit if accessed directly.
+defined( 'ABSPATH' ) || exit; // @codeCoverageIgnore
+
+use GatherPress\Core\Traits\Singleton;
+use WP_Comment;
+
+/**
+ * Records which RSVPs actually turned up.
+ *
+ * State lives in a single comment meta key holding the GMT timestamp of the
```
carstingaxion: Speaks only about the comment meta, but the term relation is not mentioned. I'd add that here too.

includes/core/classes/rsvp/class-check-in.php
```diff
@@ -0,0 +1,259 @@
...
+	use Singleton;
+
+	/**
+	 * Taxonomy for tracking RSVP check-in state.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @var string
+	 */
+	public const TAXONOMY = '_gatherpress_rsvp_checkin';
```
carstingaxion: Im unsure about this, but I would leave the underscore away. I thought our convention is to use the underscore as first indicator for shadow-source taxonomies, but I might be wrong and we (the GatherPress team) have chosen to use this as a convention for non-public taxonomies. A decision I might have been missed.

includes/core/classes/rsvp/class-check-in.php
```diff
@@ -0,0 +1,259 @@
...
+	 * @param int $rsvp_id The comment ID to test.
+	 *
+	 * @return bool True when the comment exists and is an RSVP.
+	 */
+	private function is_rsvp( int $rsvp_id ): bool {
+		$comment = get_comment( $rsvp_id );
+
+		return $comment instanceof WP_Comment && Rsvp::COMMENT_TYPE === $comment->comment_type;
+	}
+}
```
carstingaxion: This looks like it could be already there somewhere in the codebase.

includes/core/classes/rsvp/class-check-in.php
carstingaxion: Ok, now that I read @mauteri|s longer comment a second time, I see. Ignore my previous one.

includes/core/classes/rsvp/class-check-in.php
wppoland: Good catch, @carstingaxion! Updated in d9c1244c8 to clarify both the taxonomy term for status querying and the comment meta for the GMT timestamp.

carstingaxion: Thanks for the update @wppoland.

The last missing piece to look for (from what I can see now) is the `is_rsvp()` method. I was relatively sure, that I had seen such a thing in the past within GatherPress codebase. And yes, indeed; inside `RSVP/Token` we already have `is_valid_rsvp_comment()`, which has the exact same shape as your `is_rsvp()`.

Question for @mauteri : Do we have a nice place for a shared `is_valid_rsvp()` method, that could handle both cases? And should we do that at all?

mauteri: Good find, and it goes further than the one pair: the same check is hand-rolled in about a dozen places across `Form`, `Query`, `Setup`, `Storage`, and the two RSVP blocks, each spelled slightly differently. Let's do it.

**Where:** a public static on `Rsvp\Rsvp`, which already owns `COMMENT_TYPE`, `get()` and `is_enabled()`. Static because most callers, `Token` and `Check_In` included, are handed a bare ID or comment and hold no instance. Accepting either shape covers everything in the codebase:

```php
public static function is_rsvp( int|WP_Comment $comment ): bool {
    $comment = get_comment( $comment );

    return $comment instanceof WP_Comment && self::COMMENT_TYPE === $comment->comment_type;
}
```

**Name:** `is_rsvp()`, not `is_valid_rsvp()`. It answers one question, whether the comment exists and is an RSVP. "Valid" promises more than that (approved, not spam, a good token) and it checks none of those. `Token::is_valid_rsvp_comment()` has the same problem in its name; it only ever did the type check.

**Scope for this PR:** add `Rsvp::is_rsvp()`, call it from `Check_In` in place of the private `is_rsvp()`, and swap `Token`'s private copy since it was named here. The remaining sites are #2240, a mechanical sweep that touches six unrelated files and shouldn't ride along with check-in.

@wppoland, over to you for that change when you get a chance.

wppoland: Done in 27d9ae420.

- Added `Rsvp::is_rsvp( int|WP_Comment $comment ): bool` to `Rsvp\Rsvp`.
- Replaced the private `is_rsvp()` in `Check_In` with `Rsvp::is_rsvp()`.
- Replaced `Token::is_valid_rsvp_comment()` in `Token::__construct()` with `Rsvp::is_rsvp()`.
- Added unit tests in `Test_Rsvp::test_is_rsvp()` and updated `Test_Token`.

mauteri:
Thanks @wppoland and @carstingaxion. Two changes I'd like before this merges, plus a correction and one thing already settled.

## 1. Drop the check-in timestamp

@carstingaxion's privacy point on #2061 lands. Storing arrival times is a liability we can't justify with a concrete use case today, and we shouldn't keep data we have no use for. We can add it back if a real one appears, and adding is far easier than removing once sites have rows.

Concretely: `META_KEY` and both the `update_comment_meta` / `delete_comment_meta` calls go, `get_check_in_time()` goes, `gatherpress_rsvp_checked_in` loses its `$timestamp` argument, and the Checked in column becomes a yes/dash rather than a formatted time.

Two things get simpler. The "a repeated check-in keeps the first time rather than moving it" decision becomes moot, since assigning a term is idempotent on its own. And we stop writing the same fact to two places in two unbatched calls that can drift.

## 2. Make this a general RSVP flag API, with check-in as its first consumer

This is the bigger ask, and I know it grows the PR, so here is the reasoning.

A taxonomy holding one boolean is the shape that's hardest to evolve: narrow now, and a data migration to widen later. What I'd rather have is one place where per-RSVP state lives, that we and companion plugins both write through.

The API isn't only tidiness. `wp_set_object_terms()` without `$append = true` silently wipes every other term on the comment, so two features writing to the same taxonomy directly would clobber each other. A shared API is what makes the taxonomy safe to share at all.

So for this PR:

- Rename the taxonomy to `_gatherpress_rsvp_flag`.
- Add a small public `Rsvp\Flag` API, along the lines of `add()`, `remove()`, `has()`, `get()` (all flags on an RSVP in one call, so a list table with several columns isn't doing a query per column), and a count of RSVPs on an event carrying a given flag.
- `add()` always appends. Nothing else in the codebase should call `wp_set_object_terms()` on this taxonomy.
- `Check_In` becomes a thin consumer of it: `check_in()` adds the `checked-in` flag, `is_checked_in()` asks whether it's there, `count_checked_in()` counts it. The public check-in API in this PR stays exactly as it is.
- The delete sweep should clear *all* flags for the deleted comment rather than just `checked-in`, which matters as soon as there's a second one.
- Generic `gatherpress_rsvp_flag_added` / `gatherpress_rsvp_flag_removed` actions alongside the check-in specific ones would be useful, following `docs/developer/hooks-naming-convention.md`.

I'd keep flag slugs free-form rather than requiring registration, so a companion plugin can add one without coordinating with us.

Doing the rename now rather than later is a one-line constant edit today versus a data migration once sites have `term_taxonomy` rows.

Things that could plausibly live here:

- `auto-added` or `host` for #1240, where the creator automatically attends. That RSVP was never a human action, and telling it apart matters for counts and for not sending organizers a confirmation of their own RSVP.
- `walk-in` for someone with no prior RSVP added at the door, a natural follow-up given this PR deliberately allows check-in regardless of status.
- `no-show`. It looks derivable (attending, event over, not checked in), but only if you know check-in was actually run. Otherwise every attendee looks like a no-show.
- `imported` for #665.
- `first-timer` and `familiar-face`, which Meetup.com surfaces on its RSVP lists. RSVPs already carry an email or user ID, so working these out is a quick query at RSVP or check-in time; the work is agreeing the rules, not running it. Worth noting they have to be written when they're true rather than recomputed later, since whether someone was a first-timer *at that event* stops being recoverable once they've attended more. Whether we build these or a companion plugin does, this is where they'd live.

The boundary matters as much as the list:

- Anything carrying a value stays meta. Guest counts, plus-ones.
- Anything one-of-N stays its own taxonomy. Status and provider are dimensions, not flags.
- Nothing private goes here. `show_in_rest` and `public` apply to the whole taxonomy, so a moderation flag would need its own.

## 3. `delete_check_in()`'s docblock has it backwards

It reads as though WordPress clearing comment meta is what makes the method necessary. It's the other way round: `wp_delete_comment()` does delete commentmeta, but it never touches term relationships. The term is the half core doesn't clean up, so once the meta is gone this method is the only cleanup there is. Worth rewriting so nobody removes it later assuming core handles it.

## 4. The underscore prefix is right

Noting it since @carstingaxion raised and withdrew it. `_gatherpress_venue`, `_gatherpress_rsvp_status` and `_gatherpress_rsvp_provider` all carry it; `gatherpress_topic`, the only user-facing one, doesn't. The registration args here are byte-identical to the existing two RSVP taxonomies, which is what I'd want.

mauteri:
@wppoland one bit of housekeeping before you pick up the review notes above.

The failing PHPStan check on this PR is not your code. CI was running PHPStan with a 2 GB memory limit and no result cache, and the codebase had quietly grown into that ceiling: a passing run on another branch reported 2 GB used against the 2 GB limit, and this branch tips over at 2.05 GB. Nothing you could have fixed from here, and the `Result is incomplete because of severe errors` line it prints is a consequence of the worker being killed rather than a list of type errors.

That is fixed on `develop` now, in #2267, which raises the limit to 4 GB. Merging `develop` into `feat/rsvp-check-in` should turn PHPStan green.

Worth doing in the same pass as the changes in my previous comment so you only rebuild once.

mauteri:
One clarification on my earlier comment, @wppoland: I asked for the flag API in this PR, and that is more than you signed up for. Your call how to split it.

Opened #2268 to hold the API and the flags that could follow. Two ways to take it, and either is fine by us:

1. **Fold it in here.** Rename the taxonomy, add the `Rsvp\Flag` API, and make `Check_In` a thin consumer of it, as described above. One review, one merge.
2. **Keep this PR to check-in.** Rename the taxonomy to `_gatherpress_rsvp_flag` and route writes through a single private helper that always passes `$append = true`, then let the API itself be a separate PR against #2268, from you or from anyone else.

The only part I would ask for either way is the taxonomy name. That is a one-line constant edit now and a data migration once sites have `term_taxonomy` rows, so it is the piece that gets expensive to defer. Everything else can move.

The timestamp removal and the `delete_check_in()` docblock note in my earlier comment still apply in both cases.

wppoland:
Resolved the test coverage check on `class-check-in.php`. All CI checks are now fully passing and green.

wppoland:
@mauteri Implemented Option 2 based on your feedback:

- Renamed the taxonomy constant to `_gatherpress_rsvp_flag` and set `$append = true` in `wp_set_object_terms()` so assigning check-in never clobbers other flags on the RSVP.
- Dropped arrival timestamp tracking (`META_KEY`, `update_comment_meta`, `delete_comment_meta`, and `get_check_in_time()`).
- Dropped the timestamp parameter from the `gatherpress_rsvp_checked_in` action (`do_action( 'gatherpress_rsvp_checked_in', $rsvp_id )`).
- In `List_Table`, the Checked in column now renders `__( 'Yes', 'gatherpress' )` or `'-'`.
- Corrected the `delete_check_in()` docblock to explain that core handles commentmeta on comment deletion but does not touch term relationships.
- Updated unit tests in `Test_Check_In` (including a test verifying that check-in appends without clobbering existing flag terms) and `Test_List_Table`, and updated the changelog entry.

includes/core/classes/rsvp/class-setup.php
```diff
@@ -172,6 +173,22 @@ public function register_taxonomy(): void {
...
+
+		register_taxonomy(
+			Check_In::TAXONOMY,
+			'comment',
+			array(
+				'labels'             => array(),
+				'hierarchical'       => false,
+				'public'             => true,
+				'show_ui'            => false,
+				'show_admin_column'  => false,
```
carstingaxion, line 185: The changelog says "shown in a Checked in column on the RSVPs screen" 

Should the registration or the changelog be updated?

carstingaxion (dismissed):

includes/core/classes/rsvp/class-rsvp.php
```diff
@@ -162,6 +162,35 @@ public function __construct( int $post_id ) {
...
+	/**
+	 * Checks whether a comment ID or object is an RSVP comment.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @param mixed $comment Comment ID or comment object.
+	 *
+	 * @return bool True if the comment exists and has the RSVP comment type, false otherwise.
+	 */
+	public static function is_rsvp( mixed $comment ): bool {
```
mauteri: Two things on `is_rsvp()`, one naming and one shape.

**Name.** `Rsvp::is_rsvp( $comment )` repeats its own subject. What the method actually tests is the comment's type, so `Rsvp::is_comment_type( $comment )` says it plainly and pairs with `get_comment_type()`, which it's the predicate form of.

**Shape.** The version in 27d9ae420 was right, and the follow-up in 453b033e5 overshot. `get_comment()` already accepts an int, a numeric string or a `WP_Comment` and returns null for anything it can't find, so the `mixed` type, the `instanceof` shortcut and the `is_int`/`ctype_digit` triage reimplement the function called on the next line. The one thing in there that earns its place is the `<= 0` guard: `get_comment( 0 )` falls back to the global comment, so in a loop an unrelated comment would read as an RSVP. Keep that, with the reason on it.

```php
public static function is_comment_type( int|WP_Comment $comment ): bool {
	// get_comment( 0 ) falls back to the global comment, which in a loop
	// would make an unrelated comment read as an RSVP.
	if ( is_int( $comment ) && $comment <= 0 ) {
		return false;
	}

	$comment = get_comment( $comment );

	return $comment instanceof WP_Comment && self::COMMENT_TYPE === $comment->comment_type;
}
```

With a typed parameter, `null`, `'invalid'` and `1.5` become the caller's bug and PHP reports it, so those three test cases can go. The `0`, missing-ID, regular-comment and object cases stay.

Small tidy alongside it: `Token` fetches the comment and then does `! $comment || ! Rsvp::is_rsvp( $comment )`. It can hand the ID straight in and drop both the fetch and the null check.

wppoland: Done in e66a24b1:
1. Renamed to `Rsvp::is_comment_type( int|WP_Comment $comment ): bool`.
2. Adopted the suggested simplified shape with the `<= 0` guard for the global comment.
3. Updated `Token::__construct()` to pass `$comment_id` directly to `Rsvp::is_comment_type()`, dropping the intermediate fetch and null check.
4. Cleaned up caller bug test cases (`null`, `'invalid'`, `1.5`) while keeping `0`, missing ID, regular comment, and object assertions.

includes/core/classes/rsvp/class-setup.php
wppoland: The column is rendered directly via `List_Table::get_columns()` on the custom RSVP screen (`'checked_in' => __( 'Checked in', 'gatherpress' )`), so the changelog description is accurate. Setting `'show_admin_column' => false` here prevents WordPress core from trying to inject an automatic taxonomy column into the standard comments list table.

includes/core/classes/rsvp/class-check-in.php
```diff
@@ -0,0 +1,245 @@
...
+	public function check_in( int $rsvp_id ): bool {
+		if ( ! Rsvp::is_rsvp( $rsvp_id ) ) {
+			return false;
+		}
+
+		if ( $this->is_checked_in( $rsvp_id ) ) {
+			return true;
+		}
+
+		wp_set_object_terms( $rsvp_id, self::TERM, self::TAXONOMY );
```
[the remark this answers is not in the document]
mauteri: Done in 853f47275. `check_in()` now returns false without firing when `wp_set_object_terms()` comes back as a `WP_Error`, and the undo, renamed `uncheck_in()`, does the same when `wp_remove_object_terms()` returns anything but true. Unchecking an RSVP that is not checked in counts as done and announces nothing. Covered by `test_check_in_fails_when_the_term_cannot_be_assigned`, `test_uncheck_in_fails_when_the_term_cannot_be_removed`, and `test_uncheck_in_is_idempotent_when_not_checked_in`.

includes/core/classes/rsvp/class-check-in.php
```diff
@@ -0,0 +1,245 @@
...
+
+		$timestamp = current_time( 'mysql', true );
+
+		update_comment_meta( $rsvp_id, self::META_KEY, $timestamp );
+
+		/**
+		 * Fires after an RSVP has been checked in.
+		 *
+		 * @since 0.36.0
```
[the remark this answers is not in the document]
mauteri: No longer applies: the timestamp meta left in 9aec415f0, so the term is the only store and there is no second write to keep consistent. The term write itself is now checked, see the thread above.

includes/core/classes/rsvp/class-check-in.php
```diff
@@ -0,0 +1,245 @@
...
+		wp_remove_object_terms( $rsvp_id, self::TERM, self::TAXONOMY );
+		delete_comment_meta( $rsvp_id, self::META_KEY );
+
+		/**
+		 * Fires after an RSVP's check-in has been cleared.
+		 *
+		 * @since 0.36.0
+		 *
+		 * @param int $rsvp_id The RSVP comment ID.
+		 *
```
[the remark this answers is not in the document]
mauteri: No longer applies: the timestamp meta left in 9aec415f0, so there is no `delete_comment_meta()` here any more. The term removal is checked in 853f47275, and `uncheck_in()` reports failure without announcing anything when it does not succeed.

includes/core/classes/rsvp/class-rsvp.php
```diff
@@ -162,6 +162,25 @@ public function __construct( int $post_id ) {
...
+		if ( ! $comment instanceof WP_Comment && ! is_numeric( $comment ) ) {
```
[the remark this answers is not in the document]
mauteri: Addressed in e66a24b1f. `Rsvp::is_comment_type( int|WP_Comment $comment )` rejects zero and negative IDs before `get_comment()` runs, so the global-comment fallback cannot make an unrelated comment read as an RSVP, and the typed parameter makes decimals and strings a caller error PHP reports.

mauteri (approved):

mauteri:
@wppoland heads up that I pushed two commits to your branch, 96f261f54 and 853f47275, changing the check-in vocabulary. What changed and why:

**Labels.** The pair was "Check in" and "Clear check-in", a verb and a hyphenated noun, and "Undo check in" read like a toggle in the bulk dropdown. Both now follow the table's existing spam pair instead: row actions **Check in** / **Not Checked In**, bulk actions **Mark as Checked In** / **Not Checked In**. No hyphens anywhere, and it reads the same for online events, where "arrived" would not.

**Keys, method, hook.** The undo follows the `un` pattern of `approve` / `unapprove` and `spam` / `unspam`: the action key is `uncheck_in`, the method is `Check_In::uncheck_in()`, and the hook is `gatherpress_rsvp_unchecked_in`. Nothing had shipped, so the rename was free.

**One behavior change** from the CodeRabbit thread that was still valid: both writers now check the term operation before firing their action. A `WP_Error` from `wp_set_object_terms()`, or anything but true from `wp_remove_object_terms()`, returns false with nothing announced, and unchecking an RSVP that is not checked in counts as done, mirroring the idempotent `check_in()`. Three tests cover those paths.

If you have any objections to any of this, let me know and we can handle it in a new PR rather than reopening this one.

PR #2065, opened 2026-07-30
Add an optional local-time line to the Event Date block

wppoland:
## Proposed changes:

* New `showViewerTime` attribute on the Event Date block. With it on, the block also shows the event time in the reader's own timezone.

A remote attendee currently converts the time themselves, and the case that catches people out is the one where the event falls on a different calendar day where they are: an 18:00 New York event is the next morning in Tokyo. The label carries the date only in that case, so the common same-day case stays short.

Only the browser knows the reader's timezone, so `render.php` emits an empty `<span>` carrying the event's GMT datetimes and its own timezone, and `view.js` fills it. Two deliberate silences:

* A reader already in the event's timezone gets nothing. A local group sees no change at all, rather than a duplicate of the time they are already reading.
* Without JavaScript the placeholder stays `hidden`, since an unconverted second copy of the same time would be worse than nothing.

The editor toggle renders the same label through the shared helper, so the author sees what the setting does instead of toggling something invisible. Existing blocks are untouched: the attribute defaults to `false`.

### Other information:

15 Jest tests for the helper (`viewer-time.test.js`): both label shapes, the cross-day case, the four "say nothing" paths, timezone and datetime parsing failures, and a manual UTC offset as the event timezone (which `Intl` does accept, so it formats rather than bailing). 2 PHPUnit tests for the render: the placeholder and its three data attributes when the attribute is set, and its absence by default. `Test_Event_Date` runs 11 tests / 17 assertions green, PHPCS and ESLint clean.

## Testing instructions:

1. Set your site timezone to something far from you, for example `America/New_York` if you are in Europe, and publish an event at 6:00 pm.
2. Add or select the Event Date block, and in the sidebar turn on **Show the reader their local time**. The editor immediately shows your own local equivalent next to the event time.
3. View the event on the frontend. You get, for example, `August 20, 2026 6:00 pm to 8:00 pm EDT` followed by `21/08/2026, 0:00 to 2:00 your time`, with the date present because the event crosses midnight where you are.
4. Change the event timezone to your own and reload: the extra line disappears entirely.
5. Turn the toggle off and confirm the markup goes back to exactly what it was before.

Verified in a browser on WP 7.0.2 / PHP 8.2.29 with 0.35.0-beta.1, from `Europe/Warsaw`: the New York event rendered `21/08/2026, 0:00 to 2:00 your time`, and after switching the event to `Europe/Warsaw` the placeholder stayed hidden and empty.

Props motylanogha


src/blocks/event-date/render.php
```diff
@@ -34,7 +34,41 @@
...
+
+if ( ! empty( $attributes['showViewerTime'] ) ) {
+	$gatherpress_datetime = $gatherpress_event->get_datetime();
+
+	if ( ! empty( $gatherpress_datetime['datetime_start_gmt'] ) ) {
+		$gatherpress_viewer_time = sprintf(
+			'<span class="gatherpress-event-date__viewer-time" data-gatherpress-start-gmt="%1$s" data-gatherpress-end-gmt="%2$s" data-gatherpress-timezone="%3$s" hidden></span>',
+			esc_attr( $gatherpress_datetime['datetime_start_gmt'] ),
+			esc_attr( $gatherpress_datetime['datetime_end_gmt'] ?? '' ),
+			esc_attr( $gatherpress_datetime['timezone'] ?? '' )
```
mauteri: Don't need to escape this here. You'll want to escape when you output which you're doing with `wp_kses` on line 61.

src/blocks/event-date/viewer-time.js
```diff
@@ -0,0 +1,137 @@
...
+	const start = spansDays
+		? formatInTimezone(
+			startGmt,
+			viewer,
+			{ ...dayOptions, ...timeOptions },
+			locale
+		)
+		: viewerStart;
+
+	const viewerEnd = endGmt
```
mattcowan: `viewerEnd` always formats with `timeOptions`, so the end never carries a date even when it falls on a different day for the reader. A 15–17 June event (09:00 New York) shows a London reader `2:00 PM to 10:00 PM your time`, which reads as a single 8-hour afternoon rather than a three-day event. The same thing happens on a single-day evening event whose end crosses midnight for the reader: an 18:00–20:00 New York event gives London` 11:00 PM to 1:00 AM` your time, where 1:00 AM is the next day. `spansDays `only compares the start, so the end is never considered. Worth noting `Event::get_display_datetime()` already solves exactly this — it picks `get_time_end() `vs `get_datetime_end() `based on `is_same_date() `— so there's an existing rule to mirror rather than invent. The changelog entry also promises the date is included "when the event falls on a different day for them", which currently holds only for the start.

Image

src/blocks/event-date/render.php
```diff
@@ -34,7 +34,41 @@
...
 }
+
+// The reader's timezone is only knowable in the browser, so this emits an empty
+// placeholder carrying the event's GMT datetimes and its own timezone for
+// `view.js` to fill. It stays hidden for a reader already in the event's
+// timezone, and for a reader without JavaScript, where an unconverted second
+// copy of the same time would be worse than nothing.
+$gatherpress_viewer_time = '';
+
+if ( ! empty( $attributes['showViewerTime'] ) ) {
```
mattcowan: Both GMT values are emitted regardless of `displayType`, but `get_display_datetime()` honours it (`$show_start`/`$show_end` at class-event.php:208-213). So a block set to show the start only still renders a full range in the local-time line. The editor doesn't do this —` edit.js:270` gates `endGmt` on `showEndTime` — so the author previews a start-only label while the reader gets a range. In the screenshot the block is set to `displayType: 'start'`: the main line correctly reads `June 15, 2030 9:00 am EDT`, while the local-time line still announces `2:00 PM to 10:00 PM your time`. Gating this condition on `displayType` (and gating `startGmt` in the editor too) would bring the two back in line.

Image

src/blocks/event-date/block.json
```diff
@@ -95,6 +99,7 @@
 	},
 	"textdomain": "gatherpress",
 	"editorScript": "file:./index.js",
+	"viewScript": "file:./view.js",
```
mattcowan: This is the only one of the ten blocks here using classic `viewScript`; the other nine all use `viewScriptModule`. Combined with `supports.interactivity.clientNavigation: true` at line 85, core won't disable Query Loop's enhanced pagination for this block, and `view.js` only runs once on `DOMContentLoaded` — so after a client-side page change the newly-inserted placeholders keep their hidden attribute and stay empty. Intended, or should this follow the `viewScriptModule` pattern the rest of the plugin uses?

src/blocks/event-date/style.scss
```diff
@@ -2,3 +2,10 @@
 	// This block has customizable padding, border-box makes that more predictable.
 	box-sizing: border-box;
 }
+
+.gatherpress-event-date__viewer-time {
+	// Reads as a secondary note next to the event's own time rather than a
+	// second competing timestamp.
+	margin-inline-start: 0.5em;
+	opacity: 0.75;
```
mattcowan: `opacity` on real text is theme-dependent in a way a fixed value can't account for: the span inherits the theme's colour, and 0.75 over white pushes anything lighter than roughly `#444` below the 4.5:1 floor. A theme using `#555` body copy goes 7.46:1 → 3.98:1, and one sitting on the AA-minimum `#767676` goes 4.48:1 → 2.85:1 (WCAG 1.4.3). The other opacity uses in the plugin are admin/editor surfaces where the palette is known — this is the first on frontend text the theme owns. `src/formats/tooltip/style.scss` exposes `--gatherpress-tooltip-bg-opacity` as a custom property, which would let a theme opt out here.

src/blocks/event-date/viewer-time.js
```diff
@@ -0,0 +1,137 @@
...
+export function getViewerTimeLabel( {
+	startGmt = '',
+	endGmt = '',
+	eventTimezone = '',
+	viewerTimezone = undefined,
+	locale = undefined,
+} = {} ) {
+	const viewer = viewerTimezone ?? getViewerTimezone();
+
+	if ( ! startGmt || ! viewer || viewer === eventTimezone ) {
```
mattcowan: The same-zone check is a string compare, so it only suppresses the line on an exact match. A site configured with a UTC offset rather than a city defeats it for every reader: with the event timezone stored as `+05:30`, a reader in Asia/Kolkata gets `5:30 PM to 7:30 PM your time` next to the identical time they're already reading. `wp_timezone_string()` returns offsets for offset-configured sites, so this isn't rare. Comparing resolved offsets for the event instant rather than the zone names would cover it.

src/blocks/event-date/viewer-time.js
```diff
@@ -0,0 +1,137 @@
...
+	} catch {
+		// Intl is unavailable or refused the call: no viewer time to offer.
+		return '';
+	}
+}
+
+/**
+ * Format a GMT datetime in a given timezone.
+ *
+ * Returns an empty string rather than throwing when the timezone is not one
```
mattcowan, line 20: This says `Intl` only takes IANA names, but the test at `viewer-time.test.js:89 `asserts the opposite ("handles a manual UTC offset, which Intl accepts as a timezone") and passes. Worth correcting so nobody later "fixes" the offset handling on the strength of the docblock.


```suggestion
 * Returns an empty string rather than throwing when the timezone is not one
 * `Intl` accepts. GatherPress allows manual UTC offsets ("+05:30") as an event
 * timezone, which `Intl` does accept alongside IANA names.
```

src/blocks/event-date/viewer-time.js
```diff
@@ -0,0 +1,137 @@
+/**
+ * WordPress dependencies
+ */
+import { __, sprintf } from '@wordpress/i18n';
+
+/**
+ * Resolve the timezone the browser is running in.
+ *
+ * @since 0.35.0
```
mattcowan: Small update, this should be .36. Six new tags are affected: `viewer-time.js` lines 9, 29 and 72, `view.js` line 15, and `class-test-event-date.php` lines 318 and 374.

mattcowan (commented):
Thank you for submitting this wppoland. This is a good feature, and I like that it stays quiet when it has nothing useful to add: a reader already in the event's timezone sees no change at all. Running the editor preview through the same helper the frontend uses is a nice touch too, since the author sees exactly what a reader will get.

One thing should be fixed before it lands. The end time is always shown as a bare clock time with no date attached. A 15-17 June event tells a London reader 2:00 PM to 10:00 PM your time, which looks like a single afternoon - the event actually ends at 10:00 PM two days later. The same thing happens on a one-day evening event when the end crosses midnight for the reader. That is the confusion this feature exists to prevent, so it is worth getting right.

The rule already exists in the codebase, which should make this an easy fix: get_display_datetime() checks is_same_date() and prints a full end date when the event spans days, instead of just a time.

Six smaller notes inline: the local-time line ignores displayType and disagrees with the editor preview; the classic viewScript will not re-run after client-side navigation, where every other block here uses viewScriptModule; opacity: 0.75 on theme-coloured text can fall below the AA contrast floor; the same-timezone check misses sites configured with a UTC offset; one docblock contradicts its own test; and the @since tags need to be 0.36.0 now that 0.35.0 has shipped.

mattcowan (dismissed):
Changes look good to merge, several look better than originally proposed, will need final approval by mauteri. Thank you!

wppoland:
Thanks @mattcowan, and @mauteri here is what changed since the review, since a few of these went further than the comment asked.

## The two that were real defects

**The end never carried a date.** `viewerEnd` always formatted with `timeOptions`, so a multi-day event or an evening event crossing midnight for the reader read as a single afternoon. It now mirrors the rule already in `Event::get_display_datetime()`, `is_same_date()` deciding between `get_time_end()` and `get_datetime_end()`, measured in the viewer's calendar rather than the event's.

**`displayType` was ignored.** Both GMT values went out regardless, so a start-only block still announced a range. Both sides now derive from the same `in_array( $type, array( 'start', 'both' ), true )` rule the class uses, and `edit.js` gates `startGmt` too.

## Three things the review did not ask for

**An end-only block now converts its end** rather than rendering nothing. Gating strictly on the start was the literal reading, but it silently dropped the feature for that one setting while the changelog promises every reader a converted time. `get_display_datetime()` shows the end alone there, so the label does too.

**`sprintf` semantics restored.** A script module cannot import `@wordpress/i18n`, so the sentence was filled with `String.replace` on a literal pattern. That fills only the first copy of a placeholder, leaves `%%` doubled, and cannot see `%1$s` as the same placeholder as `%s`. The last one reaches readers: translators normalize a lone `%s` to `%1$s` routinely, and such a locale rendered the raw token instead of a time. Now a positional-tolerant regex, the same shape `helpers/interactivity.js` uses for the attendee count.

**An empty `displayType`** read as both in `render.php` and as neither in `edit.js`. Unreachable through the UI, since `block.json` defaults to `both`, but it is the same author-versus-reader mismatch, so both sides normalize it.

## Two residuals, documented rather than fixed

`isSameZoneAt()` compares resolved offsets, which closes the `+05:30` case, but it depends on `Intl` accepting offset time zones. An engine predating that ECMA-402 change returns an empty string from `formatInTimezone()`, the equality fails, and the redundant line comes back. The docblock says so now rather than asserting acceptance unconditionally.

The comparison is also taken at the start instant only, so a DST transition inside the event window in exactly one of the two zones would be missed. It needs a fixed-offset or no-DST counterpart to happen at all, and the start is the instant readers plan around, so it is noted in the code rather than covered.

## Verification

JS 1024 passing across 54 suites, PHP 1651 tests and 6903 assertions single site, 323 and 856 multisite. `lint:php`, `lint:js`, `lint:css` and `lint:phpstan` all clean.

Negative controls rather than green ticks: reverting the three source changes fails exactly seven tests, one per fix, and nothing else.

Checked in a real browser against a Query Loop with enhanced pagination on: `data-wp-router-region` is present, so core did not disable it, and after a client-side page change the second event's line renders correctly with no full page load. The control run on the pre-fix build reproduced both the undated end and the empty start-only line.

Branch is behind `develop`; say the word if you want it brought forward before merge.

mauteri (changes requested):
@wppoland Thanks for the rework on Matt's feedback, all eight of those look addressed to me.

Stepping back from the details though, I'd like to change the shape of this before it goes in.

Screenshot 2026-08-30 at 11 44 25 AM

Two times sitting side by side makes for a long string, and the viewer has to work out which one applies to them.

There's also a conflict with something we already have. `User::user_set_timezone()` hooks the `gatherpress_timezone` filter, which fires inside `Event::get_datetime()`, so a logged-in user with a saved timezone preference already gets the main line rendered in their timezone server-side. If their saved preference differs from their browser, travelling or deliberately planning in another zone, this block shows them their saved timezone on the main line and a browser-derived "your time" next to it. Two different answers to the same question, with the browser guess contradicting a setting they made explicitly.

So I'd like this to extend that feature rather than sit beside it. What I'm picturing:

* The event's own time stays as the visible datetime, and the viewer's local time moves into a tooltip on it. That keeps the line short while still leaving both times reachable, which matters for someone in London who is actually travelling to the New Mexico event.
* Allow the toggle only when "Append time zone" is on, so whichever time is showing is always labelled.
* Add a global setting, off by default, that organizers turn on when it suits their events. Gate it with the `show_if` property in the Settings API so it hides when the timezone setting is off (`class-settings.php:505-522` has the rendering side).

One dependency: our tooltip is currently mouse-only, so putting the local time behind it today would make it unreachable by touch or keyboard. I've filed #2217 to fix that, and this should build on it rather than work around it. Worth waiting for that to land.

Two smaller things:

Let's use "viewer" consistently. The identifiers already do (`showViewerTime`, `viewerTimezone`, `getViewerTimeLabel`, the `viewer-time` file and class), but the comments and UI strings say "reader", including `Show the reader their local time` and its help text. "Reader" is what we use for screen readers everywhere else, and #2201 sitting alongside this is entirely about screen readers, so it's worth not overloading. I've opened #2216 to write these rules down properly.

When you get to the tooltip markup, pair it with a visually hidden `.screen-reader-text` span carrying the same string. `aria-describedby` can't point at a `::after` pseudo-element, so hidden real text is the practical route.

Happy to talk it through if you'd rather sketch the approach before writing code.

wppoland:
Thanks for the thoughtful review, @mauteri. The tooltip approach makes a lot of sense - it keeps the line clean and naturally avoids the visual clutter of having two long datetime strings side-by-side.

I agree with all points:
1. Moving the viewer's local time into a tooltip on the event's datetime, paired with a visually hidden `.screen-reader-text` span for accessibility.
2. Enabling the toggle only when "Append time zone" is active.
3. Adding the global setting gated with `show_if` in the Settings API (`class-settings.php`).
4. Consistent "viewer" terminology across comments and UI strings (per #2216).

I'll put together a PR for #2217 to get the tooltip touch/keyboard/a11y support landed first, and then re-shape this PR on top of it once that's in.

wppoland:
@mauteri I have updated this PR to reshape the feature into the tooltip approach based on your feedback:

1. **Global Setting**:
   - Added a `show_viewer_timezone` checkbox under **Events > Date & Time** (defaults to disabled).
   - Configured with a `show_if` dependency so it is only visible when `show_timezone` is enabled.

2. **Tooltip Architecture**:
   - Merged `develop` to leverage the shared GatherPress tooltip system (from #2227).
   - The Event Date block-level toggle is now gated on both the global setting and the block display setting having timezone visible.
   - When active and the viewer is in a different timezone from the event, the datetime element (or the anchor tag if `isLink` is true) dynamically binds `data-wp-class--gatherpress-tooltip`, `data-wp-bind--data-gatherpress-tooltip`, and `tabindex="0"`.
   - Screen readers receive the converted time via a nested `.screen-reader-text` element.
   - Viewers in the same timezone see the datetime as usual with no tooltip markup or extra notes.

3. **Editor Experience & Tests**:
   - The editor inspector and canvas reflect the tooltip presentation and global setting status.
   - All PHPUnit and Jest test suites have been updated to cover the tooltip bindings, context data, and suppression logic.
   - All linters (`lint:js`, `lint:php`, `lint:phpstan`) and test suites pass cleanly.

Ready for review!

wppoland:
All unit tests, lints, and CI checks are now 100% green across all PHP versions (8.1, 8.2, 8.4, 8.5) and multisite. Ready for review when you have a moment!

carstingaxion:
With no changes to any settings, the tooltip reads kind of unconformtable to me. I'm not sure if I do not understand or translate wrong OR if the text should be adjusted.

image

_Using the PR Playground with WordCamp Sydney to test._

src/blocks/event-date/render.php
```diff
@@ -26,15 +27,89 @@
...
+				// The view script is a module and cannot import `@wordpress/i18n`,
+				// so the sentence it fills in is translated here instead.
+				/* translators: 1: event start in the viewer's timezone, 2: event end in the viewer's timezone. */
+				'rangeFormat'   => __( '%1$s to %2$s your time', 'gatherpress' ),
+				/* translators: %s: event start in the viewer's timezone. */
+				'singleFormat'  => __( '%s your time', 'gatherpress' ),
+			),
+			JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP
+		);
+	}
```
carstingaxion, line 78: I guess this is the relevant code that should show a nice rangeFormat for the Sydney event (I commented earlier).

carstingaxion: Ah, and now I understand what makes this uncomfortable: Its showing the bare end-time.

carstingaxion:
Tested in the PR Playground and can confirm

- [x] global settings are accepted
- [x] block settings are accepted
- [x] works with linked and unlinked event-date block
- [x] (when enabled) is hidden when event timezone and client timezone match
- [x] is shown nicely when timezones differ
- [ ] is shown buggy for events accross multiple days (nnot All-day events, but with start and end times)
- [ ] shows the start and end time in the tooltip, even the block hides it, when is an All-day event
   image

wppoland:
Thank you for the thorough testing and screenshots from the Playground, @carstingaxion! Both reported issues are resolved in commit 31a8d952:

### 1. Events across multiple days (e.g. WordCamp Sydney)
- **Problem:** When an event spans across days in the viewer's calendar, the start time was formatted with only the clock time while the end time had a full date (e.g. `8:00 to 3.11.2024, 10:00 your time`), making the range look asymmetric or like the whole event took place on the end date.
- **Fix:** When the range spans across calendar days for the viewer (`rangeSpansDays`), both start and end now carry the date (e.g. `2.11.2024, 8:00 to 3.11.2024, 10:00 your time`), making multi-day intervals clear and unambiguous.

### 2. All-day events
- **Problem:** All-day events intentionally hide start and end times in the block, but the viewer time tooltip was converting midnight timestamps to the viewer's local timezone (e.g. `1.11.2024, 15:30 to 3.11.2024, 15:29 your time`).
- **Fix:** 
  - Suppressed viewer time tooltip context in `render.php` when `$gatherpress_event->is_all_day()` is true.
  - Added `isAllDay` guard in `getViewerTimeLabel()` in `viewer-time.js`.
  - In `edit.js`, disabled the inspector toggle for all-day events with descriptive help text (`All-day events do not show viewer local time.`).

### 3. Verification & Tests
- `npm run test:unit:js` (all 67 suites, 1,196 tests pass)
- `npm run lint:js` & `npm run lint:css` (0 errors)
- `npm run build` (compiled cleanly)
- `vendor/bin/phpcs` (0 errors, 0 warnings)
- `composer test:phpstan` (0 errors across 145 files)

Ready for testing in the Playground!

src/blocks/event-date/render.php
```diff
@@ -26,15 +27,90 @@
...
+<div <?php echo wp_kses_data( $gatherpress_wrapper_attrs ); ?>>
+	<?php if ( ! empty( $attributes['isLink'] ) ) : ?>
+		<?php if ( $gatherpress_viewer_time_context ) : ?>
+			<a
+				href="<?php echo esc_url( get_permalink( $gatherpress_post_id ) ); ?>"
+				data-wp-class--gatherpress-tooltip="state.hasViewerTime"
+				data-wp-bind--data-gatherpress-tooltip="state.viewerTimeLabel"
+			>
+				<?php echo esc_html( $gatherpress_display ); ?>
+				<span class="screen-reader-text" data-wp-text="state.viewerTimeSrLabel"></span>
```
mattcowan, line 99: Add `gatherpress--screen-reader-text gatherpress-tooltip-notice` to this span. `initTooltip()` in `src/formats/tooltip/view.js` only skips its own screen-reader span when a direct child has the `gatherpress-tooltip-notice` class, so once the view module has applied the `data-wp-class` and `data-wp-bind` directives it appends a second one and a screen reader hears the converted time twice. The same applies to the span at line 111 here and to the two spans in `edit.js` at lines 352 and 366.

```suggestion
				<span class="screen-reader-text gatherpress--screen-reader-text gatherpress-tooltip-notice" data-wp-text="state.viewerTimeSrLabel"></span>
```

src/blocks/event-date/render.php
```diff
@@ -26,15 +27,90 @@
...
 
-// Mirrors core/post-date's isLink attribute: link the datetime to the event.
-if ( ! empty( $attributes['isLink'] ) ) {
-	$gatherpress_display = sprintf(
-		'<a href="%s">%s</a>',
-		esc_url( get_permalink( $gatherpress_post_id ) ),
-		$gatherpress_display
+$gatherpress_settings           = Settings::get_instance();
+$gatherpress_show_timezone_attr = $attributes['showTimezone'] ?? '';
+$gatherpress_is_timezone_active = '' !== $gatherpress_show_timezone_attr
```
mattcowan, line 32: This gate does not see the per-event `gatherpress_show_timezone` meta, but `get_display_datetime()` answers `always` and `never` before the block is asked. An event set to `never` shows no zone but still gets the tooltip. An event set to `always` with block `showTimezone: 'no'` shows the zone but gets no tooltip, and the editor toggle says "Time zone must be appended" next to a preview that appends it. `edit.js` line 238 has the same gap; `timezonePreference` is already in scope there.

```suggestion
$gatherpress_timezone_preference = $gatherpress_event->get_timezone_preference();
$gatherpress_is_timezone_active  = 'never' !== $gatherpress_timezone_preference
	&& (
		'always' === $gatherpress_timezone_preference
		|| ( '' !== $gatherpress_show_timezone_attr
			? 'yes' === $gatherpress_show_timezone_attr
			: (bool) $gatherpress_settings->get( 'show_timezone' ) )
	);
```

mattcowan (commented):
Re-reviewed at `6df233c6`. Both issues Carsten found are fixed: I ran the Sydney range for Berlin and New York viewers, all-day, same-zone, offset-site, and a positional translation, and each comes out right. PHPUnit (2036) and Jest (1196) pass.

Two inline comments. The first is a duplicate screen-reader announcement from the shared tooltip script, fixed by one class on the span. The second is the "zone is shown" gate not honoring the per-event `always`/`never` preference that the datetime itself honors. I will approve when both land.

@mauteri I believe the shape now matches your Aug 30 ask (tooltip, global setting with `show_if`, toggle gated on the appended zone, "viewer" wording).

wppoland:
Both suggestions landed in 62acbf76!

- Added `gatherpress--screen-reader-text gatherpress-tooltip-notice` to the screen-reader spans in `render.php` and `edit.js`, ensuring `initTooltip()` skips appending a duplicate screen-reader span.
- Gated timezone display in both `render.php` and `edit.js` using `get_timezone_preference()` so per-event `always` / `never` preferences are respected.

Ready for another look. Thank you @mattcowan!

mattcowan (approved):
Thank you @wppoland, re-reviewed at 62acbf76 and it looks like my comments are resolved, so I will approve, but two small items before merge: the branch conflicts with `develop`. Also change the new `@since 0.36.0` tags to `@since TBD`. The check added in #2326 after my earlier review now requires this.

PR #2066, opened 2026-07-30
Add a public API for online events

### LINKED ISSUE #2047: Add a public API for online events (setter, term helpers, REST parity)

wppoland:
Fixes #2047 (parts 1 and 2)

## Proposed changes:

Part 1, the missing getter and setter on `Event`:

* `is_online(): bool` reads the online-event term directly. Distinct from `maybe_get_online_event_link()`, which answers "should this visitor see the link right now" rather than "is this event online".
* `set_online( bool $is_online, string $link = '' ): bool` owns the term and the link meta together. Marking an event online sets the term and stores the link; turning it off removes the term and deletes the link, so a room URL is not left behind on an event that moved to a venue. Re-marking online without a link clears the old one rather than leaving a stale value under a new setting.

Part 2, the slug:

* `Venue\Setup::ONLINE_EVENT_TERM_SLUG`, so the literal has one home.
* `is_online_event_term_slug()`, symmetric with the existing `is_venue_term_slug()`. The two are deliberately mutually exclusive, which the tests assert.
* `get_online_event_term_id( string $event_post_type = '' ): ?int`. Returns null while the sentinel term does not exist yet, which is the ordinary state between activation and `add_online_event_term()` running, not an error.
* The resolved term ID rides along in the editor settings that already carry `venuePostTypes`, as `onlineEventTermId`, so components can stop each running their own `getEntityRecords` lookup for the same term.
* The three PHP call sites that had the literal now use the constant.

Part 3 is deliberately untouched: the issue asks for the `is_online` REST representation to be settled before anyone writes it, and I agree a registered field reading through to the term beats duplicating the state in meta.

The JS literals listed in the issue are also left for a follow-up. `onlineEventTermId` is what those six call sites need, and swapping them is a separate, mechanical change that deserves its own review rather than riding along here.

This is what [WordPress/wordcamp.org#1825](https://github.com/WordPress/wordcamp.org/issues/1825) is waiting on, so the groups site can drop its own term-plus-meta coordination.

### Other information:

4 tests on `Test_Event` for the getter and setter: no terms, the full online / offline round trip including link clearing, online without a link, and both methods bailing when the instance wraps no post. 3 on `Venue\Test_Setup` for the predicate, the term-ID resolver in both its states, and the editor settings exposure.

`Venue\Test_Setup` is `@group multisite`, so those three only run in the multisite suite. Local runs of both: single-site 1651 tests, multisite 326 tests, in each case with exactly the same failures a clean `develop` produces in my environment (27 and 2, all from the missing `gatherpress_events` table and `wp_`-prefix expectations). PHPCS and PHPStan clean.

## Testing instructions:

1. On an event with no venue, run `( new \GatherPress\Core\Event\Event( $id ) )->is_online()` and confirm `false`.
2. Call `set_online( true, 'https://meet.example.org/room' )`. The event now carries the `online-event` term, `gatherpress_online_event_link` holds the URL, and `is_online()` is `true`. The Online Event block renders the link as before.
3. Call `set_online( false )`. The term is gone, the link meta is deleted, and `is_online()` is `false`.
4. `Venue\Setup::get_instance()->get_online_event_term_id()` returns the term ID, and `is_online_event_term_slug( 'online-event' )` is `true` while `is_venue_term_slug( 'online-event' )` stays `false`.
5. In the editor, `wp.data.select( 'core/editor' ).getEditorSettings().gatherpress.config.onlineEventTermId` is that same ID.

Verified on WP 7.0.2 / PHP 8.2.29 with 0.35.0-beta.1:

```
term id helper: 2
is_online_event_term_slug(online-event): true   is_online_event_term_slug(_a-venue): false
before: is_online=false
after set_online(true):  is_online=true  link=https://meet.example.org/beta-room  terms: online-event
after set_online(false): is_online=false link=""                                  terms: (none)
```

No notices with `WP_DEBUG_LOG` on.

Props motylanogha

### DIFF
--- a/.github/changelog/2047-online-event-api
+++ b/.github/changelog/2047-online-event-api
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Added a public API for online events: `Event::is_online()` and `Event::set_online()`, which own the online-event term and the link meta together, plus `Venue\Setup::ONLINE_EVENT_TERM_SLUG`, `is_online_event_term_slug()` and `get_online_event_term_id()`. The resolved term ID is exposed to the block editor so components stop looking it up individually.
--- a/includes/core/classes/blocks/class-online-event.php
+++ b/includes/core/classes/blocks/class-online-event.php
@@ -13,7 +13,7 @@ namespace GatherPress\Core\Blocks;
 defined( 'ABSPATH' ) || exit; // @codeCoverageIgnore
 
 use GatherPress\Core\Traits\Singleton;
-use GatherPress\Core\Venue\Setup;
+use GatherPress\Core\Venue\Setup as Venue_Setup;
 use WP_Block;
 
 /**
@@ -116,14 +116,18 @@ final class Online_Event {
 			return false;
 		}
 
-		$taxonomy    = Setup::get_instance()->taxonomy_for_event_post_type( $event_post_type );
+		$taxonomy    = Venue_Setup::get_instance()->taxonomy_for_event_post_type( $event_post_type );
 		$venue_terms = get_the_terms( $post_id, $taxonomy );
 
 		if ( ! is_array( $venue_terms ) ) {
 			return false;
 		}
 
-		return in_array( 'online-event', wp_list_pluck( $venue_terms, 'slug' ), true );
+		return in_array(
+			Venue_Setup::ONLINE_EVENT_TERM_SLUG,
+			wp_list_pluck( $venue_terms, 'slug' ),
+			true
+		);
 	}
 
 	/**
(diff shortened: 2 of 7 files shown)


wppoland:
Closing in favour of #2049, which predates this by a couple of days and implements the same public API for #2047. I have left my two design observations (constant placement, and `set_online()` returning `bool` so a failed write is detectable) as review notes over there rather than keeping a competing PR open.

PR #2067, opened 2026-07-30
Cache the iCalendar responses and answer conditional requests

### LINKED ISSUE #1682: Add HTTP cache headers + object-cache backing for ICS feeds

Calendar apps subscribe to ICS feeds and poll them frequently — Outlook checks every 15-60 minutes, Google Calendar every 24 hours (configurable down to shorter intervals), Apple Calendar between 5 minutes and a week. Each unauthenticated hit on `/event/<slug>/ical`, `/feed/ical`, `/event/feed/ical`, `/venue/<slug>/feed/ical`, `/topic/<slug>/feed/ical`, etc. currently regenerates the full ICS payload via `Calendar\Setup::send_ics_file()` → `get_ical_feed()` / `get_ical_file()` → `Calendar::get_ical_event_string()` for every queried event.

For active sites with a few hundred subscribers across multiple calendars that's a non-trivial amount of repeated work that mostly returns identical bytes.

Add HTTP cache headers + object-cache backing:

- **HTTP headers:** `Cache-Control: max-age=...` (configurable, default ~15 min), `ETag` hashed from the rendered ICS body, `Last-Modified` from the most recent event save. Hook into `Setup::send_ics_headers()` which already owns the header writes.
- **Object cache:** wrap each per-event `get_ical_event_string()` output and each per-feed `get_ical_list()` aggregate under cache keys keyed by event ID / feed scope. Use `wp_cache_set` / `wp_cache_get` against the `gatherpress_calendar` group.
- **Bust on save:** hook `save_post_gatherpress_event` (and `deleted_post`, `updated_postmeta` for datetime / venue meta keys, `set_object_terms` for taxonomy changes) to invalidate the affected per-event keys and the feed aggregates the event participates in.
- **Conditional GET:** if request `If-None-Match` matches the current ETag, return `304 Not Modified` with an empty body. Same for `If-Modified-Since` vs `Last-Modified`.

- Page cache layer (Cloudflare, etc.) handles this opaquely if site owner configures it. Real, but not all sites have one — and per-event invalidation needs origin awareness.
- WordPress's built-in feed caching (`feed_cache_class` etc.) is for incoming feed fetches, not outgoing.

### Context

Surfaced as one of three operational gotchas during the #603 thread — see https://github.com/GatherPress/gatherpress/issues/603#issuecomment-4364756998.

### Code of Conduct

wppoland:
Fixes #1682 (headers, conditional GET, object cache and invalidation)

## Proposed changes:

* ICS responses now send `Cache-Control: public, max-age=…`, an `ETag` hashed from the body, and `Last-Modified`.
* A request whose `If-None-Match` or `If-Modified-Since` still matches gets `304 Not Modified` with no body. `If-None-Match` wins when both are present, per RFC 9110, and weak validators (`W/"…"`) and multi-tag headers are handled.
* The rendered body is kept in the object cache under `gatherpress_calendar`, for exactly the window the response advertises, so the HTTP layer and the object cache cannot disagree about freshness.
* New `Calendar\Cache` owns all of that, plus `gatherpress_calendar_max_age` (default 15 minutes, `0` disables caching entirely and restores the previous `no-store` headers).

**Invalidation is by version stamp rather than key deletion.** Every cache key carries the timestamp of the last calendar-relevant change, so one option write makes every stale entry unreachable, with no need to know which feeds a given event appeared in. That timestamp is also what `Last-Modified` reports, so a client's validator and the server's cache always agree. Post saves and deletions, `gatherpress_`-prefixed meta writes, and term changes all stamp it, for event-bearing and venue post types.

RSVP status changes deliberately do **not** stamp it: they arrive on the same `set_object_terms` hook, against a comment taxonomy, and do not alter a VEVENT. Without that guard every RSVP on the site would flush every calendar, which is the opposite of the point. There is a test for it.

The issue also lists per-event and per-feed cache keys. Those are an optimization on top of this rather than a correctness requirement, since the version stamp already bounds staleness, so I left them out to keep this reviewable.

### Other information:

12 tests in `Calendar\Test_Cache`: hooks, the version stamp seeding once and then holding still, `remember()` rendering once then serving the cache, rebuilding after a stamp, the max-age filter disabling caching, a negative max age clamping to zero, event and venue saves stamping while a regular post does not, GatherPress meta stamping while `_edit_lock` does not, and the RSVP-taxonomy and unknown-taxonomy guards.

Local runs: single-site 1659 tests, multisite 323, each with the same failures a clean `develop` produces in my environment (27 and 2, from the missing `gatherpress_events` table and `wp_`-prefix expectations). PHPCS and PHPStan clean.

## Testing instructions:

Against any published event, with `curl -D -`:

1. `GET /event/<slug>/ical/` now answers with `Cache-Control: public, max-age=900`, an `ETag`, and `Last-Modified`.
2. Repeat with `-H 'If-None-Match: "<that etag>"'`: `304 Not Modified`, no body.
3. Repeat with `W/` in front of the tag, and separately with a future `If-Modified-Since`: both `304`.
4. Send a wrong `If-None-Match`: `200` with the body, so a stale client still gets served.
5. Edit the event's date, then repeat step 1: the `ETag` and `Last-Modified` have both moved, the old tag revalidates as `200`, and the body carries the new `DTSTART`.
6. RSVP to an event and confirm the calendar `ETag` does not move.
7. `add_filter( 'gatherpress_calendar_max_age', '__return_zero' )` and confirm the response goes back to `no-store` and rebuilds every time.

Verified on WP 7.0.2 / PHP 8.2.29 with 0.35.0-beta.1:

```
200: Cache-Control: public, max-age=900   ETag: "bfb90b21…"   Last-Modified: Thu, 30 Jul 2026 11:21:27 GMT
If-None-Match (exact)  -> 304      If-None-Match (W/ weak) -> 304      If-Modified-Since future -> 304
If-None-Match (stale)  -> 200, 450 bytes
after editing the event: ETag "cfff9495…", Last-Modified 11:22:43, old tag -> 200, DTSTART:20260821T170000Z
```

`debug.log` stayed empty throughout.

Props motylanogha

### DIFF
--- a/includes/core/classes/calendar/class-cache.php
+++ b/includes/core/classes/calendar/class-cache.php
@@ -10,7 +10,7 @@
  * the responses are validated against.
  *
  * @package GatherPress\Core\Calendar
- * @since 0.35.0
+ * @since 0.36.0
  */
 
 namespace GatherPress\Core\Calendar;
@@ -31,7 +31,7 @@ use GatherPress\Core\Traits\Singleton;
  * against, so the HTTP layer and the object cache cannot disagree about how
  * fresh a response is.
  *
- * @since 0.35.0
+ * @since 0.36.0
  */
 final class Cache {
 
@@ -43,7 +43,7 @@ final class Cache {
 	/**
 	 * Object cache group for rendered calendar payloads.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
 	 *
 	 * @var string
 	 */
@@ -52,7 +52,7 @@ final class Cache {
 	/**
 	 * Option holding the GMT timestamp of the last calendar-relevant change.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
 	 *
 	 * @var string
 	 */
@@ -65,7 +65,7 @@ final class Cache {
 	 * use by default, so a subscriber's next poll is usually a conditional
 	 * request that costs a 304 rather than a full render.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
 	 *
 	 * @var int
 	 */
@@ -76,7 +76,7 @@ final class Cache {
 	 *
 	 * This method initializes the object and sets up necessary hooks.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
 	 */
 	protected function __construct() {
 		$this->setup_hooks();
@@ -87,7 +87,7 @@ final class Cache {
 	 *
 	 * This method adds hooks for different purposes as needed.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
 	 *
 	 * @return void
 	 */
@@ -106,7 +106,7 @@ final class Cache {
 	/**
 	 * Seconds a client may reuse a calendar response without revalidating.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
 	 *
 	 * @return int Max age in seconds. Zero disables client caching.
 	 */
@@ -118,7 +118,7 @@ final class Cache {
 		 * a rendered body is kept in the object cache, so the two cannot drift.
 		 * Return 0 to send `no-cache` and rebuild on every request.
 		 *
-		 * @since 0.35.0
+		 * @since 0.36.0
 		 *
 		 * @param int $max_age Seconds a calendar response stays fresh.
 		 *
@@ -134,7 +134,7 @@ final class Cache {
 	 * has a stable validator to hand out, rather than one that moves on every
 	 * request and defeats the point.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
 	 *
 	 * @return string Timestamp in `Y-m-d H:i:s` GMT.
 	 */
@@ -153,7 +153,7 @@ final class Cache {
 	/**
 	 * Stamp the calendar as changed, invalidating every cached response.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
 	 *
 	 * @return void
 	 */
@@ -164,7 +164,7 @@ final class Cache {
 	/**
 	 * Return a cached calendar payload, rendering it on a miss.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
 	 *
 	 * @param string   $key      Cache key, unique to the request's scope.
 	 * @param callable $renderer Builds the payload when the cache misses.
@@ -195,7 +195,7 @@ final class Cache {
 	/**
 	 * Namespace a cache key with the current calendar version stamp.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
 	 *
 	 * @param string $key Scope-specific key.
 	 *
@@ -208,7 +208,7 @@ final class Cache {
 	/**
 	 * Stamp the calendar when an event or venue post changes.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
 	 *
 	 * @param int|string $post_id The post that changed.
 	 *
@@ -227,7 +227,7 @@ final class Cache {
 	 * written without touching the post row, so `save_post` alone would miss
 	 * them.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
 	 *
 	 * @param int|string $meta_id  Meta row ID (unused; part of the hook signature).
 	 * @param int|string $post_id  The post the meta belongs to.
@@ -252,7 +252,7 @@ final class Cache {
 	 * Venue association and topics decide which feeds an event appears in, and
 	 * term writes do not touch the post row either.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
 	 *
 	 * @param int|string $object_id  The object whose terms changed.
 	 * @param array      $terms      Terms set (unused; part of the hook signature).
@@ -278,7 +278,7 @@ final class Cache {
 	/**
 	 * Whether a post type contributes to calendar output.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
 	 *
 	 * @param string $post_type The post type to test.
 	 *
--- a/includes/core/classes/calendar/class-cache.php
+++ b/includes/core/classes/calendar/class-cache.php
@@ -9,6 +9,11 @@
  * defines the `Cache` class, which holds the rendered bodies and the timestamp
  * the responses are validated against.
  *
+ * Bodies are stored as transients rather than in the object cache directly, so
+ * the win does not depend on the site having a persistent backend: with one,
+ * transients are object cache entries anyway; without one, they persist in the
+ * options table instead of evaporating at the end of the request.
+ *
  * @package GatherPress\Core\Calendar
  * @since 0.36.0
  */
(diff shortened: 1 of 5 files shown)


carstingaxion (changes requested):
I read through all the logic and found nothing strange or obviously wrong.

Besides the since tags, that do not match if this really goes into 0.36.0

Thanks for your work and the added tests @wppoland, I think we can get this in soon.

@mauteri, I like it.

mauteri:
Yes, the since tags will need to be changed from `0.35.0` to `0.36.0`. Thx!

wppoland:
Bumped the `@since` tags on the caching surface to `0.36.0`. Thanks both.

includes/core/classes/calendar/class-cache.php
```diff
@@ -0,0 +1,291 @@
...
+	 */
+	public function remember( string $key, callable $renderer ): string {
+		$max_age = $this->get_max_age();
+
+		if ( 0 === $max_age ) {
+			return (string) $renderer();
+		}
+
+		$versioned_key = $this->get_versioned_key( $key );
+		$cached        = wp_cache_get( $versioned_key, self::CACHE_GROUP );
```
mauteri: We tend to use transients instead of cache since most sites probably won't have persistent cache and won't benefit from this.

includes/core/classes/calendar/class-cache.php
```diff
@@ -0,0 +1,291 @@
...
+
+	/**
+	 * Stamp the calendar as changed, invalidating every cached response.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @return void
+	 */
+	public function touch(): void {
+		update_option( self::LAST_MODIFIED_OPTION, current_time( 'mysql', true ), false );
```
mauteri, line 173: Could we rename `touch()` to `mark_changed()`, and the three callbacks below to `mark_changed_for_post()` / `mark_changed_for_meta()` / `mark_changed_for_terms()`?

`touch` is Unix shorthand for updating a file's mtime and doesn't really carry over into PHP, so a reader has to open the method to learn what it does. Your own docblock already says it better than the name does:

> Stamp the calendar as changed, invalidating every cached response.

`mark_changed()` is that sentence.

One reason I'd favor it over `invalidate()` or `bump_last_modified()`: it names the fact being recorded rather than a storage mechanism, so it stays accurate whichever way the transients-vs-options thread lands.

The `_for_post` / `_for_meta` / `_for_terms` suffixes are good and worth keeping. It's only the stem I'd change.

includes/core/classes/calendar/class-cache.php
```diff
@@ -0,0 +1,291 @@
...
+	}
```
mauteri, line 174: Separate from the naming: I'd soften "invalidating every cached response", because that isn't quite what happens.

`get_versioned_key()` namespaces the key with `md5( $this->get_last_modified() )`, so a new stamp doesn't invalidate anything. It changes the namespace and subsequent reads look elsewhere. The previous entries are neither deleted nor flagged: they are orphaned, and linger in the object cache until their TTL expires or the backend evicts them.

That's a sensible tradeoff, since most object cache backends handle delete-by-pattern badly, but worth stating rather than implying the entries go away:

> Stamp the calendar as changed. Cached responses are namespaced by this stamp, so a new value moves every lookup to a fresh key and strands the old entries until they expire.

includes/core/classes/calendar/class-cache.php
```diff
@@ -0,0 +1,291 @@
...
+	 * Stamp the calendar when an event or venue post changes.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @param int|string $post_id The post that changed.
+	 *
+	 * @return void
+	 */
+	public function touch_for_post( $post_id ): void {
+		if ( $this->is_calendar_post_type( (string) get_post_type( (int) $post_id ) ) ) {
```
mauteri, line 234: `mark_changed_for_post()` here, if the rename above lands.

includes/core/classes/calendar/class-cache.php
```diff
@@ -0,0 +1,291 @@
...
+	 * @param int|string $meta_id  Meta row ID (unused; part of the hook signature).
+	 * @param int|string $post_id  The post the meta belongs to.
+	 * @param string     $meta_key The meta key that changed.
+	 *
+	 * @return void
+	 *
+	 * @SuppressWarnings(PHPMD.UnusedFormalParameter) Required by WP's *_post_meta signature.
+	 */
+	public function touch_for_meta( $meta_id, $post_id, $meta_key = '' ): void {
+		if (
```
mauteri, line 257: `mark_changed_for_meta()` here, if the rename above lands.

includes/core/classes/calendar/class-cache.php
```diff
@@ -0,0 +1,291 @@
...
+	 * @param array      $terms      Terms set (unused; part of the hook signature).
+	 * @param array      $tt_ids     Term taxonomy IDs (unused; part of the hook signature).
+	 * @param string     $taxonomy   The taxonomy that changed.
+	 *
+	 * @return void
+	 *
+	 * @SuppressWarnings(PHPMD.UnusedFormalParameter) Required by WP's set_object_terms signature.
+	 */
+	public function touch_for_terms( $object_id, $terms = array(), $tt_ids = array(), $taxonomy = '' ): void {
+		// Comment taxonomies share this hook: RSVP status changes do not alter
```
mauteri, line 283: `mark_changed_for_terms()` here, if the rename above lands.

includes/core/classes/calendar/class-cache.php
```diff
@@ -0,0 +1,291 @@
...
+		// Everything a VEVENT is built from: the post itself, its datetimes and
+		// venue meta, its terms, and its deletion. Each one stamps the calendar
+		// rather than reasoning about which feeds it belongs to.
+		add_action( 'save_post', array( $this, 'touch_for_post' ) );
+		add_action( 'deleted_post', array( $this, 'touch_for_post' ) );
+		add_action( 'updated_post_meta', array( $this, 'touch_for_meta' ), 10, 3 );
+		add_action( 'added_post_meta', array( $this, 'touch_for_meta' ), 10, 3 );
+		add_action( 'deleted_post_meta', array( $this, 'touch_for_meta' ), 10, 3 );
+		add_action( 'set_object_terms', array( $this, 'touch_for_terms' ), 10, 4 );
+	}
```
mauteri, line 113: If the rename lands, these read nicely as `mark_changed_for_post` / `mark_changed_for_meta` / `mark_changed_for_terms`: the callback name then states what the hook is for, rather than what it does to a timestamp.

Matching references live in `calendar/class-setup.php` and both test files, plus the `@covers` annotations, so it's a find-and-replace across four files.

mauteri (changes requested):

wppoland:
Both done in 8f77b8b7, thanks — the naming point in particular is right, and so is the correction underneath it.

**Transients.** Switched. Your reasoning holds in both directions: with a persistent backend a transient *is* an object cache entry, so nothing is lost, and without one it now persists in the options table instead of evaporating at the end of the request. The previous version quietly did nothing on exactly the hosts that need the help most.

One consequence worth stating rather than leaving implicit, since it interacts with your second point: the stranded entries are now option rows rather than object cache entries. They still carry `$max_age`, so WordPress's daily `delete_expired_transients()` collects them, but on a site with many distinct feed scopes and frequent edits there will be a window where expired-but-not-yet-collected rows sit in `wp_options`. I think that is the right trade for correctness on non-persistent hosts, and it is bounded by the max age rather than growing without limit. Flagging it so it is a decision rather than a surprise.

Because a transient name is an option name, the versioned key is now `md5( stamp . ':' . key )` behind the prefix rather than a concatenation, so a long scope key cannot push the name past the 172-character ceiling. There is a test asserting that with a deliberately long key.

**Naming.** `touch()` is now `mark_changed()`, and the three callbacks are `mark_changed_for_post()` / `mark_changed_for_meta()` / `mark_changed_for_terms()`. You are right that the docblock was already the better name. Renamed across the class, both test files and the `@covers` annotations; `calendar/class-setup.php` turned out not to reference them, so it was three files rather than four.

**The docblock claim.** Corrected, and I took your wording nearly verbatim because it is more accurate than mine was:

> Stamp the calendar as changed. Cached responses are namespaced by this stamp, so a new value moves every lookup to a fresh key and strands the old entries until they expire.

I also fixed the same overstatement in the class docblock and in the changelog entry, both of which said "invalidates".

**Verification.** `Test_Cache` passes locally, 15 tests including a new one asserting the payload really is readable back as a transient and that the name stays within the option-name limit. PHPCS clean on all three files. `Test_Setup::test_get_ical_list_and_feed` fails in my local environment, but it fails identically with this change stashed — it needs the `gatherpress_events` table that the non-wp-env setup does not create, so it is environmental rather than a regression. CI is the check that matters for it.

Props motylanogha

mauteri (approved):

PR #2068, opened 2026-07-30
Resolve the online-event term once in the editor

### LINKED ISSUE #2047: Add a public API for online events (setter, term helpers, REST parity)

wppoland:
Fixes #2047 (the JS half of part 2)

**Stacked on #2066** and branched from it, so this PR shows that commit too. Review or merge #2066 first; this one only makes sense once `gatherpress.config.onlineEventTermId` exists.

## Proposed changes:

The issue's table lists seven JS call sites carrying the `'online-event'` literal, four of which each ran their own `getEntityRecords()` query for the same term. They now share two helpers in `src/helpers/venue.js`:

* `isOnlineEventTermSlug( slug )` for the comparisons, mirroring the PHP predicate and deliberately mutually exclusive with being a venue term. Plus `ONLINE_EVENT_TERM_SLUG` for the one place that still needs the string.
* `getOnlineEventTermId( select, taxonomy )` for the id. It prefers the value PHP resolved and exposed, and keeps the taxonomy query as a fallback for a context where the `block_editor_settings_all` filter did not run, so nothing regresses in a custom editor setup.

Call sites updated: `blocks/online-event/edit.js`, `blocks/venue/edit.js`, `blocks/venue/helpers.js`, `components/OnlineEvent.js`, `components/VenueNavigator.js`, `components/VenueTermsCombobox.js`, and `hasOnlineEventTerm()` in `helpers/event.js`.

Two of those held the whole term object only to read `.id`, so they now hold the id itself, which is why the diff touches a few lines beyond the lookups.

### Other information:

6 tests added to `helpers/venue.test.js`: the predicate on the sentinel and on a venue slug, and the resolver preferring the setting (asserting the query is not called), falling back to the query, returning null with neither, and returning null when the query resolves nothing.

Full JS suite locally: 53 suites, 998 tests, all passing. PHP suite unchanged at 1651 tests with the same 27 pre-existing failures my environment produces on a clean `develop`. ESLint clean.

## Testing instructions:

1. Edit an event and open **Venue settings** in the sidebar.
2. Toggle **This is an online Event** on. The event picks up the online-event term; toggle it off and the term goes away.
3. Confirm `wp.data.select( 'core/editor' ).getEditorSettings().gatherpress.config.onlineEventTermId` matches the term id being written.
4. Pick a physical venue with the toggle on, and confirm the online term is preserved alongside it (`VenueNavigator` and `VenueTermsCombobox` both rely on the id for that).
5. Check the browser console is clean throughout.

Verified in the editor on WP 7.0.2 / PHP 8.2.29: `onlineEventTermId` came through as `2`, toggling on wrote `_gatherpress_venue: [ 2 ]`, toggling off left `[]`, and the console had no errors.

Props motylanogha

### DIFF
--- a/.github/changelog/2047-online-event-api
+++ b/.github/changelog/2047-online-event-api
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Added a public API for online events: `Event::is_online()` and `Event::set_online()`, which own the online-event term and the link meta together, plus `Venue\Setup::ONLINE_EVENT_TERM_SLUG`, `is_online_event_term_slug()` and `get_online_event_term_id()`. The resolved term ID is exposed to the block editor so components stop looking it up individually.
--- a/.github/changelog/2047-online-event-js-literals
+++ b/.github/changelog/2047-online-event-js-literals
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+The editor now resolves the online-event term once through `gatherpress.config.onlineEventTermId` instead of seven call sites repeating the slug, several of which each ran their own taxonomy query for the same term. Adds `isOnlineEventTermSlug()` and `getOnlineEventTermId()` to the venue helpers.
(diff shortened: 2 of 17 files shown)


wppoland:
Closing alongside #2066, in favour of #2049, which covers this JS half as well. Notes moved to #2049.

PR #2074, opened 2026-07-31
Stop marking a new event dirty before the author edits anything

### LINKED ISSUE #2054: New event is marked dirty before the author edits anything
A brand new event is marked as having unsaved changes before the author touches anything: opening `post-new.php?post_type=gatherpress_event`, waiting for the editor to settle, then navigating away triggers the browser's "Leave site?" prompt, and "Save draft" is active immediately. A new page or post on the same install does not do this, and an existing event opened and left alone does not either (`isEditedPostDirty()` is `false` there).

Verified on GatherPress 0.35.0-beta.1, WP 7.0.2, PHP 8.2.29, default theme, no other plugins. Not a 0.35 regression as far as I can tell, the code path is older.

### Cause

`src/components/DateTimeRange.js` writes the datetime meta from a mount effect:

```js
useEffect( () => {
	const payload = JSON.stringify( { ...dateTimeMetaData, dateTimeStart: ..., dateTimeEnd: ..., timezone } );
	editPost( { meta: { gatherpress_datetime: payload } } );
}, [ dateTimeStart, dateTimeEnd, timezone, dateTimeMetaData, editPost ] );
```

On an existing event the payload equals the stored meta, so the edit is a no-op and the post stays clean. On a new event the stored meta is empty while the store already holds the defaults (tomorrow 18:00, +2h, site timezone), so the first effect run is a real edit and the post is dirty from the start.

The write cannot simply be deferred: `Event\Setup::set_datetimes()` returns early when `gatherpress_datetime` is empty, so an event saved without that meta gets no row in the datetimes table.

### Options

1. Fill the default server side at save time: let `set_datetimes()` fall back to the same default (tomorrow 18:00 in the site timezone, +2h) when the meta is absent, then let the client effect skip its first run on a clean new post. No lost data, and the browser prompt only appears after a real edit.
2. Supply the default through `default_post_metadata` for `gatherpress_datetime` so the editor's baseline already matches the payload. Cheaper, but the saved payload then equals the baseline and Gutenberg may not send the field at all, which lands back on the empty-meta path in 1. It also makes events that legitimately have no datetime report one over REST.

Option 1 looks like the safe one, but it changes save semantics, so I did not want to send a PR without a maintainer opinion first. Happy to write it if you agree on the direction (and on the default belonging in PHP).

mauteri:
Fixes #2054

### Description of the Change

Opening `post-new.php?post_type=gatherpress_event` and navigating away triggers the browser's "Leave site?" prompt, and Save draft is active immediately. A new post or page doesn't do this, and neither does an existing event.

`DateTimeRange` writes `gatherpress_datetime` from a mount effect. On an existing event that payload equals the stored meta, so the edit is a no-op. On a new event the store already holds the defaults while the stored meta is empty, so the first run is a real edit and the post is dirty before anyone touches it.

The write can't simply be dropped. `Event\Setup::set_datetimes()` bails when the meta is empty, so an event saved without it gets **no row in the events table** and is invisible to every upcoming and past query while looking correct in the editor.

So both halves move, which is @wppoland's option 1 from the issue:

- **Client** — the effect skips while `isCleanNewPost()` is true. Any real edit clears that, and the effect runs from then on.
- **Server** — `set_datetimes()` seeds the editor's default (tomorrow 18:00 in the site timezone, running two hours) when nothing is stored.

### Two details in the seed that matter

**It doesn't write `gatherpress_datetime`.** An earlier version did, and it broke an existing test by clobbering meta set by callers who insert the post first and add meta second. The seed now only writes the derived row and leaves the editor-owned meta alone.

**It only applies when nothing is stored.** Guarded on `gatherpress_datetime_start`. Without that, every save with absent meta would recompute "tomorrow" and walk the event's date forward a day at a time.

### This makes datetime-less events unreachable

That's the intended outcome rather than a side effect, and it's a deliberate call by @mauteri: such an event was invisible to every event query, which is worse than the cosmetic bug being fixed here.

Three existing tests encoded the old contract, and all three now encode the new one:

- The empty-params case leaves `data_get_display_datetime`, since an event with no datetime can't occur.
- `test_get_datetime`'s new-event block asserts the seeded default rather than empties. The `new Event( 0 )` case is unchanged and still correct: no post means no hook.
- `test_get_event_counts_with_no_date` is reframed as `..._for_an_event_created_without_dates` and asserts it counts as upcoming. Reframed rather than deleted so it still catches a seeding regression.

The `—` fallback in `get_display_datetime()` is now unreachable through normal flows. Left in place as cover for a row deleted out from under us rather than expanding this PR.

### How to test the Change

1. Open `post-new.php?post_type=gatherpress_event`, let the editor settle, and navigate away. No "Leave site?" prompt, and Save draft is inactive.
2. Change the date, then navigate away. The prompt appears, as it should.
3. Create an event, set only a title, publish. It appears under Upcoming with tomorrow's date rather than vanishing from both views.
4. Confirm an existing event still saves its dates normally.

### Changelog Entry

> Fixed - Fixed a brand new event being marked as having unsaved changes before the author edits anything, which made the browser warn about leaving the page and enabled Save draft immediately.

### Credits

Props @wppoland for the report and the diagnosis, including the reason the client write can't just be deferred.

### Checklist:

`npm run lint:php`, `npm run lint:phpstan`, and `npx eslint` are clean. Full PHP suite: **OK (1648 tests, 6895 assertions)**.

### DIFF
--- a/.github/changelog/2054-new-event-dirty-state
+++ b/.github/changelog/2054-new-event-dirty-state
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed a brand new event being marked as having unsaved changes before the author edits anything, which made the browser warn about leaving the page and enabled Save draft immediately.
--- a/includes/core/classes/event/class-setup.php
+++ b/includes/core/classes/event/class-setup.php
@@ -14,6 +14,7 @@ namespace GatherPress\Core\Event;
 // Exit if accessed directly.
 defined( 'ABSPATH' ) || exit; // @codeCoverageIgnore
 
+use DateTimeImmutable;
 use Exception;
 use GatherPress\Core\Event;
 use GatherPress\Core\Feed;
@@ -745,10 +746,26 @@ final class Setup {
 		$data = get_post_meta( $post_id, 'gatherpress_datetime', true );
 
 		if ( empty( $data ) ) {
-			return;
-		}
+			// No meta means the editor never wrote one, which is the normal
+			// path for an event saved without touching the date controls
+			// (#2054), and for any programmatic insert that omits it. Without
+			// a fallback such an event gets no row in the events table and is
+			// invisible to every upcoming / past query while still looking
+			// correct in the editor.
+			//
+			// Seed the default only when nothing is stored yet. Recomputing it
+			// on every save would walk the event's date forward a day at a
+			// time, and the meta itself is deliberately left alone so a caller
+			// that inserts the post first and sets `gatherpress_datetime`
+			// afterwards is not overwritten.
+			if ( '' !== (string) get_post_meta( $post_id, 'gatherpress_datetime_start', true ) ) {
+				return;
+			}
 
-		$data = json_decode( (string) $data, true ) ?? array();
+			$data = $this->get_default_datetime();
+		} else {
+			$data = json_decode( (string) $data, true ) ?? array();
+		}
 
 		$event  = new Event( $post_id );
 		$params = array(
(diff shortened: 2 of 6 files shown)


mauteri:
@wppoland @carstingaxion mind reviewing this one? Thx!

PR #2081, opened 2026-08-01
docs: explain Advanced Query Loop event lists

### LINKED ISSUE #2037: Document building curated event lists with Advanced Query Loop
### What needs documenting

How to build a **curated list of specific events** (for example, three featured events for the quarter) using [Advanced Query Loop](https://wordpress.org/plugins/advanced-query-loop/) together with GatherPress.

### Background

This started as a proposal from @carstingaxion in #1151 to add a third option beside "Upcoming" and "Past" on the event query block, revealing a picker for selecting one or more specific events.

We decided not to build that natively. The dividing line we're working to:

- **GatherPress owns event-specific query semantics.** Upcoming/Past, include-unfinished, and order-by-datetime all require knowledge of the `gatherpress_events` table and datetime handling. Nothing else can express them, so they're ours.
- **AQL owns generic query features.** "Include these specific posts" has no event-specific logic in it at all. AQL's "Include Posts" control already does it, with title-or-ID lookup and a matching exclude control.

0.34.0 was a deliberate bet on extending AQL rather than competing with it (`src/integrations/aql/index.js` adds a GatherPress "Event Query Settings" panel to AQL blocks). Reimplementing Include Posts would be the first step back from that, and it invites the rest: taxonomy query, meta query, exclude current post. Each is defensible alone; together they're a second copy of AQL that we maintain forever.

So the answer for curated event lists is AQL, and that answer needs to live somewhere a user will find it.

### Scope

A user-facing doc covering two scenarios, both of which route through AQL.

**Scenario 1: a curated list of specific events**

1. Installing Advanced Query Loop.
2. Inserting an AQL Query Loop block and setting the post type to Events.
3. Where the GatherPress "Event Query Settings" panel appears and what it controls.
4. Using AQL's "Include Posts" control to hand-pick events.
5. The tradeoff worth calling out: the two control sets only meet on AQL's variation, so this path starts from the AQL block rather than the GatherPress event query variation.

**Scenario 2: mixing events with posts (or other post types) in one loop**

Split out of #1123, where the request was to have events show up in regular post listings. Core `WP_Query` has always accepted `post_type => array( 'post', 'gatherpress_event' )`; only the block UI withholds it, and AQL's "Additional Post Types" control exposes it.

6. Using AQL's "Additional Post Types" to combine events with posts.
7. Two limits that must be stated plainly, because neither is fixable by configuration:
   - **Ordering.** Events order by event datetime via the custom table join; posts order by publish date. One query, one `ORDER BY`. There is no setting that yields "event date for the events, publish date for the posts" in a single interleaved list.
   - **The event panel disappears.** Our integration gates on `usePostTypeSupports( 'gatherpress-event-date', query.postType )` (`src/integrations/aql/index.js`), so once a non-event post type joins the query the "Event Query Settings" panel stops rendering and upcoming/past is unavailable. Arguably correct, since those filters are meaningless for posts, but users need to know the mixed loop is a plain chronological list.

### Note

Point 5 above is a real seam and arguably a better issue than the original ask. The integration gates on `'advanced-query-loop' === props.attributes.namespace` (`src/integrations/aql/index.js`), so a user has to choose one variation or the other. Worth a separate ticket if we want to close that gap rather than document around it.

PuvaanRaaj:
Fix #2037 

## Summary
- document curated event lists using Advanced Query Loop
- explain where GatherPress event controls and AQL generic controls meet
- document ordering and filtering limits for mixed event and post lists

## Validation
- `npm run lint:md:docs` (run via the repository's installed wp-scripts binary; pass)
- `git diff --check`

Fixes #2037

### DIFF
--- a/docs/user/blocks/event-query.md
+++ b/docs/user/blocks/event-query.md
@@ -22,3 +22,67 @@ GatherPress includes a block-variation of the `core/query` block, called _Event 
 9. Allows to filter the queried events by Author, Keyword, Topic or Venue (and any other additionally registered taxonomies).
 10. **Filter by current venue** toggle — when placed on a venue page, the query is automatically scoped to that venue's events. The toggle is hidden on regular posts where there's no venue context to use. In templates or template parts, the toggle is shown with a note that the filter only takes effect when the template renders on a venue page.
 11. The variation is automatically loaded, when an editor chooses the „Event“ post type in a regular `core/query` block.
+
+## Curated and mixed event lists with Advanced Query Loop
+
+[Advanced Query Loop](https://wordpress.org/plugins/advanced-query-loop/) (AQL)
+adds generic query controls to the WordPress Query Loop block. GatherPress
+adds event-specific controls when an AQL query targets Events.
+
+Use AQL when you need to hand-pick events or combine events with other post
+types. Use the GatherPress Event Query variation when you only need the usual
+upcoming or past event lists.
+
+### Create a curated list of events
+
+Use this approach for a list such as featured events for a quarter.
+
+1. Install and activate Advanced Query Loop.
+2. Add an **Advanced Query Loop** block to the page or template.
+3. In AQL's query settings, set the post type to **Events**.
+4. In the **Event Query Settings** panel added by GatherPress, choose whether
+   the list contains upcoming or past events, configure inclusion of unfinished
+   events, and choose event ordering.
+5. In AQL's **Include Posts** control, search for and select the individual
+   events to display.
+6. Design the loop template as usual, for example with a title, Event Date,
+   and RSVP blocks.
+
+The **Include Posts** control belongs to AQL. It can find events by title or
+ID and can be paired with AQL's exclude control when needed.
+
+#### Why start with the AQL block
+
+GatherPress event query settings and AQL's generic controls appear together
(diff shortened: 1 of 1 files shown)


mauteri (approved):

mauteri:
@PuvaanRaaj thanks again for this one.

Heads up on credits: Props Bot has flagged your account as unlinked, which means we don't have a WordPress.org profile to attribute your work to. Our credits files store wp.org usernames and the release tooling validates each one against profiles.wordpress.org, so right now we have no way to include you.

We're cutting 0.35.0 shortly and would like you in the credits for it. Two ways to sort it:

- Link your GitHub and WordPress.org accounts, which is the durable fix and gets you props automatically on everything going forward: https://make.wordpress.org/core/2020/03/19/associating-github-accounts-with-wordpress-org-profiles/
- Or just reply here with your WordPress.org username and we'll add it manually for this release.

Either works. If you don't have a WordPress.org account yet, you can create one at https://login.wordpress.org/register.

PuvaanRaaj:
Hi @mauteri , i have already created wordpress account : puvaanraaj2001

mauteri:
@PuvaanRaaj great, i'll ensure you get credit in 0.35.0. :-)

PR #2121, opened 2026-08-09
Run screenshot generation in Playwright's own container

mauteri:
## Description

Browser downloads do not work on these runners. The transfer reaches **100% of 167 MB** and the process then stalls indefinitely.

| run | approach | result |
| --- | --- | --- |
| [31287296394](https://github.com/GatherPress/gatherpress/actions/runs/31287296394) | 10-minute cap | stalled mid-download |
| re-run | same | same |
| [31288339635](https://github.com/GatherPress/gatherpress/actions/runs/31288339635) | 30-minute cap | stalled **after** reaching 100% |
| [31313721827](https://github.com/GatherPress/gatherpress/actions/runs/31313721827) | 3 x 8-minute attempts | all three stalled identically |

That last one settles it. Three independent attempts failing at the same point means the failure is deterministic, so no timeout, retry or cache key can fix it.

The cache was only ever hiding this. The last green run, [28679615230](https://github.com/GatherPress/gatherpress/actions/runs/28679615230) on 3 July, never downloaded anything: it hit a warm cache. A dependabot bump rotated the key and exposed a problem that had been latent for months.

## The change

Run the job in `mcr.microsoft.com/playwright:v1.58.2-noble`, which ships the browsers and their system libraries preinstalled. The download, the apt phase and the browser cache all stop existing.

**Minus 95 lines.** Gone: the version-resolve step, the browser cache, the retry loop, `playwright install`, `playwright install-deps`, and the timeouts written to contain them. The job cap drops from 60 back to 30, since nothing slow is left.

## Keeping the tag in step

`v1.58.2-noble` must track `@playwright/test` in `package-lock.json` — Playwright refuses to run against a mismatched browser build. There is a comment above the image saying so. That is the one maintenance cost, and it is a visible one-line edit rather than a silent cache miss.

## Testing

YAML parses. `workflow_dispatch` runs from the default branch, so this can only be verified by merging and dispatching, same as every other attempt here. The difference is that this one removes the failing step rather than trying to survive it.

## Types of changes

CI fix.

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.

### DIFF
--- a/.github/workflows/screenshot-generation.yml
+++ b/.github/workflows/screenshot-generation.yml
@@ -86,11 +86,36 @@ jobs:
       uses: actions/setup-node@v4
       with:
         node-version-file: '.nvmrc'
+        # Enable built-in functionality for caching and restoring dependencies, which is disabled by default.
+        # The actions/setup-node uses actions/cache under the hood.
+        # https://github.com/actions/setup-node#caching-global-packages-data
         cache: 'npm'
 
+    # Restoring the short lived node_modules cache
+    # to be used across all workflows running on the last commit.
+    # https://github.com/actions/cache/blob/main/caching-strategies.md#creating-a-short-lived-cache
+    - uses: actions/cache/restore@v4
+      id: node_modules-cache
+      with:
+        path: |
+          ./node_modules
+        key: ${{ runner.os }}-node_modules-${{ github.sha }}-${{ hashFiles('package-lock.json') }}
+
     - name: NPM install
+      if: steps.node_modules-cache.outputs.cache-hit != 'true'
       run: npm ci
 
+    # Creating a short lived node_modules cache
+    - uses: actions/cache/save@v4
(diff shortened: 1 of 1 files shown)


mauteri:
Superseded by #2122. The container was aimed at the wrong cause and would have broken this job in four separate ways: the default shell drops from bash to sh (so the `[ x == y ]` locale check silently takes the wrong branch), and `jq`, `gh` and `fuser` are all absent from the image — plus the font set differs, which would have regenerated all eight locales and shipped a typeface change to wordpress.org.

The real cause is that `.nvmrc` pins Node 24.17.0 (#1962), and playwright-core's bundled yauzl deadlocks unpacking the browser archive on 24.16/24.17 (nodejs/node#63487, microsoft/playwright#41133). Fixed upstream only in Playwright 1.60.

PR #2124, opened 2026-08-09
Scope radio option focus to the field being edited

wppoland:
Found while checking GatherPress against the WordPress 7.1 field guide, specifically the line that plugins "relying on reaching across the editor document boundary should review their JavaScript and CSS for compatibility".

All 20 blocks are already `apiVersion: 3`, so the headline iframe item costs nothing. This is the one place that reaches across the boundary.

### The defect

`addRadioOption()` and `removeRadioOption()` locate the option to focus with a page-wide query against the global `document`:

```js
const radioOptionElements = document.querySelectorAll(
	'.gatherpress-radio-option .rich-text',
);
```

That is wrong in two independent ways.

**It does not see the canvas.** The options render inside the editor iframe, so they live in the iframe's document, not the admin one. The query returns an empty list and focus never moves: press Enter to add an option and the caret stays where it was. Under 7.0 this already affects anyone whose post editor is iframed. [WordPress 7.1 makes the post editor iframed unconditionally](https://make.wordpress.org/core/2026/08/03/iframed-editor-changes-in-wordpress-7-1/), so it stops being conditional.

**It ignores which block you are editing.** The selector matches every radio option in the post. With two radio fields in one form, adding an option to the first focuses the *last option of the last block*, and removing one moves the caret into the wrong field. That part is a plain bug today, iframe or not.

The same applies to `document.createRange()` and the bare `getSelection()` in the removal path, which operate on the admin document while the element belongs to the canvas.

### The change

The lookup now starts from the element that received the keystroke and is scoped to its own `.gatherpress-radio-group`, with the range and selection taken from `element.ownerDocument`. This is the `ownerDocument` approach the dev note recommends, and it fixes the cross-block case for free.

No markup changed, only the three handlers. That is deliberate: #2112 rewrites this component's JSX, and this PR does not touch a single line of it, so the two should merge in either order without conflicting.

### Tests

`test/unit/js/src/blocks/form-field/types/radio.test.js`, three cases:

- two radio fields on one page, Enter in the first, focus must land on the new option in *that* block
- the block rendered inside an iframe document, focus must resolve in the canvas
- removing an empty option moves focus back a step

The first two fail on `develop` and pass with this change. The third passes either way and is there to hold the refactored removal path.

```
Tests:       3 passed, 3 total
```

Full JS suite stays green (54 suites, 999 tests). `wp-scripts lint-js` clean. `src/blocks/*/types/**/*` sits in the coverage gate's exclude list, so these are not gate-mandated, they are the evidence.

I have not run this against a live 7.1 build, so the iframe half is verified by the jsdom test and the dev note rather than by hand in the editor. The cross-block half reproduces on any current version.

Part of #2098 (WordPress 7.1 readiness). Deliberately not `Fixes`, since that issue also needs the `Tested up to` bump, which is a release-time edit — see the full checklist audit in https://github.com/GatherPress/gatherpress/issues/2098#issuecomment-5268453334

### DIFF
--- a/src/blocks/form-field/types/radio.js
+++ b/src/blocks/form-field/types/radio.js
@@ -111,9 +111,15 @@ export default function RadioField( {
 				return;
 			}
 
-			// Move cursor to end of text.
+			// Move cursor to end of text. The canvas can be torn down inside
+			// the timer, which leaves the element without a view to select in.
 			const ownerDocument = element.ownerDocument;
-			const selection = ownerDocument.defaultView.getSelection();
+			const selection = ownerDocument.defaultView?.getSelection();
+
+			if ( ! selection ) {
+				return;
+			}
+
 			const range = ownerDocument.createRange();
 
 			range.selectNodeContents( element );
--- a/src/blocks/form-field/types/radio.js
+++ b/src/blocks/form-field/types/radio.js
@@ -81,23 +81,56 @@ export default function RadioField( {
 		}
 	};
 
-	const addRadioOption = () => {
-		const newOptions = [ ...radioOptions, { label: '', value: '', id: uuidv4() } ];
-		setAttributes( { radioOptions: newOptions } );
+	// Moves focus to one option's editable label once React has re-rendered
+	// the group, optionally dropping the caret at the end of the text.
+	//
+	// The lookup starts from an element inside the group rather than from the
+	// global `document` for two reasons. The editor canvas is an iframe, so the
+	// options live in that iframe's document and a top-level `document` query
+	// matches nothing. Scoping to this block's own group also stops a second
+	// radio field on the same post from swallowing the focus.
+	const focusOption = ( scopeElement, targetIndex, placeCaretAtEnd = false ) => {
+		const group = scopeElement?.closest( '.gatherpress-radio-group' );
+
+		if ( ! group ) {
+			return;
+		}
 
 		setTimeout( () => {
-			const radioOptionElements = document.querySelectorAll(
+			const element = group.querySelectorAll(
 				'.gatherpress-radio-option .rich-text',
-			);
-			const lastOption =
-				radioOptionElements[ radioOptionElements.length - 1 ];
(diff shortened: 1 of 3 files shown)


src/blocks/form-field/types/radio.js
```diff
@@ -81,23 +81,56 @@ export default function RadioField( {
...
+
+			element.focus();
+
+			if ( ! placeCaretAtEnd ) {
+				return;
+			}
+
+			// Move cursor to end of text.
+			const ownerDocument = element.ownerDocument;
+			const selection = ownerDocument.defaultView.getSelection();
```
mattcowan: One low-severity finding: `ownerDocument.defaultView.getSelection()` is dereferenced inside the 50 ms timer with no guard. If the canvas is torn down within that window (switch to Code Editor, device-preview toggle), defaultView is null and the timer throws. The pre-diff code used the global getSelection(), which couldn't throw, so this is a genuinely new failure mode. Suggested fix is one character plus a return:

```
const selection = ownerDocument.defaultView?.getSelection();

if ( ! selection ) {
	return;
}
```

mattcowan (commented):
Other than the one nitpick comment on radio.js, this looks good!

src/blocks/form-field/types/radio.js
mauteri: Good catch — taking the guard.

One wrinkle worth recording, because it changes what the test can assert. The obvious regression test is to detach the iframe mid-timer, but jsdom does not model a discarded browsing context the way a browser does: after `iframe.remove()` the document keeps a live `defaultView`, and the throw comes from inside jsdom's own `window.getSelection` (`Window.js:920`) rather than from dereferencing null. So that test fails whether or not the guard is present, and proves nothing about it.

The added case nulls `defaultView` directly instead, which is what a browser leaves behind. It fails without the guard and passes with it.

`c74f701a`. Full JS suite green — 55 suites, 1002 tests.

Issue #2127, opened 2026-08-09
Expose events and RSVP counts through the WordPress Abilities API


wppoland:
### What

Register a small set of read-only GatherPress abilities with the [WordPress Abilities API](https://make.wordpress.org/core/2026/07/29/new-execution-lifecycle-filters-for-the-abilities-api-in-wordpress-7-1/), so anything that speaks that API can ask a site what is coming up and how many people replied.

Proposed surface, in a `gatherpress` ability category:

| Ability | Input | Returns |
|---|---|---|
| `gatherpress/get-upcoming-events` | `count` (1-50, default 5) | id, title, url, start, end, timezone |
| `gatherpress/get-event-rsvp-counts` | `event_id` | response counts keyed by status, plus `all` |

Both are annotated `readonly` + `idempotent`, both are permission checked, and neither writes anything.

### Why

Today the only way for an outside client to answer "what is this group's next event" is to know that GatherPress uses the `gatherpress_event` post type, that dates live in prefixed post meta rather than `post_date`, and that RSVPs are comments with a `gatherpress_rsvp` comment type. That is a lot of internal structure to depend on, and all of it is free to change.

An ability is a stable contract over that. It also puts GatherPress in front of the tooling being built on this API in the 7.1 cycle: WP-CLI can list abilities, and AI/MCP clients discover them automatically. For an events plugin, "what's on" and "how many are coming" are exactly the questions such a client asks.

This came out of checking GatherPress against the 7.1 field guide after Carsten shared the pre-release checklist in Slack. The audit result itself was good news, all 20 blocks are already `apiVersion: 3` and editor CSS already reaches the canvas via `enqueue_block_assets`, so this is the opportunity rather than the repair. (The one real defect that audit turned up is separate, #2124.)

### Compatibility

No version gate needed. `wp_register_ability()` is `@since 6.9.0` and GatherPress requires 7.0, so it is always present.

One trap worth recording: the `meta.public` flag is **new in 7.1** and does not exist in 7.0. Registering with `public` alone would silently expose nothing for anyone on 7.0. The proposal uses `meta.show_in_rest`, which exists in both.

### Status

Implemented with full test coverage in the PR linked below, so this is a concrete proposal rather than a request for someone else to build it. Entirely happy for the answer to be no, or for the shape of the abilities to change first, which is why this issue exists before review starts.

Open questions I would rather you decide than guess:

[...]

PR #2128, opened 2026-08-09
Expose events and RSVP counts through the WordPress Abilities API

wppoland:
Implements #2127. Please read that issue first, it carries the rationale and the two open questions I would rather you answered than guessed at.

### What this adds

A `gatherpress` ability category and two read-only abilities:

| Ability | Input | Returns |
|---|---|---|
| `gatherpress/get-upcoming-events` | `count` (1-50, default 5) | `id`, `title`, `url`, `start`, `end`, `timezone` |
| `gatherpress/get-rsvp-counts` | `post_id` | counts keyed by RSVP status, plus `all` |

Both carry input and output JSON Schema, and are annotated `readonly` + `idempotent` + non-destructive. Neither writes anything.

They live in two domain classes, `includes/core/classes/event/class-abilities.php` and `includes/core/classes/rsvp/class-abilities.php`, each instantiated by its own subsystem loader. RSVP is its own post type support, so its ability answers for any post type declaring `gatherpress-rsvp`, event or not.

`get-upcoming-events` delegates to `Event\Query::get_upcoming_events()` and `get-rsvp-counts` to `Rsvp::responses()`, so the abilities stay a thin contract over logic that already exists and is already tested.

### Who this is for

1. **A visitor, through an assistant.** "What's coming up at this meetup?" A chat client or site assistant calls `get-upcoming-events` and gets titles, permalinks, and start/end in GMT plus the event timezone. Today that answer needs either scraping the archive or knowing GatherPress keeps datetimes in a custom table rather than post meta, which is exactly the coupling the Abilities API exists to remove. Works logged out.
2. **An organizer, through WP-CLI or an admin-side assistant.** "How many people are coming on Saturday?" `get-rsvp-counts` returns the breakdown by status plus an `all` total, without the caller needing to know RSVPs are comments with a `gatherpress_rsvp` comment type and a status taxonomy.
3. **A site owner wiring up something small.** The two composed, list what is upcoming then ask for counts on each, to build a digest email or a dashboard widget through a documented read-only interface rather than internal classes.

### Manual testing

Both abilities are read-only, so the run endpoint takes `GET` with the payload under an `input` parameter.

```bash
curl -s 'https://example.test/wp-json/wp-abilities/v1/abilities/gatherpress/get-upcoming-events/run?input[count]=3'
curl -s 'https://example.test/wp-json/wp-abilities/v1/abilities/gatherpress/get-rsvp-counts/run?input[post_id]=123'
```

Four checks worth running, all four verified against a live site rather than only in the unit tests:

1. **Public listing.** Logged out, `get-upcoming-events`. Returns the same events the archive shows, soonest first.
2. **The count clamp.** `input[count]=0` and `input[count]=999` are clamped to 1 and 50 before reaching the query.
3. **The password gate.** Give a published event a post password. Anonymous gets `401 rest_ability_cannot_execute`; the author or an admin gets `200` with the breakdown.
4. **RSVP support, not event support.** `add_post_type_support( 'page', 'gatherpress-rsvp' )`, then ask for counts on a published page: `200`. Remove the support and ask again: `401`.

### Compatibility, which is the part worth reviewing closely

**No version gate is needed.** `wp_register_ability()` is `@since 6.9.0` and the plugin requires 7.0, so it is always available. I checked the 6.9.0 and 7.0.0 tags rather than trusting the docs.

**Registration uses `meta.show_in_rest`, not `meta.public`.** This matters: `public` is new in 7.1 and does not exist in 7.0. Registering with `public` alone would silently expose nothing to anyone still on 7.0. `show_in_rest` is present in both, and under 7.1 `public` merely seeds it. There is a comment in the source saying so, because it is the kind of thing that gets "simplified" later.

### Permissions

- **`get-upcoming-events` is public.** The listing is the event archive in another shape: it runs an ordinary `WP_Query`, so WordPress already limits it to what the current viewer may read, and an anonymous caller gets exactly the published events the archive would show them. Gating it behind `read` locked out those visitors for nothing. The Abilities API will not let the callback be dropped outright, `WP_Ability::__construct()` requires a callable `permission_callback`, so it is `__return_true`.
- **`get-rsvp-counts` is gated**, by `Event::can_read_rsvps()`, the same rule the roster REST endpoint uses. That rejects any post type without `gatherpress-rsvp` support, keeps a published post's responses behind its password gate, and requires `read_post` for anything unpublished. It never looks at `gatherpress-event-date`.

The RSVP callback also refuses a zero ID rather than constructing an `Rsvp` around a post that does not exist. The permission callback already covers that, but the ability is a public entry point and should not depend on callers reaching it in the right order.

### Testing

`test/unit/php/includes/tests/core/classes/event/class-test-abilities.php` and `.../rsvp/class-test-abilities.php`, 19 tests between them.

```
OK (19 tests, 49 assertions)
```

Coverage on both new classes is **100%**, 74/74 and 57/57 statements, verified from the clover report rather than assumed. Whole suite stays green, single site and multisite:

```
OK (1727 tests, 7110 assertions)
OK (323 tests, 856 assertions)
```

PHPCS, PHPStan and PHPMD all clean.

Two notes on the tests, since both are the kind of thing a reviewer would otherwise have to work out:

- `test_register_abilities` unregisters before re-firing `wp_abilities_api_init`. The plugin already registered during boot, so firing the action over a populated registry is a duplicate registration and the Abilities API correctly reports it as incorrect usage. Unregistering first keeps the registration code genuinely under test instead of asserting on what boot happened to leave behind. Both classes answer the same action, so both abilities are cleared before it fires.
- The two classes share the `gatherpress` ability category, and either subsystem can be the one that boots first, so each registers it and whichever runs second stands down via `wp_has_ability_category()`. That keeps either class loadable on its own. The alternative is a category per subsystem; say the word if you would rather have that.
- The `do_action()` calls carry a targeted `phpcs:ignore` for `PrefixAllGlobals.NonPrefixedHooknameFound`. Those are core hook names, not GatherPress ones, so the sniff is a false positive here.

### Not included, on purpose

Past events, venue detail, and anything that writes an RSVP. Write abilities deserve their own discussion about annotations and permissions, and I did not want to bundle that decision into a first pass.

`readme.txt` still says `Tested up to: 7.0`. Worth bumping once 7.1 ships on 19 August, but that is a release-manager call, so it is not in this PR.


includes/core/classes/class-abilities.php
```diff
@@ -0,0 +1,285 @@
...
+	 */
+	public function can_read_event( $input = null ): bool {
+		$post_id = is_array( $input ) ? (int) ( $input['event_id'] ?? 0 ) : 0;
+		$post    = $post_id ? get_post( $post_id ) : null;
+
+		if ( ! $post || ! post_type_supports( $post->post_type, 'gatherpress-event-date' ) ) {
+			return false;
+		}
+
+		return current_user_can( 'read_post', $post_id );
```
mauteri: Let's set the capability as a constant on the class. [Example](https://github.com/GatherPress/gatherpress/blob/develop/includes/core/classes/rsvp/class-rsvp.php#L48).

includes/core/classes/class-abilities.php
```diff
@@ -0,0 +1,285 @@
...
+
+	/**
+	 * Whether the current user may read site content.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @return bool True when the current user can read.
+	 */
+	public function can_read(): bool {
+		return current_user_can( 'read' );
```
mauteri: Let's set the capability as a constant on the class. [Example](https://github.com/GatherPress/gatherpress/blob/develop/includes/core/classes/rsvp/class-rsvp.php#L48).

mauteri (changes requested):
@jmarx can you review this one too since you've done a bit on the Abilities API and GatherPress in the past?

mauteri:
@wppoland Also thinking maybe we don't create a generic Abilities class and break it up into `includes/core/classes/rsvp/class-abilities.php` and `includes/core/classes/event/class-abilities.php`

includes/core/classes/class-abilities.php
```diff
@@ -0,0 +1,285 @@
...
+				'execute_callback'    => array( $this, 'get_upcoming_events' ),
+				'permission_callback' => array( $this, 'can_read' ),
+				'meta'                => array(
+					'annotations'  => array(
+						'readonly'    => true,
+						'destructive' => false,
+						'idempotent'  => true,
+					),
+					// `show_in_rest` rather than the `public` flag: `public` is new in
+					// WordPress 7.1 and GatherPress still supports 7.0, where it is ignored.
```
mauteri: I don't think we need this comment

mauteri (changes requested):

includes/core/classes/class-abilities.php
```diff
@@ -0,0 +1,285 @@
...
+	 * @since 0.36.0
+	 *
+	 * @param mixed $input Input passed to the ability.
+	 *
+	 * @return int[] Response counts keyed by status, including an `all` total.
+	 */
+	public function get_event_rsvp_counts( $input = null ): array {
+		$post_id = is_array( $input ) ? (int) ( $input['event_id'] ?? 0 ) : 0;
+		$counts  = array();
+
```
mauteri: You need a `post_type_supports` check for `gatherpress-rsvp` here.

mauteri (changes requested):

includes/core/classes/class-abilities.php
```diff
@@ -0,0 +1,285 @@
...
+							'description' => __( 'The event to count responses for.', 'gatherpress' ),
+						),
+					),
+					'required'   => array( 'event_id' ),
+				),
+				'output_schema'       => array(
+					'type'                 => 'object',
+					'additionalProperties' => array( 'type' => 'integer' ),
+				),
+				'execute_callback'    => array( $this, 'get_event_rsvp_counts' ),
```
mauteri: I'd call this `get_rsvp_counts` since RSVP doesn't have to be tightly bound to events since it has its own post type support.

includes/core/classes/class-abilities.php
```diff
@@ -0,0 +1,285 @@
...
+				'permission_callback' => array( $this, 'can_read_event' ),
```
mauteri: You have a `gatherpress-event-date` support check on this which can be problematic if they post type that has RSVP doesn't have that support. Should have a different permission callback for this that is different from event. It's good to think of RSVP and Event as 2 supports and not the same thing.

mauteri (changes requested):

jmarx:
@wppoland Can you give me a couple specific use cases you had in mind for this?  Is this aimed at admins/organizers, members, or anonymous visitors?

jmarx:
@wppoland To be more specific, How would you reproduce this with testing instructions besides running the unit tests?

includes/core/classes/class-abilities.php
```diff
@@ -0,0 +1,285 @@
...
+							'title'    => array( 'type' => 'string' ),
+							'url'      => array( 'type' => 'string' ),
+							'start'    => array( 'type' => 'string' ),
+							'end'      => array( 'type' => 'string' ),
+							'timezone' => array( 'type' => 'string' ),
+						),
+					),
+				),
+				'execute_callback'    => array( $this, 'get_upcoming_events' ),
+				'permission_callback' => array( $this, 'can_read' ),
```
jmarx: I am not sure we need a permission callback on this one since get upcoming events should be viewable by anyone who visits the site?

jmarx (changes requested):
@wppoland let me know what you think about that first comment. I'll have some more later. But I am also super interested to know where/how you think this will be used(I asked a couple of questions down below).

wppoland:
@mauteri Done, the generic class is gone. `includes/core/classes/class-abilities.php` is replaced by `includes/core/classes/event/class-abilities.php` and `includes/core/classes/rsvp/class-abilities.php`, each instantiated by its own subsystem loader (`Event\Setup::instantiate_classes()` and `Rsvp\Setup::instantiate_classes()`) rather than by `Core\Setup`.

Point by point:

**RSVP and Event as two supports.** `gatherpress/get-event-rsvp-counts` is now `gatherpress/get-rsvp-counts`, and its input is `post_id` rather than `event_id`, since nothing about the count needs the post to be an event. Its permission callback no longer looks at `gatherpress-event-date` at all. There is a test that registers a post type declaring only `gatherpress-rsvp` and asserts the counts are readable for it.

**The `post_type_supports( 'gatherpress-rsvp' )` check.** It comes in through `Event::can_read_rsvps()`, which already rejects any post type without that support and is the same rule `Event\Rest_Api` uses for the roster. Delegating keeps one definition of who may see responses instead of a second one that can drift. Worth flagging that the helper is named for Event but the rule inside it is not event-specific, so if you would rather it moved somewhere neutral I am happy to do that in this PR.

**Capability constants.** Both bare capability strings are gone. `current_user_can( 'read_post', … )` was the one in the RSVP path and it is now `Event::can_read_rsvps()`, which uses `Event::READ_CAPABILITY` and `Event::EDIT_CAPABILITY` internally. `current_user_can( 'read' )` was the one on the events listing and it is gone entirely, see the note on @jmarx's comment below. So there is no capability left in these classes to hoist into a constant. If you would still rather see one declared, say the word and I will add it.

**The `show_in_rest` comment.** Removed.

One thing worth a second opinion: both classes need the shared `gatherpress` ability category, so each registers it and whichever runs second stands down via `wp_has_ability_category()`. That keeps either class loadable on its own, at the cost of the same three lines in two files. The alternative is a category per subsystem. Tell me which you prefer.

Full suite 1727 green, multisite 323 green, phpcs, phpstan and phpmd clean, and both new classes are at 100% line coverage.

wppoland:
@jmarx Thanks, and yes, you were right about that first one.

**On the permission callback for `get-upcoming-events`.** I agreed and removed the gate. It was `current_user_can( 'read' )`, which is wrong twice: it locks out anonymous visitors who can already see the same events on the archive page, and on a site that allows subscriber registration it is not much of a gate anyway. The listing runs an ordinary `WP_Query`, so WordPress already limits it to what the current viewer may read, and an anonymous caller gets exactly the published events the events archive would show them. The callback is now `__return_true`, with a test asserting a logged-out visitor passes it.

Note the Abilities API will not let the callback be dropped outright, `WP_Ability::__construct()` requires a callable `permission_callback`, so `__return_true` is the way to say "public".

The RSVP counts ability is the opposite case and stays gated. It defers to `Event::can_read_rsvps()`, the same rule the roster REST endpoint uses, so a password-protected event withholds its counts from anyone who has not satisfied the password, and a non-published post needs `read_post`.

**Use cases.** Three, in the order I care about them:

1. *Anonymous or logged-in visitor, through an assistant.* "What's coming up at this meetup?" A chat client or site assistant calls `gatherpress/get-upcoming-events` and gets titles, permalinks, and start/end in GMT plus the event timezone. Today that answer requires either scraping the archive or knowing that GatherPress keeps datetimes in a custom table rather than post meta, which is exactly the coupling the Abilities API exists to remove.

2. *Organizer, through WP-CLI or an admin-side assistant.* "How many people are coming to Saturday's event?" `gatherpress/get-rsvp-counts` returns the breakdown by status plus an `all` total, without the caller needing to know RSVPs are comments with a `gatherpress_rsvp` comment type and a status taxonomy.

3. *Site owner wiring up something small.* The two composed: list what is upcoming, then ask for counts on each, to build a digest email or a dashboard widget, entirely through a documented read-only interface rather than internal classes.

What it is deliberately not: there is no write ability here. Nothing checks anyone in, changes a response, or edits an event. Both abilities are annotated `readonly`, `idempotent`, and non-destructive, so a client can call them speculatively.

**Manual testing, no unit tests involved.** Set up first: create at least one published event with a future date, and RSVP to it as a couple of users.

*Over REST, which is the path most clients take.* Both abilities are annotated read-only, so the run endpoint requires `GET` and takes the payload under an `input` parameter rather than a JSON body.

```bash
# Anonymous. Returns the upcoming events.
curl -s 'https://example.test/wp-json/wp-abilities/v1/abilities/gatherpress/get-upcoming-events/run?input[count]=3'

# Anonymous, counts for a public event.
curl -s 'https://example.test/wp-json/wp-abilities/v1/abilities/gatherpress/get-rsvp-counts/run?input[post_id]=123'

# Confirm both show up in the listing at all.
curl -s 'https://example.test/wp-json/wp-abilities/v1/abilities' | grep gatherpress
```

*Or in PHP, if you would rather skip the auth plumbing.*

```php
$request = new WP_REST_Request( 'GET', '/wp-abilities/v1/abilities/gatherpress/get-rsvp-counts/run' );
$request->set_query_params( array( 'input' => array( 'post_id' => 123 ) ) );
print_r( rest_do_request( $request )->get_data() );
```

*The four checks I would actually run:*

1. **Public listing.** Logged out, run `get-upcoming-events`. Expect the same events the events archive shows, soonest first, with `id`, `title`, `url`, `start`, `end`, `timezone`. On my run: `200`, one event, `{"id":175,"title":"Manual Check Meetup","start":"2026-08-19 14:23:43","timezone":"America/New_York", …}`.
2. **The count clamp.** Pass `input[count]=0` and `input[count]=999`. Neither should reach the query as-is, the value is clamped to 1 and 50.
3. **The password gate.** Give a published event a post password. Logged out, `get-rsvp-counts` for it should be refused. Log in as the event's author or an admin and it should succeed. This is the case that made me change the callback in 46647f03. On my run: `401 rest_ability_cannot_execute` anonymous, `200` with the breakdown as admin.
4. **RSVP support, not event support.** Add `gatherpress-rsvp` support to a post type that has no `gatherpress-event-date` support:
   ```php
   add_action( 'init', function () { add_post_type_support( 'page', 'gatherpress-rsvp' ); } );
   ```
   `get-rsvp-counts` should work for a published page. Then remove the support and try again, it should be refused. That is the split @mauteri asked for, and it now holds. On my run: `200` with the breakdown while the support is on, `401` once it is removed.

I walked all six of those against a live site rather than only the unit tests, and they behave as described.

Happy to fold any of this into the PR description or into `docs/` if you would rather it lived somewhere permanent.

includes/core/classes/class-abilities.php
wppoland: Resolved by removing the capability rather than hoisting it. This was `current_user_can( 'read' )` on the upcoming events listing, and per @jmarx's comment below the listing should not be gated at all, so the callback is now `__return_true`. Nothing left here to make a constant. Details in the summary comment.

includes/core/classes/class-abilities.php
wppoland: Resolved by delegation. This was `current_user_can( 'read_post', $post_id )` in the RSVP path, now `Event::can_read_rsvps()`, which uses `Event::READ_CAPABILITY` and `Event::EDIT_CAPABILITY` internally and is the same rule the roster REST endpoint follows. So there is no bare capability string left in either class. Happy to declare one explicitly if you would still prefer it.

includes/core/classes/class-abilities.php
wppoland: Removed.

includes/core/classes/class-abilities.php
wppoland: Done. The RSVP ability now gates on `Event::can_read_rsvps()`, whose first check is `post_type_supports( $post->post_type, 'gatherpress-rsvp' )`. There is a test registering a post type that declares only `gatherpress-rsvp` and asserting the counts are readable for it, plus one asserting a post type with neither support is refused.

includes/core/classes/class-abilities.php
wppoland: Renamed. The ability is `gatherpress/get-rsvp-counts`, the method is `get_rsvp_counts()`, and the input is `post_id` rather than `event_id` since nothing about the count requires the post to be an event. It now lives in `includes/core/classes/rsvp/class-abilities.php`.

includes/core/classes/class-abilities.php
wppoland: Agreed, and split. The RSVP ability has its own permission callback that never looks at `gatherpress-event-date`; it keys purely on `gatherpress-rsvp` support. The two abilities are now in separate classes under `event/` and `rsvp/`, each loaded by its own subsystem Setup, so they are genuinely two supports rather than one.

includes/core/classes/class-abilities.php
wppoland: Agreed, removed. It was `current_user_can( 'read' )`, which locked out the anonymous visitors who can already see the same events on the archive. The listing runs an ordinary `WP_Query`, so WordPress already scopes it to what the viewer may read. Note the API will not let the callback be dropped outright (`WP_Ability::__construct()` requires a callable `permission_callback`), so it is `__return_true`, with a test asserting a logged-out visitor passes. Use cases and manual test steps are in the comment below.

wppoland:
Closing out the two things I left open, so neither of you has to adjudicate them. Both are easy to reverse if you disagree.

**1. The shared ability category stays shared.** Both classes register `gatherpress` and whichever runs second stands down via `wp_has_ability_category()`. I considered a category per subsystem, which would drop the guard and the thirteen duplicated lines, and decided against it: a category is a user-facing grouping, not a reflection of which PHP class did the registering. The abilities are already namespaced `gatherpress/…`, a client browsing categories should see one GatherPress group rather than two, and if venue or topic abilities land later that approach would give one plugin four categories. Splitting it would be letting file layout drive public API to save duplication that is cheap and self-documenting. The guard is what keeps each class independently loadable, which was the point of the split.

**2. No capability constant, because there is no capability left to name.** Both bare strings are gone rather than hoisted. `current_user_can( 'read' )` on the events listing went away entirely once the listing became public, and `current_user_can( 'read_post', … )` in the RSVP path is now `Event::can_read_rsvps()`, which uses `Event::READ_CAPABILITY` and `Event::EDIT_CAPABILITY` internally. Declaring a constant nothing reads would be worse than declaring none. @mauteri, if you would still rather see one, say so and I will add it.

**One thing I got wrong in the last push, now fixed in 6d1e915.** I had added a `Rsvp\Abilities::SUPPORT = 'gatherpress-rsvp'` constant, and it was never read. The support check lives inside `Event::can_read_rsvps()`, and the rest of the codebase spells that string as a literal in about eight places, so a constant in this one class was both dead and inconsistent. Removed, and the test now uses the literal like everything else.

Still green after it: 1727 single site, 323 multisite, 100% coverage on both new classes, phpcs, phpstan and phpmd clean.

**The one question genuinely worth your call**, restated because it is easy to lose in the thread: `Event::can_read_rsvps()` is named for Event but the rule inside it is not event-specific, it checks `gatherpress-rsvp` support and nothing about event dates. So `Rsvp\Abilities` reaching for an `Event\Event` method is the one place the two supports still touch. Leaving it there means one definition of who may read responses, shared with the roster REST endpoint, which I think is worth more than the tidier dependency graph. Moving it to `Rsvp` with a delegating alias on `Event` would finish the decoupling but changes a public method shipped in 0.35.1, so it is your call, not mine. Happy to do it in this PR or a follow-up.

jmarx:
> @jmarx Thanks, and yes, you were right about that first one.
> 
> **On the permission callback for `get-upcoming-events`.** I agreed and removed the gate. It was `current_user_can( 'read' )`, which is wrong twice: it locks out anonymous visitors who can already see the same events on the archive page, and on a site that allows subscriber registration it is not much of a gate anyway. The listing runs an ordinary `WP_Query`, so WordPress already limits it to what the current viewer may read, and an anonymous caller gets exactly the published events the events archive would show them. The callback is now `__return_true`, with a test asserting a logged-out visitor passes it.
> 
> Note the Abilities API will not let the callback be dropped outright, `WP_Ability::__construct()` requires a callable `permission_callback`, so `__return_true` is the way to say "public".
> 
> The RSVP counts ability is the opposite case and stays gated. It defers to `Event::can_read_rsvps()`, the same rule the roster REST endpoint uses, so a password-protected event withholds its counts from anyone who has not satisfied the password, and a non-published post needs `read_post`.
> 
> **Use cases.** Three, in the order I care about them:
> 
> 1. _Anonymous or logged-in visitor, through an assistant._ "What's coming up at this meetup?" A chat client or site assistant calls `gatherpress/get-upcoming-events` and gets titles, permalinks, and start/end in GMT plus the event timezone. Today that answer requires either scraping the archive or knowing that GatherPress keeps datetimes in a custom table rather than post meta, which is exactly the coupling the Abilities API exists to remove.
> 2. _Organizer, through WP-CLI or an admin-side assistant._ "How many people are coming to Saturday's event?" `gatherpress/get-rsvp-counts` returns the breakdown by status plus an `all` total, without the caller needing to know RSVPs are comments with a `gatherpress_rsvp` comment type and a status taxonomy.
> 3. _Site owner wiring up something small._ The two composed: list what is upcoming, then ask for counts on each, to build a digest email or a dashboard widget, entirely through a documented read-only interface rather than internal classes.
> 
> What it is deliberately not: there is no write ability here. Nothing checks anyone in, changes a response, or edits an event. Both abilities are annotated `readonly`, `idempotent`, and non-destructive, so a client can call them speculatively.
> 
> **Manual testing, no unit tests involved.** Set up first: create at least one published event with a future date, and RSVP to it as a couple of users.
> 
> _Over REST, which is the path most clients take._ Both abilities are annotated read-only, so the run endpoint requires `GET` and takes the payload under an `input` parameter rather than a JSON body.
> 
> ```shell
> # Anonymous. Returns the upcoming events.
> curl -s 'https://example.test/wp-json/wp-abilities/v1/abilities/gatherpress/get-upcoming-events/run?input[count]=3'
> 
> # Anonymous, counts for a public event.
> curl -s 'https://example.test/wp-json/wp-abilities/v1/abilities/gatherpress/get-rsvp-counts/run?input[post_id]=123'
> 
> # Confirm both show up in the listing at all.
> curl -s 'https://example.test/wp-json/wp-abilities/v1/abilities' | grep gatherpress
> ```
> 
> _Or in PHP, if you would rather skip the auth plumbing._
> 
> ```
> $request = new WP_REST_Request( 'GET', '/wp-abilities/v1/abilities/gatherpress/get-rsvp-counts/run' );
> $request->set_query_params( array( 'input' => array( 'post_id' => 123 ) ) );
> print_r( rest_do_request( $request )->get_data() );
> ```
> 
> _The four checks I would actually run:_
> 
> 1. **Public listing.** Logged out, run `get-upcoming-events`. Expect the same events the events archive shows, soonest first, with `id`, `title`, `url`, `start`, `end`, `timezone`. On my run: `200`, one event, `{"id":175,"title":"Manual Check Meetup","start":"2026-08-19 14:23:43","timezone":"America/New_York", …}`.
> 2. **The count clamp.** Pass `input[count]=0` and `input[count]=999`. Neither should reach the query as-is, the value is clamped to 1 and 50.
> 3. **The password gate.** Give a published event a post password. Logged out, `get-rsvp-counts` for it should be refused. Log in as the event's author or an admin and it should succeed. This is the case that made me change the callback in [46647f0](https://github.com/GatherPress/gatherpress/commit/46647f03ed4e1113e0293b772266215bd70b164a). On my run: `401 rest_ability_cannot_execute` anonymous, `200` with the breakdown as admin.
> 4. **RSVP support, not event support.** Add `gatherpress-rsvp` support to a post type that has no `gatherpress-event-date` support:
>    ```
>    add_action( 'init', function () { add_post_type_support( 'page', 'gatherpress-rsvp' ); } );
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    `get-rsvp-counts` should work for a published page. Then remove the support and try again, it should be refused. That is the split @mauteri asked for, and it now holds. On my run: `200` with the breakdown while the support is on, `401` once it is removed.
> 
> I walked all six of those against a live site rather than only the unit tests, and they behave as described.
> 
> Happy to fold any of this into the PR description or into `docs/` if you would rather it lived somewhere permanent.

@wppoland Thank you for those details! Actually, I was most interested in the `what to prompt/what answer should I get back` part of it. Please allow me a little time to run the tests on my local abilities api set up and I'll report back to you as soon as I can. I am very happy to  have some abilities introduced on gatherpress! So thank you for opening this door.

Also, thanks for making that other fix.

cc @mauteri

wppoland:
@jmarx Understood, you wanted the prompt-and-answer shape rather than the plumbing. Here it is. Everything below is real output from a site with three published events, captured just now, not written from memory.

**What the client sees when it discovers the abilities.** This is the whole contract an assistant has to work from, so it is worth checking it reads sensibly on its own:

```
gatherpress/get-upcoming-events
  Get upcoming events — List the events that have not started yet, soonest first.
  input: { "count": integer, 1–50, default 5 }

gatherpress/get-rsvp-counts
  Get RSVP counts — Count the responses to one post that takes RSVPs, broken down by status.
  input: { "post_id": integer } (required)
```

---

### Prompt 1, the common one

> **"What events are coming up?"**

The assistant calls `gatherpress/get-upcoming-events` with `{"count": 3}` and gets back:

```json
[
  {
    "id": 720,
    "title": "Monthly WordPress Meetup",
    "url": "http://localhost:8889/?gatherpress_event=monthly-wordpress-meetup",
    "start": "2026-08-22 22:00:00",
    "end": "2026-08-23 00:00:00",
    "timezone": "America/New_York"
  },
  {
    "id": 721,
    "title": "Block Theme Workshop",
    "url": "http://localhost:8889/?gatherpress_event=block-theme-workshop",
    "start": "2026-08-29 22:00:00",
    "end": "2026-08-30 00:00:00",
    "timezone": "America/New_York"
  },
  {
    "id": 722,
    "title": "Contributor Day",
    "url": "http://localhost:8889/?gatherpress_event=contributor-day",
    "start": "2026-09-12 22:00:00",
    "end": "2026-09-13 00:00:00",
    "timezone": "America/New_York"
  }
]
```

An answer built from that reads like: *"Three coming up: Monthly WordPress Meetup on Sat 22 Aug at 6pm ET, Block Theme Workshop on Sat 29 Aug, and Contributor Day on Sat 12 Sep."* Note `start` and `end` are GMT and `timezone` is the event's own zone, so the client has what it needs to render local time without guessing. **This works logged out**, which is the change I made after your comment.

### Prompt 2, the organizer one

> **"How many people are coming to the Monthly WordPress Meetup?"**

Two calls: `get-upcoming-events` to resolve the name to an id, then `get-rsvp-counts` with `{"post_id": 720}`:

```json
{
  "all": 6,
  "attending": 4,
  "not_attending": 1,
  "waiting_list": 1,
  "no_status": 0
}
```

Answer: *"Four attending, one on the waiting list, one declined."* The name-to-id hop is the bit worth watching in your testing, since it is the assistant doing the joining rather than the API.

### Prompt 3, the one that should be refused

> **"How many people are coming to [a password-protected event]?"**

`get-rsvp-counts` returns `401 rest_ability_cannot_execute` for anyone who has not satisfied the password, and the real breakdown for the author or an admin. Worth pointing an assistant at deliberately, because a refusal it handles badly is more damaging than a wrong number.

### Prompt 4, the boundary

> **"How many people are coming to [a plain page]?"**

Refused, unless that post type declares `gatherpress-rsvp` support. With `add_post_type_support( 'page', 'gatherpress-rsvp' )` it returns the breakdown; without it, `401`. That is @mauteri's RSVP-is-its-own-support point, observable from the client side.

---

**Two things I would genuinely like your read on**, since you have an Abilities client set up and I only have curl and `rest_do_request`:

1. **Are the descriptions enough for a model to pick the right ability unprompted?** "List the events that have not started yet, soonest first" is accurate, but I do not know whether it beats a generic `wp/get-posts` in practice when someone asks "what's on next month". If it loses, the description is the thing to change.
2. **Is `post_id` the right input name now?** It was `event_id`, renamed because RSVPs are their own post type support. Clearer for the general case, but it may make an assistant less likely to connect it to an event it just listed. If so, the fix is in the description rather than the parameter.

No rush at all, and thanks, the framing questions have already made this better than what I opened.

jmarx (approved):
LGTM. Thanks for the details @wppoland I verified that the prompts work as expected using a live LLM. Great work!

wppoland:
@mauteri @jmarx approved this on the 21st after checking the prompts against a live LLM. Your change requests from the 17th were all addressed on the 18th, so this is waiting on your second look rather than on more work.

Recap of what moved:

- The capability is a class constant now, and the two abilities live in separate classes, `event/class-abilities.php` and `rsvp/class-abilities.php`, rather than sharing one.
- The RSVP ability gates on `Event::can_read_rsvps()`, whose first check is `post_type_supports( $post->post_type, 'gatherpress-rsvp' )`. It never looks at `gatherpress-event-date`, so a post type with one support and not the other behaves correctly. There is a test for that.
- Renamed to `gatherpress/get-rsvp-counts` and `get_rsvp_counts()`, with `post_id` rather than `event_id` as the input, since nothing about the ability is event-specific.
- The comment you flagged is gone.
- The permission callback on `get-upcoming-events` is removed, per jmarx. It was `current_user_can( 'read' )`, which locked out anonymous visitors who can already see the same events on the archive.

No conflicts with `develop`.

mauteri (approved):

PR #2152, opened 2026-08-14
Stop an email address standing in for a responder's name

mauteri:
## Proposed changes

Stops an email address standing in for a responder's name.

`Rsvp::save()` takes no name, so a response saved through it had nothing to store, and the store fell back to asking the provider what to call the identity. For an email identity the provider answered with the address itself, so that address was written into the display-name column — and every reader takes the name from there. The attendee list, the responses endpoint, and the core comments route all published it.

* **`Rsvp\Response\Provider\Base`** — `get_display_name()` is no longer abstract and returns an empty string. An identity that only identifies someone — an address, a URL, an external ID — has no name in it, so providers that cannot name a responder now say so rather than handing back the identifier.
* **`Rsvp\Response\Provider\Email`** — its override is deleted; the address was never a name.
* **`Rsvp\Storage`** — a response with no name to save no longer fills that column with the provider's answer.
* **`Rsvp\Query`** — read-side cover for rows already stored that way: a response whose display-name column holds the same value as its email column reads as the neutral label for anyone who does not manage RSVPs. Scoped to that equality rather than to a name merely looking like an address, since core defaults `display_name` to `user_login` and publishes it.
* **`Rsvp\Response\Serializer`** — checks the name it resolved rather than trusting the fallback.

Responses with nothing to show read as **"Attendee"**, already the label the RSVP admin table uses for that column ([class-list-table.php:112](https://github.com/GatherPress/gatherpress/blob/develop/includes/core/classes/rsvp/class-list-table.php#L112)). "Guest" would collide with the guest count, which is a different thing.

Responses saved with a real name still report it, and RSVP managers still see what a response was actually saved with.

## Testing instructions

* `npm run test:unit:php` — green. New cases cover the store, the comment filter, and the serializer, plus a response saved with a real name so the fallback cannot swallow one, an account whose display name is an address but does not match its account email, and a response that needs both the name withheld and the profile URL resolved.
* `npm run lint:php`, `composer test:phpmd`, `npm run lint:phpstan` clean.
* Verified against a live site for three shapes — a response saved with no name, one saved before this change, and one saved with a name — as both a logged-out visitor and an administrator.

### Changelog entry

`.github/changelog/rsvp-display-name-handling` — `Type: security`.

### DIFF
--- a/includes/core/classes/rsvp/class-query.php
+++ b/includes/core/classes/rsvp/class-query.php
@@ -338,13 +338,18 @@ final class Query {
 
 		// Responses saved before the store stopped writing an address into the
 		// display-name column still carry one, and every reader of a comment
-		// takes the name from there. That write put the same address in both
-		// columns, so match on their equality rather than on the name merely
-		// looking like an address: display names are an address whenever the
-		// account was registered with one, and core publishes those.
+		// takes the name from there. Two conditions, each excluding a case the
+		// other lets through: the equality is what marks the legacy write,
+		// since only an email identity ever filled the address column and it
+		// filled both with the same value, while a display name that merely
+		// looks like an address belongs to an account registered with one and
+		// is core's to publish. The address check then keeps a row whose two
+		// columns happen to match on something that is not an address — not
+		// reachable through this store, but imports and migrations write here
+		// too — from being renamed over nothing.
 		$withhold_address = ! $withhold
 			&& ! $manages_rsvps
-			&& '' !== $comment->comment_author_email
+			&& is_email( $comment->comment_author )
 			&& $comment->comment_author === $comment->comment_author_email;
 
 		// Only fill a URL the store never wrote, so a response identified by
--- a/includes/core/classes/rsvp/class-query.php
+++ b/includes/core/classes/rsvp/class-query.php
@@ -338,10 +338,14 @@ final class Query {
 
 		// Responses saved before the store stopped writing an address into the
 		// display-name column still carry one, and every reader of a comment
-		// takes the name from there.
+		// takes the name from there. That write put the same address in both
+		// columns, so match on their equality rather than on the name merely
+		// looking like an address: display names are an address whenever the
+		// account was registered with one, and core publishes those.
 		$withhold_address = ! $withhold
 			&& ! $manages_rsvps
-			&& is_email( $comment->comment_author );
+			&& '' !== $comment->comment_author_email
+			&& $comment->comment_author === $comment->comment_author_email;
 
 		// Only fill a URL the store never wrote, so a response identified by
 		// its URL keeps the one it was saved with.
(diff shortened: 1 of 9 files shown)


includes/core/classes/rsvp/class-query.php
```diff
@@ -353,6 +361,8 @@ public function prepare_rsvp_comment( $comment ) {
 			$prepared->comment_author       = __( 'Anonymous', 'gatherpress' );
 			$prepared->comment_author_email = '';
 			$prepared->comment_author_url   = '';
+		} elseif ( $withhold_address ) {
```
mattcowan: `$withhold_address` and `$resolve` are computed as independent conditions but branch as mutually exclusive, so a response matching both only gets the first. No legacy data needed to hit it — any account whose display_name is an email address (core's default when the login is an address) qualifies. On develop an anonymous reader gets comment_author_url resolved; here it stays empty. And since responses() still resolves profile from the provider, the comment path and the responses path end up disagreeing about the same responder.

```suggestion
		} else {
			if ( $withhold_address ) {
				$prepared->comment_author = __( 'Attendee', 'gatherpress' );
			}

			if ( $resolve ) {
				$prepared->comment_author_url = (string) get_author_posts_url( (int) $comment->user_id );
			}
		}
```

includes/core/classes/rsvp/class-query.php
```diff
@@ -332,16 +332,24 @@ public function prepare_rsvp_comment( $comment ) {
...
+		$manages_rsvps = current_user_can( Rsvp::CAPABILITY );
+		$withhold      = get_comment_meta( (int) $comment->comment_ID, Rsvp::ANONYMOUS_META_KEY, true )
+			&& ! $manages_rsvps;
+
+		// Responses saved before the store stopped writing an address into the
+		// display-name column still carry one, and every reader of a comment
+		// takes the name from there.
+		$withhold_address = ! $withhold
+			&& ! $manages_rsvps
+			&& is_email( $comment->comment_author );
```
mattcowan: This can't distinguish a row saved before the store changed from a current user whose display name genuinely is their address — that user gets renamed to "Attendee" for every non-manager. Is this intended? `$comment->comment_author === $comment->comment_author_email` would be narrower, though not airtight either.

mattcowan (commented):
Nice fix, confirmed the new behavior locally for both existing and newly saved responses.

Two inline notes: one behavioral, where the branch chain drops URL resolution for a case that doesn't need legacy data to reach, and one question about the scope of the `is_email()` check.

includes/core/classes/rsvp/class-query.php
mauteri: Agreed on both counts, and the second half of this is the part I'd missed.

The two conditions are computed independently but consumed as mutually exclusive, so a responder with an account, an empty `comment_author_url`, and an address in the display-name column matches both and only gets the first. No legacy row required, as you say. Taking your suggestion as written.

The disagreement you point out between the two paths is the more useful catch. `Serializer::to_array()` resolves `profile` unconditionally from the provider, with no equivalent guard, so the comment path and the responses path would answer differently about the same responder. Splitting the branches here brings them back in line.

includes/core/classes/rsvp/class-query.php
mauteri: Not intended, no — that's too wide, and your narrower check is the right one.

`is_email()` on the display-name column catches any account whose display name happens to be an address, and since core defaults `display_name` to `user_login`, anyone who registered with an address as their login qualifies. Those users would render as "Attendee" to every non-manager, which is a live behavior change rather than the cleanup this was meant to be.

`$comment->comment_author === $comment->comment_author_email` matches what the old store actually wrote, since it put the same value in both columns. Your caveat is fair — a user whose display name genuinely equals their own account email still matches — but core publishes `display_name` in comments everywhere, so widening past the legacy shape means diverging from core on something this change didn't cause.

Worth stating plainly since it drove the wrong instinct: this branch is defensive cover for rows already in the database. The actual fix is `Base::get_display_name()` no longer inventing a name from an identifier, with callers deciding what to show instead. That argues for scoping this as tightly to the old write pattern as possible. Switching to the equality check.

PR #2173, opened 2026-08-19
Promote from the waiting list by walking the queue

### LINKED ISSUE #2172: Waiting-list promotion is bounded by free spots instead of the queue, fatals when the queue is shorter
### Describe the bug

`Rsvp::check_waiting_list()` promotes people by counting off free spots rather than by walking the waiting list, and the two are different quantities. The loop is:

```php
for ( $i = 0; $i < $remaining_spots; $i++ ) {
	$state = $waiting_list[ $i ];
	$state = $this->storage->save( Intent::attend( $state ), (int) $state->comment->comment_ID );

	if ( $state instanceof State ) {
		++$promoted_count;
		$remaining_spots -= $state->get_attendee_count();
	}
}
```

`$i` indexes the queue, `$remaining_spots` measures capacity, and the body mutates the bound while the loop runs. That fails in both directions.

**Fewer waiting than free spots is a fatal.** `$waiting_list[ $i ]` reads past the end of the array, so `Intent::attend()` receives `null`:

```
PHP Fatal error: Uncaught TypeError: GatherPress\Core\Rsvp\Response\Intent::attend():
Argument #1 ($state) must be of type GatherPress\Core\Rsvp\Response\State, null given,
called in includes/core/classes/rsvp/class-rsvp.php on line 358
```

This is not an exotic state. Any event with an attendance limit reaches it as soon as one person is on the waiting list and there is room, which is exactly what happens when an attendee cancels or an organiser raises the limit. `Rsvp::save()` calls `check_waiting_list()`, so the crash surfaces on the next RSVP rather than at some later sweep.

**More waiting than free spots stops early.** With three free spots and five people waiting, single attendees with no guests: `$i = 0` promotes and `$remaining_spots` drops to 2, `$i = 1` promotes and it drops to 1, then `$i = 2 < 1` is false. Two promoted, one seat left empty, three people still queued.

### Steps to Reproduce

1. Create an event and set `gatherpress_max_attendance_limit` to 10.
2. Have one person RSVP to the waiting list.
3. Fatal `TypeError`.

As a test against `develop`:

```php
public function test_check_waiting_list_with_more_spots_than_waiting(): void {
	$event_id = $this->factory->post->create( array( 'post_type' => Event::POST_TYPE ) );
	$rsvp     = new Rsvp( $event_id );
	$user_id  = $this->factory->user->create();

	Utility::set_and_get_hidden_property( $rsvp, 'max_attendance_limit', 10 );

	$rsvp->save( $user_id, 'waiting_list' );

	$this->assertSame( 'attending', $rsvp->get( $user_id )['status'] );
}
```

### Expected behavior

Everyone on the waiting list is promoted in order until the room is full, and no arrangement of queue length against free spots raises an error.

### Environment

`develop` at 32efc3a, WordPress 7.1, PHP 8.3. Not version specific, the loop has had this shape since the method was introduced in 0.34.0.

### Additional context

Unrelated to #1771, which was about identifier resolution for Open RSVP attendees and is closed. This is the loop bound.

I have a fix and regression tests covering both directions, PR to follow.

wppoland:
Fixes #2172.

### The bug

`check_waiting_list()` walked the queue with an index bounded by the number of free spots, and decremented that bound inside the loop. `$i` counts queue positions, `$remaining_spots` counts seats, and mixing them breaks in both directions.

**Short queue, fatal.** Fewer people waiting than free spots runs `$waiting_list[ $i ]` past the end of the array and hands `null` to `Intent::attend()`, which is typed `State`. An event with an attendance limit hits this as soon as one person is waiting and a seat is free, and since `Rsvp::save()` calls `check_waiting_list()`, it fatals on the next RSVP.

**Long queue, silent under-promotion.** Three free spots and five waiting promotes only two: the bound shrinks as the index grows and they meet in the middle. A seat stays empty with people still queued.

### The fix

Walk the queue and stop when the room is full, which is what the code above it already does for the unlimited case:

```php
foreach ( $waiting_list as $state ) {
	if ( $remaining_spots <= 0 ) {
		break;
	}

	$state = $this->storage->save( Intent::attend( $state ), (int) $state->comment->comment_ID );

	if ( $state instanceof State ) {
		++$promoted_count;
		$remaining_spots -= $state->get_attendee_count();
	}
}
```

`Collection::waiting_list()` already returns a sorted, `array_values`-reindexed list, so ordering is unchanged: oldest first, same as before.

Deliberately unchanged: an entry whose party is larger than the remaining spots is still promoted rather than skipped in favour of a smaller one behind it. That overshoot is existing behaviour and changing it is a fairness decision, not a bug fix. Happy to follow up if you want it.

### Testing

Two regression tests, one per direction, in `Test_Rsvp`:

- `test_check_waiting_list_with_more_spots_than_waiting`, ten seats and one person waiting. Errors with a `TypeError` before the fix.
- `test_check_waiting_list_fills_every_free_spot`, five queued and the limit widened to three. Returns 1 instead of 2 before the fix.

Both verified to fail against unpatched `develop` and pass after, so neither is vacuous:

```
# fix reverted
Tests: 2, Assertions: 1, Errors: 1, Failures: 1.

# fix applied
OK (2 tests, 5 assertions)
```

Whole suite green, single site and multisite:

```
OK (1719 tests, 7088 assertions)
OK (323 tests, 856 assertions)
```

`includes/core/classes/rsvp/class-rsvp.php` stays at 100% coverage, 201/201 statements. PHPCS and PHPStan clean.

### DIFF
--- a/.github/changelog/waiting-list-promotion-bounds
+++ b/.github/changelog/waiting-list-promotion-bounds
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed waiting-list promotion counting off free spots instead of walking the queue. An event with an attendance limit raised a fatal `TypeError` the moment its waiting list was shorter than the number of open spots, which is any event where someone is waiting and a seat frees up. The same loop also stopped before a longer queue had filled the room, leaving seats empty with people still waiting.
--- a/includes/core/classes/rsvp/class-rsvp.php
+++ b/includes/core/classes/rsvp/class-rsvp.php
@@ -353,8 +353,15 @@ final class Rsvp {
 		// #1771 does not apply to this design.
 		$promoted_count = 0;
 
-		for ( $i = 0; $i < $remaining_spots; $i++ ) {
-			$state = $waiting_list[ $i ];
+		// Walk the queue itself rather than counting off free spots: the two are
+		// different quantities, and indexing the list by the spot count either
+		// ran off the end of a short queue or stopped before a long one had
+		// filled the room.
+		foreach ( $waiting_list as $state ) {
+			if ( $remaining_spots <= 0 ) {
+				break;
+			}
+
 			$state = $this->storage->save( Intent::attend( $state ), (int) $state->comment->comment_ID );
 
 			if ( $state instanceof State ) {
--- a/test/unit/php/includes/tests/core/classes/rsvp/class-test-rsvp.php
+++ b/test/unit/php/includes/tests/core/classes/rsvp/class-test-rsvp.php
@@ -263,6 +263,90 @@ class Test_Rsvp extends Base {
 	}
 
 	/**
+	 * Test check waiting list when the free spots outnumber the waiting list.
+	 *
+	 * The promotion loop walked one index per free spot, so a waiting list
+	 * shorter than the number of open spots ran off the end of the array.
+	 *
+	 * @since  0.36.0
+	 * @covers ::check_waiting_list
+	 *
+	 * @return void
+	 */
+	public function test_check_waiting_list_with_more_spots_than_waiting(): void {
+		$event_id = $this->factory->post->create( array( 'post_type' => Event::POST_TYPE ) );
+		$rsvp     = new Rsvp( $event_id );
+		$user_id  = $this->factory->user->create();
+
+		Utility::set_and_get_hidden_property( $rsvp, 'max_attendance_limit', 10 );
+
+		// Saving runs the promotion sweep itself, which is where the overrun
+		// used to raise a TypeError rather than returning.
+		$rsvp->save( $user_id, 'waiting_list' );
+
+		$this->assertSame(
+			'attending',
+			$rsvp->get( $user_id )['status'],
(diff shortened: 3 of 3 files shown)


wppoland:
The one red check, `wp-org-plugin-guidelines`, is infrastructure rather than this diff. It dies in the `Install dist-archive-command` step, before any plugin code is built or inspected:

```
Warning: Failed to execute git clone --mirror -- 'git@github.com:wp-cli/dist-archive-command.git' …
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
Error: Package installation failed.
```

`wp package install` fell back to an SSH clone of a public third-party repo and the runner has no key for it. The same job passed on `fix/#2140-reduced-motion` yesterday with no change to that workflow in between, and this repo has seen this particular flake before. A re-run should clear it. I do not have re-run rights here, so it needs someone with write access, or it will sort itself out on the next push.

Everything that does exercise the change is green: PHPUnit single site and multisite, PHPStan, PHPCS, coverage, and Playwright.

mauteri (approved):

PR #2182, opened 2026-08-20
Fix editor accessibility: help text, focus outlines, key handler

### LINKED ISSUE #2143: Editor accessibility polish: unassociated help text, removed focus outlines, over-broad key handler
### Describe the bug

Three small, independent accessibility gaps in the block editor UI. Grouping them because each is a few lines; they can be fixed together or one at a time.

**1. Help text isn't associated with its control.** `src/components/GuestLimit.js:76` and `src/components/MaxAttendanceLimit.js:79` render their descriptions as a sibling `<p className="description">`. Because nothing links the text to the input, a screen reader never reads it when the field takes focus. WordPress's own components accept a `help` prop that renders the same text *and* wires up `aria-describedby` — and the checkbox controls right next to these already use it, so the fix is switching to the prop the surrounding code already uses.

**2. Focus outlines are removed.** `src/components/PatternPicker/index.scss:35` and `src/blocks/venue-detail/editor.scss:55,63` set `outline: none`. In PatternPicker the replacement focus style is identical to the hover style (border colour only); on the venue address textarea the outline is removed with nothing in its place, so a keyboard user genuinely cannot tell where focus is. Any visible replacement is fine — it just has to be distinguishable from hover, and distinguishable from the unfocused state (WCAG 2.4.7 Focus Visible, Level AA — the one item here that is a real conformance failure).

**3. A key handler fires on every key.** `src/blocks/rsvp-template/edit.js:79` puts `onKeyUp={ handleOnClick }` on a `div[role="button"]`, so Tab, arrow keys, Escape — anything at all — triggers the click action. It should respond to Enter and Space only, matching what the `button` role promises.

**This would make a good first issue** — three self-contained changes in the editor UI, each a few lines, and between them a decent tour of ARIA descriptions, focus visibility, and keyboard event handling. Item 2 is the one with real conformance weight; 1 and 3 are correctness.

### Steps to Reproduce

1. **Help text:** open an event in the editor, focus the Guest Limit or Max Attendance number field with a screen reader running — the description below it is never announced.
2. **Focus outlines:** Tab into the venue address textarea, or through the pattern picker — no visible focus indicator (pattern picker) or none at all (textarea).
3. **Key handler:** with an RSVP template block selected, press an arrow key or Tab while its `div[role="button"]` has focus — the click action fires.

_No response_

### Environment information

_No response_

### WordPress information

- WordPress 7.0.3, Theme: Twenty Twenty-Five

### Code of Conduct

faisalahammad:
Fixes #2143

## Proposed changes:
* Associates help text with its control in `GuestLimit` and `MaxAttendanceLimit` by using `NumberControl` `help` prop, which wires `aria-describedby` so screen readers announce the description on focus.
* Restores visible focus indicators in `PatternPicker` and venue-detail address textarea. PatternPicker hover and focus are now distinct (hover is border only, focus adds ring/outline). Textarea `outline: none` removed and replaced with `focus`/`focus-visible` ring in admin theme color. Restores WCAG 2.4.7.
* Guards RSVP template preview key handler to Enter (keydown) and Space (keyup) only. Previously `onKeyUp` fired on any key including Tab and arrow keys.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?

## Testing instructions:

1. Guest Limit / Max Attendance help: edit an event, focus each NumberControl with a screen reader. Help text should be announced. Before, sibling `<p>` was never associated.
2. Pattern Picker focus: open pattern picker modal, Tab through cards. Focus ring should be visible and distinct from hover.
3. Venue address textarea: Tab into the textarea. Visible focus ring should appear.
4. RSVP template: focus preview `div[role="button"]`, press Tab/arrows/Esc. Click must not fire. Press Enter and Space. Each should activate.

## Screenshots
No visual change except focus states, which are covered by testing steps above.

### Changelog entry

- [x] Automatically create a changelog entry from the details below.

<details>

#### Significance

- [x] Patch

#### Type

- [x] Fixed - for any bug fixes

#### Message

Fix editor accessibility: associate help text with controls, restore visible focus indicators, and guard RSVP template key handler to Enter and Space only.

</details>

### DIFF
--- a/src/blocks/rsvp-template/edit.js
+++ b/src/blocks/rsvp-template/edit.js
@@ -59,6 +59,23 @@ const RsvpTemplatePreview = ( {
 		setActiveRsvpId( commentId );
 	};
 
+	const handleKeyDown = ( event ) => {
+		if ( 'Enter' === event.key ) {
+			event.preventDefault();
+			handleOnClick();
+		} else if ( ' ' === event.key ) {
+			// Prevent scroll on Space; activation fires on keyup per button semantics.
+			event.preventDefault();
+		}
+	};
+
+	const handleKeyUp = ( event ) => {
+		if ( ' ' === event.key ) {
+			event.preventDefault();
+			handleOnClick();
+		}
+	};
+
 	// We have to hide the preview block if the `comment` props points to
 	// the currently active block!
 
@@ -76,7 +93,8 @@ const RsvpTemplatePreview = ( {
 			style={ style }
 			// eslint-disable-next-line jsx-a11y/no-noninteractive-element-to-interactive-role
 			onClick={ handleOnClick }
-			onKeyUp={ handleOnClick }
+			onKeyDown={ handleKeyDown }
+			onKeyUp={ handleKeyUp }
 		/>
 	);
 };
--- a/src/blocks/rsvp-template/edit.js
+++ b/src/blocks/rsvp-template/edit.js
@@ -10,7 +10,7 @@ import {
 	useInnerBlocksProps,
 } from '@wordpress/block-editor';
 import { useSelect } from '@wordpress/data';
-import { memo, useState } from '@wordpress/element';
+import { memo, useEffect, useRef, useState } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -29,10 +29,30 @@ const TemplateInnerBlocks = ( {
 		{ template: TEMPLATE },
 	);
 
+	// Activating a preview hides its own button (display: none), which strands
+	// keyboard focus. When this response becomes the active one, move focus to
+	// the revealed template content so focus follows the keyboard interaction.
+	const childrenRef = useRef( null );
+
+	const isSkippedInitial = useRef( false );
+	useEffect( () => {
+		// Skip the initial render so the default first-active preview never
+		// steals focus on editor load, and only focus after this response has
+		// just become active.
+		if ( ! isSkippedInitial.current ) {
+			isSkippedInitial.current = true;
+			return;
+		}
+
+		if ( response.commentId === ( activeRsvpId || firstRsvpId ) ) {
+			childrenRef.current?.focus();
+		}
+	}, [ activeRsvpId, firstRsvpId, response.commentId ] );
+
 	return (
 		<div { ...innerBlocksProps }>
 			{ response.commentId === ( activeRsvpId || firstRsvpId )
-				? children
+				? <div ref={ childrenRef } tabIndex={ -1 }>{ children }</div>
 				: null }
 
 			<MemoizedRsvpTemplatePreview
--- a/src/blocks/venue-detail/editor.scss
+++ b/src/blocks/venue-detail/editor.scss
@@ -58,12 +58,6 @@
 		white-space: pre-wrap;
 		overflow-wrap: anywhere;
 
-		&:focus-visible {
-			outline: 1.5px solid var(--wp-admin-theme-color, #007cba);
-			outline-offset: 2px;
-			box-shadow: 0 0 0 1px var(--wp-admin-theme-color, #007cba);
-		}
-
 		&:focus {
 			outline: 1.5px solid var(--wp-admin-theme-color, #007cba);
 			outline-offset: 2px;
--- a/src/blocks/venue-detail/editor.scss
+++ b/src/blocks/venue-detail/editor.scss
@@ -52,16 +52,22 @@
 		background: transparent;
 		border: none;
 		box-shadow: none;
-		outline: none;
 		vertical-align: baseline;
 		resize: none;
 		overflow: hidden auto;
 		white-space: pre-wrap;
 		overflow-wrap: anywhere;
 
+		&:focus-visible {
+			outline: 1.5px solid var(--wp-admin-theme-color, #007cba);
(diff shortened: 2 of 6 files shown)


src/blocks/rsvp-template/edit.js
```diff
@@ -59,6 +59,23 @@ const RsvpTemplatePreview = ( {
 		setActiveRsvpId( commentId );
 	};
 
+	const handleKeyDown = ( event ) => {
```
mattcowan, line 82: Now that Enter and Space activate properly, focus is the next thing a keyboard user runs into. handleOnClick sets activeRsvpId to this preview's own commentId, which flips isHidden on this same element and applies display: none at line 85 - so the element the user just activated disappears out from under their focus.

Reproduced in the editor on an event with three RSVP responses: focus the second preview, press Enter, and document.activeElement goes from that preview to <body>. The activation itself works - the display values across the three previews flip from none, block, block to block, none, block - it is only the focus that gets stranded. The screenshots below are the same region before and after the keypress; the focus ring this PR restores is visible in the first and gone in the second.

This is not a regression, since the previous onKeyUp handler did the same thing on any key. But moving focus to the revealed content on activation would finish the job this PR starts.

Screenshot before keypress:

Image

Screenshot after keypress:

Image

src/components/PatternPicker/index.scss
```diff
@@ -29,10 +29,22 @@
 		text-align: left;
 		transition: border-color 120ms ease-in-out;
 
-		&:hover,
+		&:hover {
+			border-color: var(--wp-admin-theme-color, #007cba);
+		}
+
+		&:focus-visible {
```
mattcowan, line 36: The :focus-visible block is dead code: the :focus block below it has the same specificity and byte-identical declarations, so source order means :focus always wins. The ring does work, but it comes from :focus, which also means it shows on mouse clicks. Collapsing to :focus-visible alone keeps the keyboard ring and drops the mouse one. If you would rather keep a fallback for older browsers, the usual shape is :focus plus a :focus:not(:focus-visible) reset.


```suggestion
		&:focus-visible {
			border-color: var(--wp-admin-theme-color, #007cba);
			box-shadow: 0 0 0 1px var(--wp-admin-theme-color, #007cba);
			outline: 2px solid var(--wp-admin-theme-color, #007cba);
			outline-offset: 1px;
		}
```

src/blocks/venue-detail/editor.scss
```diff
@@ -52,16 +52,22 @@
...
 		border: none;
 		box-shadow: none;
-		outline: none;
 		vertical-align: baseline;
 		resize: none;
 		overflow: hidden auto;
 		white-space: pre-wrap;
 		overflow-wrap: anywhere;
 
+		&:focus-visible {
```
mattcowan: Same duplicate pattern here - :focus-visible at 61 and an identical :focus at 67. A text input is the one case where showing the ring on mouse focus is arguably right, so if that is deliberate, keeping only :focus states it more clearly than having both.


```suggestion
		&:focus-visible {
			outline: 1.5px solid var(--wp-admin-theme-color, #007cba);
			outline-offset: 2px;
			box-shadow: 0 0 0 1px var(--wp-admin-theme-color, #007cba);
		}
```

mattcowan (commented):
Good fix, and it covers all three items from #2143. I checked the help prop change rather than assuming it: NumberControl really does wire aria-describedby to a rendered help paragraph, so that is a genuine improvement over the unassociated description paragraph and not just the text moving.

One thing to fix before merge: keyboard activation on the RSVP preview drops focus to <body>, which I reproduced in the editor with screenshots attached. It is the natural next step now that the keys themselves work. The other two comments are the same small thing in both stylesheets - a :focus-visible block that an identical :focus block immediately overrides.

src/blocks/rsvp-template/edit.js
faisalahammad: Fixed. When a preview becomes active, focus moves to the revealed template content via a ref'd focusable div. Initial render is skipped so focus is not stolen on editor load.

src/components/PatternPicker/index.scss
faisalahammad: Fixed. Removed the duplicate :focus block. Hover draws only the border; :focus-visible draws the ring, so mouse clicks no longer show the ring.

src/blocks/venue-detail/editor.scss
faisalahammad: Fixed. Kept only :focus on this text input, since showing the ring on mouse focus is correct here. Removed the redundant :focus-visible block.

mattcowan:
@faisalahammad Thanks for turning these around so fast. All three descriptions sound right to me, but I do not think the push landed, though. The PR still shows `c7a376a6` as its head, and `fix/2143-editor-a11y-polish` on your fork is still sitting on that same commit (not seeing anything here under the three resolved conversations). At that commit `rsvp-template/edit.js` has no focus handling, and both stylesheets still have the `:focus-visible` block followed by the `:focus` one. Almost certainly just a local branch that has not been pushed yet.

Give me a shout onceyou've checked if this is pushed, and I will re-check.

mattcowan (approved):
Changes look good to me, thank you! Will need a final review by mauteri

mauteri (approved):

PR #2189, opened 2026-08-20
fix(event-date): make dates machine-readable

### LINKED ISSUE #2142: Event dates are not machine-readable

The event date block renders its date as plain text inside a `<div>` (`src/blocks/event-date/render.php`). There's no `<time datetime="…">` element, so nothing but a human reader can tell that string is a date — assistive technology, browser extensions offering "add to calendar," and search engines all see undifferentiated text.

To reproduce: view any event and inspect the date output. It sits in a `<div>` as plain text, with no `<time>` element and no `datetime` attribute.

This isn't a WCAG failure — dates render correctly for people reading them. It's a semantics and interoperability improvement (with a small SEO benefit, since search engines read `<time>`). Worth doing, not urgent.

### Designs

Wrap the output in a `<time>` element carrying a machine-readable `datetime` in ISO 8601 format, which the block already has available server-side:

```html
<time datetime="2026-08-21T18:00:00+00:00">August 21, 2026 6:00 pm to 8:00 pm UTC</time>
```

**This would make a good first issue** — a single template change in one file, and a good introduction to semantic HTML and how ISO 8601 datetime attributes work. The trickiest part is deciding how to represent a start-and-end range, which is a genuinely interesting question to think through: one `<time>` per endpoint is the usual answer.

Leaving the markup alone and relying on the plugin's existing structured-data output, if any — but `<time>` is the standard signal for this, costs nothing, and helps assistive technology and browser features that never see JSON-LD.

### Code of Conduct

faisalahammad:
Fixes #2142

## Proposed changes:
* Emit a `<time>` element with an ISO 8601 `datetime` attribute for each displayed Event Date endpoint.
* Add unfiltered ISO datetime accessors so display-format filters cannot corrupt machine-readable values.
* Keep links outside the `<time>` elements and keep the no-date placeholder without a `<time>` element.
* Share display-part logic through `Event::get_display_datetime_parts()` and use the unfiltered accessor for the core/post-date path.

### Other information:

## Before / After

| Before | After |
| ------ | ----- |
| Event dates rendered as plain text. | Event dates include machine-readable `<time>` elements. |

## Testing instructions:

1. Create and publish an event with start `2026-08-21 18:00 UTC` and end `2026-08-21 20:00 UTC`.
2. Add an Event Date block with `displayType=both` and `isLink=false`; confirm start and end render as sibling `<time>` elements with full ISO 8601 `datetime` values.
3. Set `displayType` to `start`, then `end`; confirm one `<time>` element renders for each setting.
4. Set the end date to the next day; confirm the end `datetime` retains its full ISO 8601 value.
5. Enable `isLink`; confirm the link wraps the `<time>` elements and no `<time>` is nested inside another link.
6. Create an event without dates; confirm output is `—` with no `<time>` element.
7. Set block timezone override to `yes` while global timezone display is off; confirm the frontend includes the timezone.
8. Run the markup through the W3C validator and confirm no `<time>` or `<a>` errors.

### Changelog entry

- [x] Manually created `.github/changelog/2142-machine-readable-event-date`.

#### Significance

Patch

#### Type

Fixed

#### Message

Make event dates machine-readable by adding time elements with datetime attributes to the Event Date block. The timezone also displays when the block overrides the global setting to yes.

### Review follow-up

Implemented review feedback in commit `1b6bd4da`: removed redundant escaping from the final-render path, restored explanatory comments, documented `gatherpress_datetime_format`, and reverted unrelated test `@since` changes.

Addressed the two open threads in commit `62581de5`: each human text part (separator, timezone, and the helper's `$human`) is escaped again at output time before it is wrapped, with `wp_kses()` as the second line, and a new `test_render_escapes_separator_markup` locks that in. The all-day ISO accessors now emit date only (`Y-m-d`) so an all-day event no longer carries a zoned timestamp that crosses two dates for readers in another zone.

Manual testing completed; targeted PHPUnit, PHP lint, PHPStan, and production build passed.


src/blocks/event-date/render.php
```diff
@@ -12,21 +12,117 @@
...
+	: '';
+
+$gatherpress_separator = '';
+if ( ! empty( $gatherpress_human_start ) && ! empty( $gatherpress_human_end ) ) {
+	$gatherpress_separator = $gatherpress_separator_attr
+		? $gatherpress_separator_attr
+		: __( 'to', 'gatherpress' );
+}
+
+$gatherpress_show_tz  = $gatherpress_show_tz_attr
```
mattcowan: This changes front-end output in a case the PR does not mention. When the global "show timezone" setting is off but the block's own showTimezone is yes, the block now prints the timezone where it previously did not - get_display_datetime() passes an empty format string to get_datetime_start() in that case, which returns ''.

Confirmed by rendering the same event both ways with the global setting off: the method gives September 1, 2026 6:00 pm to 8:00 pm, this template gives September 1, 2026 6:00 pm to 8:00 pm UTC.

I think the new behaviour is the correct one, and it matches edit.js:120, so the editor and front end now agree where they did not before. But feeds, emails, the admin list table and core/post-date all still go through get_display_datetime(), so the same event will print two different strings depending on the surface. Worth calling out in the changelog at least, and worth deciding whether the method should change too.

Frontend on current develop:

Image

Frontend from branch:

Image

Admin: 

Image

test/unit/php/includes/tests/core/classes/event/class-test-event.php
```diff
@@ -1280,6 +1280,148 @@ public function test_get_formatted_datetime_returns_empty_for_unparsable_value()
 		);
 	}
 
+	/**
+	 * Machine-readable datetime accessors bypass display-format filters.
+	 *
+	 * @covers ::get_datetime_start_iso
+	 * @covers ::get_datetime_end_iso
+	 */
+	public function test_get_iso_datetime_accessors(): void {
```
mattcowan, line 1416: The two new tests both cover the Event accessors, and they cover them well. Nothing covers the render rewrite, though, which is the larger half of the change - 128 lines that reimplement the display logic. The existing block render tests create their event without save_datetimes(), so they only ever hit the em-dash placeholder branch, and no test asserts that a <time datetime> is emitted at all.

A parity test would be cheap and would have caught the timezone divergence above: for a handful of attribute combinations, assert the template's text output (tags stripped) equals get_display_datetime() with the same arguments. I ran exactly that over 11 combinations plus multi-day and no-datetime events - it passes today except for the timezone case, so it would lock in the behaviour rather than requiring new expectations.

src/blocks/event-date/render.php
```diff
@@ -12,21 +12,117 @@
...
+
+$gatherpress_human_start = $gatherpress_show_start
+	? $gatherpress_event->get_datetime_start( $gatherpress_start_datetime_format )
+	: '';
+$gatherpress_iso_start   = ( $gatherpress_show_start && ! empty( $gatherpress_human_start ) )
+	? $gatherpress_event->get_datetime_start_iso()
+	: '';
+
+if ( $gatherpress_show_end ) {
+	if ( $gatherpress_show_start && $gatherpress_event->is_same_date() ) {
```
mattcowan: This branch leans on is_same_date(), which is exposed to exactly the filter corruption the ISO accessors were added to avoid. It compares two get_datetime_start/get_datetime_end calls that still run through gatherpress_datetime_format (class-event.php:346-348), so a filter that ignores its $format argument and returns, say, 'Y-m-d H:i' makes a same-day event compare unequal - I confirmed it flips from true to false - and this line then takes the else branch and prints the full end date instead of collapsing to the end time.

It ends up comparing 2026-09-01 18:00 against 2026-09-01 20:00 instead of 2026-09-01 against itself, because the filter overrides the 'Y-m-d' argument. Repro on any same-day event: add_filter( 'gatherpress_datetime_format', fn() => 'Y-m-d H:i' ); then ( new Event( $id ) )->is_same_date().

Giving is_same_date() the same unfiltered treatment as the new accessors would close it. The ICS DTSTART/DTEND in class-calendar.php are in the same position and out of scope here, but it is the same defect class if you want to sweep it.

src/blocks/event-date/render.php
```diff
@@ -12,21 +12,117 @@
...
-// Mirrors core/post-date's isLink attribute: link the datetime to the event.
+
+$gatherpress_display_type      = $attributes['displayType'] ?? '';
+$gatherpress_start_format_attr = $attributes['startDateFormat'] ?? '';
+$gatherpress_end_format_attr   = $attributes['endDateFormat'] ?? '';
+$gatherpress_separator_attr    = $attributes['separator'] ?? '';
+$gatherpress_show_tz_attr      = $attributes['showTimezone'] ?? '';
+
+// Resolve show_start / show_end the same way Event::get_display_datetime does.
+$gatherpress_show_start = $gatherpress_display_type
```
mattcowan: This is Event::get_display_datetime() copied into the template, which makes a third implementation of the same rules alongside edit.js. That is understandable - you need the start and end separately to wrap each in a <time>, and the method only returns a finished string - but the copies have already drifted, which is what the timezone comment above is about.

Returning a parts array from the Event class (start, separator, end, timezone) and letting both the method and the block build from it would give you the <time> wrapping without a fork.

One small thing the copy inherits: the __( 'to', 'gatherpress' ) fallback on line 73 is unreachable, because block.json defaults separator to the literal "to" and WP_Block fills defaults in before render. Repro: add_filter( 'gettext', fn( $t, $text, $d ) => 'gatherpress' === $d && 'to' === $text ? 'BIS' : $t, 10, 3 ); then render the block - the output still reads to. Pre-existing and not caused by this PR, but now it exists in two places.

mattcowan (commented):
This and #2065 both rewrite the tail of src/blocks/event-date/render.php. I ran a three-way merge of the two versions against their merge base: there is one conflict, in the wp_kses() output block, and the resolution is to keep this PR's widened allowlist (which adds time), then append #2065's viewer-time <span> after it. The two changes are otherwise independent; nothing else in either file collides.

Whichever merges second should expect that one conflict and nothing else.

On this PR: Thank you for submitting this, the ISO accessors do what the description says. I forced gatherpress_datetime_format to 'Y' and the accessor still returned 2026-09-01T18:00:00+00:00 while the filtered path collapsed to 2026. A New York event correctly carries -04:00, and the <time> counts are right: two for a range, one for a single endpoint, none for the placeholder.

One blocker: the timezone divergence at render.php:76. It changes front-end output and leaves the block disagreeing with feeds, emails and the admin list. Screenshots attached to that comment.

Three nitpicks, none of them blocking: no test covers the rewritten template, is_same_date() is open to the same filter corruption this PR closes elsewhere, and the display logic now exists in a third copy.

Separately, this overlaps #2065 with exactly one merge conflict, noted above.

src/blocks/event-date/render.php
faisalahammad: Fixed. get_display_datetime() now computes the timezone show flag the same way as the block and the editor: the block override wins, and global off with no override shows no timezone. Added a regression test for the override-yes global-off case, and a parity test that locks the block render to the method so the two cannot drift. The feeds, emails, admin list and core/post-date still show timezone only when the global setting is on, which is unchanged from before.

test/unit/php/includes/tests/core/classes/event/class-test-event.php
faisalahammad: Added the parity test. It runs the block render across seven attribute combinations (displayType both/start/end, isLink on/off, showTimezone yes/no, custom separator), strips the time and anchor tags, and asserts the remaining text equals get_display_datetime() called with the same arguments. Also added a machine-readable test that asserts time datetime elements are emitted for start and end, and that the placeholder path emits none.

src/blocks/event-date/render.php
faisalahammad: Fixed. is_same_date() now compares the date portion of the unfiltered ISO accessors instead of the filter-passing formatted strings. A format filter that ignores its argument no longer flips a same-day event to unequal. Added a regression test that installs a gatherpress_datetime_format filter returning a fixed Y-m-d H:i and asserts the event still compares same-date.

src/blocks/event-date/render.php
faisalahammad: I kept this scoped and did not refactor to a parts array, so the block keeps its own copy of the display logic. The parity test added in this PR locks the block render to get_display_datetime so the two cannot silently drift. On the to fallback: it is unreachable because block.json defaults separator to the literal to and WP fills that in before render, so I am leaving it as is.

faisalahammad:
The review feedback is addressed and all four review threads are resolved. Commit 17cd2ee fixes the timezone override, makes is_same_date immune to the datetime format filter, and adds a parity test plus a machine-readable time element test. All PHPCS and PHPStan checks pass, and the event test classes are green (99 tests). Ready for another look.

src/blocks/event-date/render.php
```diff
@@ -12,21 +12,117 @@
...
+}
+
+if ( $gatherpress_parts ) {
+	$gatherpress_display = implode( ' ', $gatherpress_parts );
+} else {
+	$gatherpress_display = esc_html( Event::DATETIME_PLACEHOLDER );
+}
+
+// Mirrors core/post-date's isLink: anchor wraps the <time> elements
+// so interactive content is not inside <time>.
```
mauteri: Logic should not be in templates as it is hard to follow and isn't covered by tests. Please simplify and apply to comprehensive methods. Also escaping shouldn't happen at this stage. Escaping happens when we output, not before unless there is a good reason to.

mauteri (changes requested):

src/blocks/event-date/render.php
```diff
@@ -36,5 +132,13 @@
...
-	<?php echo wp_kses( $gatherpress_display, array( 'a' => array( 'href' => true ) ) ); ?>
+	<?php
+	echo wp_kses(
+		$gatherpress_display,
+		array(
+			'a'    => array( 'href' => true ),
+			'time' => array( 'datetime' => true ),
+		)
+	);
+	?>
```
mauteri, line 68: We escape here so no need before.

src/blocks/event-date/render.php
faisalahammad: Moved all display logic to Event::get_display_datetime_parts(). It returns raw start, separator, end and timezone parts. The template only builds markup from them. Escaping happens at output only now. get_display_datetime() wraps the new method, so output stays same. Added tests for the new method.

src/blocks/event-date/render.php
faisalahammad: Removed escaping before final output. The template now escapes each text part only while building final markup, then applies the final allowed HTML output.

faisalahammad:
Both mauteri threads addressed in 3b5c062. Display logic now lives in Event::get_display_datetime_parts() with tests. Template is thin markup only and escapes at output. Please take another look when you have a moment.

mauteri:
@faisalahammad tests are failing.

src/blocks/event-date/render.php
```diff
@@ -16,17 +16,38 @@
...
+
+	return empty( $iso )
+		? esc_html( $human )
+		: sprintf( '<time datetime="%s">%s</time>', esc_attr( $iso ), esc_html( $human ) );
+};
+
+$gatherpress_output_parts = array_filter(
+	array(
+		$gatherpress_render_part( $gatherpress_parts['start'], $gatherpress_event->get_datetime_start_iso() ),
+		esc_html( $gatherpress_parts['separator'] ),
```
mauteri: remove escape here since it is done later.

src/blocks/event-date/render.php
```diff
@@ -16,17 +16,38 @@
...
+		$gatherpress_render_part( $gatherpress_parts['end'], $gatherpress_event->get_datetime_end_iso() ),
+		esc_html( $gatherpress_parts['timezone'] ),
```
mauteri: remove escape here since it is done later.

src/blocks/event-date/render.php
```diff
@@ -16,17 +16,38 @@
...
 	)
 );
 
-// Mirrors core/post-date's isLink attribute: link the datetime to the event.
+$gatherpress_display = $gatherpress_output_parts
+	? implode( ' ', $gatherpress_output_parts )
+	: esc_html( Event::DATETIME_PLACEHOLDER );
```
mauteri: remove escape here since it is done later.

test/unit/php/includes/tests/core/classes/blocks/class-test-event-date.php
```diff
@@ -262,7 +262,7 @@ public function test_validate_event_with_empty_content(): void {
 	 * Mirrors core/post-date's isLink behavior: the datetime output is
 	 * wrapped in a link to the event.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
```
mauteri: why'd this change?

test/unit/php/includes/tests/core/classes/blocks/class-test-event-date.php
```diff
@@ -288,7 +288,7 @@ public function test_render_links_datetime_to_event_when_islink_set(): void {
 	/**
 	 * Coverage for the rendered block without the isLink attribute.
 	 *
-	 * @since 0.35.0
+	 * @since 0.36.0
```
mauteri: why'd this change?

mauteri (changes requested):

src/blocks/event-date/render.php
```diff
@@ -16,17 +16,38 @@
...
+	$attributes['showTimezone'] ?? ''
+);
+
+$gatherpress_render_part = static function ( $human, $iso ): string {
+	if ( empty( $human ) ) {
+		return '';
+	}
+
+	return empty( $iso )
+		? esc_html( $human )
```
mauteri: Remove escape here not needed

mauteri (changes requested):

includes/core/classes/event/class-event.php
```diff
@@ -278,84 +278,99 @@ public function get_display_datetime(
...
-
-			$end = $show_start && $this->is_same_date()
+			$end                 = $show_start && $this->is_same_date()
 				? $this->get_time_end( $end_time_format )
 				: $this->get_datetime_end( $end_datetime_format );
 		} else {
 			$end = false;
 		}
 
-		// Add separator if there's both start and end date/time.
```
mauteri, line 265: Why did you remove this comment?

includes/core/classes/event/class-event.php
```diff
@@ -524,17 +561,53 @@ public function get_formatted_datetime(
 				return '';
 			}
 
+			if ( $apply_filter ) {
+				$format = apply_filters( 'gatherpress_datetime_format', $format, $which, $local );
```
mauteri, line 775: Filters need docblock, also new filters should follow filter/action naming documentation.

mauteri (changes requested):

src/blocks/event-date/render.php
faisalahammad: Removed the redundant escaping in the latest commit; final `wp_kses()` output handles allowed markup.

src/blocks/event-date/render.php
faisalahammad: Removed the redundant escaping in the latest commit; final `wp_kses()` output handles allowed markup.

src/blocks/event-date/render.php
faisalahammad: Removed the redundant escaping in the latest commit; final `wp_kses()` output handles allowed markup.

test/unit/php/includes/tests/core/classes/blocks/class-test-event-date.php
faisalahammad: Reverted this unchanged test's `@since` value to `0.35.0`; only newly added tests use the current development version.

test/unit/php/includes/tests/core/classes/blocks/class-test-event-date.php
faisalahammad: Reverted this unchanged test's `@since` value to `0.35.0`; only newly added tests use the current development version.

src/blocks/event-date/render.php
faisalahammad: Removed the redundant escaping in the latest commit; final `wp_kses()` output handles allowed markup.

includes/core/classes/event/class-event.php
faisalahammad: Restored this comment in the latest commit. It explains why the separator is only included when both start and end values exist.

includes/core/classes/event/class-event.php
faisalahammad: Added canonical hook documentation immediately above `apply_filters( 'gatherpress_datetime_format', ... )`, including `@since 0.34.0`, parameters, and return value.

faisalahammad:
Review follow-up complete in commit `1b6bd4da`:

- Removed redundant escaping from `render.php`; final `wp_kses()` handles output.
- Restored explanatory comments in `Event::get_display_datetime_parts()`.
- Added canonical documentation for `gatherpress_datetime_format`.
- Reverted unrelated test `@since` changes.
- Manual testing completed.

Verification passed: targeted PHPUnit (`100 tests, 296 assertions`), PHP lint, PHPStan, production build, and plugin zip build. Please re-review.

faisalahammad:
CI fix and develop refresh completed.

- Merged current `develop` into `fix/2142-machine-readable-event-date` and resolved the Event internals conflict while preserving the machine-readable Event Date changes.
- Fixed all-day ISO datetime accessors so `gatherpress_datetime_format` display filters cannot corrupt `<time datetime>` values.
- Corrected the `@covers` annotation for `GatherPress\Core\Utility::remove_time_format_chars`.

Validation:
- PHPUnit: 1,921 tests, 7,488 assertions, no warnings.
- PHPStan: clean with the CI command.
- PHPCS: clean.
- Production build: successful.
- Manual testing completed from `gatherpress-pr-2189.zip`.

Pushed commit `e8be519e` to the PR branch.

faisalahammad:
## CI fix — coverage gate

The only failing check was **Check Test Coverage on Changed Files**: `includes/core/classes/event/class-event.php` at 99.41%, lines 299 and 737 uncovered.

Both lines already ran in the existing tests, but xdebug does not trace them when they are reached through same-class delegation (`get_display_datetime()` -> `get_display_datetime_parts()`, and `get_formatted_datetime()` -> the private `format_datetime()`). Locally, line 298 traced 7 hits while line 299 traced 0. This is the same tracing gap `test_format_datetime_branches` already works around, which covers only the timed branches.

Added two tests in `test/unit/php/includes/tests/core/classes/event/class-test-event.php`:

| # | Test | Line covered |
|---|------|--------------|
| 1 | `test_get_display_datetime_parts_omits_a_refused_timezone` — calls `get_display_datetime_parts()` directly for an event with `gatherpress_show_timezone = never` while the block asks for `yes`; asserts `timezone` is `false` and start/end still render | 299 |
| 2 | `test_format_datetime_routes_an_all_day_event` — invokes `format_datetime()` directly for an all-day event with the display-format filter applied and skipped, plus a `get_datetime_start_iso()` assertion showing the ISO value ignores the filter | 737 |

Verified locally: lines 299 and 737 both trace `count=1`, and `check-pr-coverage.php` reports `class-event.php` at 100% (338/338) and `class-setup.php` at 100%.

PHPUnit single-site 1923 tests, multisite 325 tests, PHPCS and PHPStan all pass. No plugin behavior changed.

carstingaxion (commented):
Tested manually in PR Playground and can confirm:

- [x] renders `<time>` elements with proper `datetime` attribute
- [x] respects All-day events 
- [x] works with multiday events with start and end time
- [x] works with multiday all-day events

includes/core/classes/event/class-event.php
```diff
@@ -231,24 +231,47 @@ public function get_display_datetime(
...
+		return $parts ? implode( ' ', $parts ) : self::DATETIME_PLACEHOLDER;
+	}
 
-		$show_end = $type
-			? in_array( $type, array( 'end', 'both' ), true )
-			: true;
+	/**
+	 * Gets raw formatted datetime parts for display.
+	 *
+	 * @since 0.34.0
```
mauteri: Should be `0.36.0`

includes/core/classes/event/class-event.php
```diff
@@ -519,12 +539,19 @@ protected static function to_day_boundary( string $datetime, string $which ): st
...
 	 *
 	 * @return string The formatted datetime.
 	 */
-	protected function get_formatted_all_day( string $format, string $which, bool $local, array $dt ): string {
+	protected function get_formatted_all_day(
+		string $format,
+		string $which,
+		bool $local,
+		array $dt,
+		bool $apply_filter
```
mauteri: Should have a default value
`bool $apply_filter = true`

includes/core/classes/event/class-event.php
```diff
@@ -678,12 +707,34 @@ public function get_formatted_datetime(
...
+	 * @param string $format       PHP date format.
+	 * @param string $which        Datetime field in event table to format ('start' or 'end').
+	 * @param bool   $local        Whether to format the date in local time (true) or GMT (false).
+	 * @param bool   $apply_filter Whether to pass $format through the gatherpress_datetime_format filter.
+	 *
+	 * @return string The formatted datetime value.
+	 *
+	 * @throws Exception If there is an issue while formatting the datetime value.
+	 */
+	private function format_datetime( string $format, string $which, bool $local, bool $apply_filter ): string {
```
mauteri: I would have the default value of `bool $apply_filter = true` here as well.

mauteri (changes requested):

includes/core/classes/event/class-event.php
faisalahammad: Changed to 0.36.0 in commit 7e370227.

includes/core/classes/event/class-event.php
faisalahammad: Added the default value, `bool $apply_filter = true`, in commit 7e370227.

includes/core/classes/event/class-event.php
faisalahammad: Added the default value here too, in commit 7e370227.

faisalahammad:
Addressed the three review threads in commit 7e370227:

- `get_display_datetime_parts()` now uses `@since 0.36.0`.
- `$apply_filter` has a default of `true` in both `get_formatted_all_day()` and `format_datetime()`. All existing calls still pass the flag explicitly, so behavior is unchanged.

PHPCS and the production build pass. Please take another look when you have a moment.

src/blocks/event-date/render.php
```diff
@@ -16,17 +16,38 @@
...
+
+	return empty( $iso )
+		? $human
+		: sprintf( '<time datetime="%s">%s</time>', esc_attr( $iso ), $human );
+};
+
+$gatherpress_output_parts = array_filter(
+	array(
+		$gatherpress_render_part( $gatherpress_parts['start'], $gatherpress_event->get_datetime_start_iso() ),
+		$gatherpress_parts['separator'],
```
mattcowan: The separator reaches the output without escaping now. The old render passed the whole display string through `esc_html()` before `wp_kses()`; this one implodes the parts raw, and `wp_kses()` now allows `<a href>`. So markup typed into the Separator control renders as markup. I rendered the same page on both this branch and develop with a separator of `<a href="https://attacker.example">to</a>`: develop shows the tag as text, this branch shows a live link.

Develop, markup is escaped as visible text, safe:

develop: the separator renders as visible escaped text, not a link

This branch, 'to' renders as a live link:

this branch: the separator renders as a live link on the word to

Escape each human part before it is wrapped: the separator here, `$gatherpress_parts['timezone']` on line 43, and `$human` on both branches of the helper at lines 33–35. That restores what `esc_html()` did before, and `wp_kses()` becomes the second line rather than the only one. A test with markup in the separator would keep it that way — the only separator case now is `' — '`.

```suggestion
		esc_html( $gatherpress_parts['separator'] ),
```

Suggestion for lines 33–35:

```suggestion
	return empty( $iso )
		? esc_html( $human )
		: sprintf( '<time datetime="%s">%s</time>', esc_attr( $iso ), esc_html( $human ) );
```

includes/core/classes/event/class-event.php
```diff
@@ -709,32 +760,64 @@ public function get_formatted_datetime(
...
+	 * by display-format customization.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @return string The ISO 8601 start datetime, or empty string when unset.
+	 *
+	 * @throws Exception If there is an issue formatting the start datetime.
+	 */
+	public function get_datetime_start_iso(): string {
+		return $this->format_datetime( 'c', 'start', true, false );
```
mattcowan: An all-day event gets a zoned timestamp range here: the test at line 1454 pins `2025-06-15T00:00:00-04:00` to `T23:59:59-04:00`, and a rendered all-day event carries `datetime="2026-09-28T00:00:00-04:00"`. A reader in another zone sees a timed event that crosses two dates. For `is_all_day()`, emit the date only — `datetime="2025-06-15"` is a valid date string for `<time>` and is what an all-day event means. `get_datetime_start( 'c' )` already returns date-only for all-day in the same test, so this accessor would match it. Same for `get_datetime_end_iso()` at line 818.

mattcowan (commented):
Reviewed at `7e370227'. mauteri's three Sep 12 points look like they have been addressed. Two inline comments: the separator and the other human parts lost their escaping in the refactor and need `esc_html()` back, with a test; all-day events should emit a date-only `datetime`.

src/blocks/event-date/render.php
faisalahammad: Fixed in 62581de5. Each human part is escaped again at output time before it is wrapped: the separator, the timezone part, and `$human` on both branches of the helper. `wp_kses()` is the second line now, not the only one. Added `test_render_escapes_separator_markup` with a separator of `<a href="https://attacker.example">to</a>`; it asserts the escaped text is present and the live anchor is not.

includes/core/classes/event/class-event.php
faisalahammad: Fixed in 62581de5. `get_datetime_start_iso()` and `get_datetime_end_iso()` now emit date only, `Y-m-d`, when `is_all_day()`, and keep the full ISO `c` format for a timed event. Updated the three all-day assertions to the date-only value.

src/blocks/event-date/render.php
faisalahammad: Fixed in 62581de5. Each human part is escaped again at output time before it is wrapped: the separator, the timezone part, and `$human` on both branches of the helper. `wp_kses()` is the second line now, not the only one. Added `test_render_escapes_separator_markup` with a separator of `<a href="https://attacker.example">to</a>`; it asserts the escaped text is present and the live anchor is not.

faisalahammad:
Updated and pushed as `62581de5`. Both open review threads are replied to and resolved. CodeRabbit is green. Please re-review when available, @mattcowan and @mauteri.

includes/core/classes/event/class-event.php
faisalahammad: The @since check requires TBD for every new symbol on develop, and release tooling resolves it at release time. I changed the new tags back to TBD so the check and the project convention both pass.

faisalahammad:
## CI fix — @since tags

The only failing check was **@since tags**: "New @since tags say TBD". The PR added 13 tags with a literal `0.36.0` on new symbols. The check requires `TBD` on develop.

Why TBD is correct: a new symbol cannot know which release it ships in. Work on develop may go out in the next minor, or be cherry-picked into a patch release first. The release tooling resolves `TBD` to the version that actually ships. A hand written version is a guess, and a wrong one every time a fix is backported.

Changed the 13 new tags from `0.36.0` to `TBD`. No code change.

| File | Tags |
|------|------|
| `includes/core/classes/event/class-event.php` | `get_display_datetime_parts`, `format_datetime`, `get_datetime_start_iso`, `get_datetime_end_iso` |
| `test/unit/php/includes/tests/core/classes/blocks/class-test-event-date.php` | 4 test docblocks |
| `test/unit/php/includes/tests/core/classes/event/class-test-event.php` | 5 test docblocks |

This reverses the earlier change in `5647652189` where `get_display_datetime_parts()` was set to `0.36.0` from review feedback. The project gate and the project convention both want `TBD`, so the literal cannot stay. I left a note on that thread.

The diff is 13 lines, all pure value replacements, so each hunk removes a literal and adds a non literal. That is the shape the check counts per change run.

Verified locally: PHPCS (3 files, 0 errors, 0 warnings), PHPStan level 8 with 4G, and the targeted PHPUnit run for the two event test classes are all green. Pushed as `79d67739`.

mattcowan (approved):
Re-reviewed at `79d67739`, both of my comments are fixed and have regression tests. 

@mauteri The new `@since` tags now say `TBD` instead of `0.36.0`. The rule changed on September 20, after your comment, in #2326 and #2327. The "New @since tags say TBD" check failed on the previous push and passes now.

PR #2194, opened 2026-08-21
Revert "Revert "Surface the resolved RSVP filter count in the editor""

mauteri:
Reverts GatherPress/gatherpress#2193

### DIFF
--- a/.github/changelog/2055-rsvp-filter-count-hint
+++ b/.github/changelog/2055-rsvp-filter-count-hint
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
+Show what the RSVP response filter labels resolve to in the editor, so the raw %d placeholder is explained rather than just displayed.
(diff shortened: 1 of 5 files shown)


mauteri:
@wppoland accidentally merged your PR and reverted a revert.

mauteri:
Closing in favor of #2207, which takes a different approach.

This PR fixed the raw `%d` by teaching the dropdown-item block about RSVP: a class-to-status map, an apiFetch to the event endpoint, and inspector copy about response counts. The dropdown blocks are independent of RSVP Response, so that knowledge does not belong in them.

#2207 inverts it. The dropdown item and the select-mode trigger apply a generic `gatherpress.dropdownItemText` filter, and the RSVP Response block registers the handler that knows about statuses and counts. The item shows the stored text while selected (so the author can see and keep the placeholder) and the resolved text when deselected, with no explanatory message needed. Anything can leverage the same filter for its own placeholders.

Thanks for the work here, it mapped the problem out.

PR #2201, opened 2026-08-21
Announce links that open in a new tab for screen readers

### LINKED ISSUE #2141: Links that open in a new tab give no warning
### Describe the bug

Several links open in a new tab without telling anyone first. Screen-reader users get no warning before their context changes, and sighted keyboard users lose their place with no explanation — the browser back button no longer returns them where they expect.

Confirmed on `develop`:

- Online event link — `src/blocks/online-event-link/render.php:64` (and the runtime swap in `view.js:69`)
- Add to calendar → Google Calendar and Yahoo Calendar — `src/blocks/add-to-calendar/template.js`
- Venue website links, where the author has set the link to open in a new tab
- A couple of editor-side links (`event-date/edit.js:374`, `form-field/edit.js:356`)

The usual fix is a visually hidden note inside the link, so screen readers announce it and sighted users see no change:

```html
<a href="…" target="_blank" rel="noopener noreferrer">
    Google Calendar<span class="screen-reader-text"> (opens in a new tab)</span>
</a>
```

WordPress core does this throughout the admin, so there's plenty of precedent to copy. This is technique G201, supporting WCAG 3.2.5 — good practice rather than an AA conformance failure, so there's no urgency; it's just a courtesy the plugin currently doesn't extend.

**This would make a good first issue** — the same small pattern repeated in a handful of places, no build-system surprises, and a nice introduction to screen-reader-only text. Whoever takes it can scope it to just the frontend links if the editor ones feel like too much.

### Steps to Reproduce

1. View an event with an online event link, or open the Add to calendar menu.
2. Inspect the link: `target="_blank"` with no accompanying text.
3. With a screen reader, activate it — a new tab opens with no prior warning.

_No response_

### Environment information

_No response_

### WordPress information

- WordPress 7.0.2, Theme: Twenty Twenty-Five

### Code of Conduct

faisalahammad:
Fixes #2141

## Proposed changes:

* Links that open in a new tab now include a visually hidden `(opens in a new tab)` warning inside the anchor, following WCAG technique G201 for Success Criterion 3.2.5. Screen reader users hear the context change before it happens; sighted users see no change.
* Covered locations: the online event link block (server render plus the client-side span-to-link swap in `view.js`), the Google Calendar and Yahoo Calendar links in the Add to calendar block, venue website links when "Open in new tab" is enabled, and linked static maps in the Venue Map block.
* The online event link warning is announced through WordPress core's `.screen-reader-text` class, so it stays hidden even when a theme does not define that class.
* For the Add to calendar block, the warning is no longer baked into the saved dropdown item. Instead it is injected on the server for every new-tab link at render time, so existing published events gain the announcement without being re-saved and dropdown labels stay clean in the editor.
* In `view.js` the warning uses a dedicated `gatherpress-new-tab-notice` marker class, so the swap logic only removes the warning this plugin generated and never touches other screen-reader spans (tooltips, admin lists).
* The idempotency check in `inject_new_tab_notices` now matches only the exact `screen-reader-text gatherpress-new-tab-notice` class attribute the plugin injects. A calendar URL or visible link text that happens to contain the marker text no longer suppresses the warning.
* Editor-side links (`event-date`, `form-field` helper text) are intentionally deferred, per the issue discussion scoping this to frontend links.

### Other information:

Yes. PHP unit tests cover the new render-time injection for the Add to calendar block: a `target="_blank"` anchor gains the notice, a download link (iCal/Outlook) stays untouched, and already-marked anchors are left unchanged. Regression cases confirm the notice is still injected when the marker text appears in the anchor href or in the visible link text. A new Jest suite (`view.test.js`) covers the online event link client-side swap: first-run init, early return, span-to-link and link-to-span transitions, notice de-duplication on repeated updates, and the missing-warning case. The existing suites for these blocks pass unchanged, and the announcement behavior needs a screen reader to verify.

## Testing instructions:

* Go to an event with an online event URL set and view it on the frontend.
* With a screen reader active (VoiceOver on macOS or NVDA on Windows), tab to the online event link. The announced name should end with "opens in a new tab". Nothing should be visible on screen and there should be no layout shift.
* Open the Add to calendar dropdown. The Google Calendar and Yahoo Calendar options should announce the warning; iCal and Outlook should not, since they download files instead of opening a tab. This applies to existing events as well as newly inserted ones.
* Create a venue with a Website value, add the Venue Details block with "Open in new tab" enabled. The website link should announce the warning. Turning the setting off should remove the warning from the announced name.
* View the venue static map. When the map image is wrapped in a new-tab link, activating it should announce the warning after the image alt text.
* Result: every new-tab link on the frontend announces the warning, sighted output is unchanged.

### Changelog entry

Changelog Entry Details

#### Significance

-   [x] Patch

#### Type

#### Message

Links that open in a new tab now include a visually hidden warning so screen readers announce the context change before it happens.

### DIFF
--- a/includes/core/classes/blocks/class-add-to-calendar.php
+++ b/includes/core/classes/blocks/class-add-to-calendar.php
@@ -65,6 +65,9 @@ final class Add_To_Calendar {
 		$render_block_hook = sprintf( 'render_block_%s', self::BLOCK_NAME );
 
 		add_filter( $render_block_hook, array( $this, 'replace_calendar_placeholders' ), 10, 2 );
+		// Priority 11 so the notice is injected after replace_calendar_placeholders
+		// (priority 10) has built the final per-service hrefs onto the anchors,
+		// and after any third-party render filter at the default priority.
 		add_filter( $render_block_hook, array( $this, 'inject_new_tab_notices' ), 11 );
 	}
 
@@ -132,6 +135,17 @@ final class Add_To_Calendar {
 	 * marker class are left untouched, keeping the transform idempotent when
 	 * the filter runs more than once against the same content.
 	 *
+	 * Uses WP_HTML_Tag_Processor rather than a regex so the tag parser owns
+	 * every decision: a target only counts when it is the actual target
+	 * attribute (not text inside another attribute), unquoted target=_blank is
+	 * matched, and the existing-notice check runs through has_class so any
+	 * quoting or class order on this or the injected span is honored.
+	 *
+	 * The processor cannot insert markup (attributes and plain text only), so
+	 * the parser records its decisions as temporary data-* markers with known,
+	 * uniform quoting and a second pass does the single splice at each marked
+	 * anchor's closing tag.
+	 *
 	 * @since 0.36.0
 	 *
 	 * @param string $block_content The block content to parse.
@@ -139,31 +153,65 @@ final class Add_To_Calendar {
 	 * @return string The block content with new-tab notices injected.
 	 */
 	public function inject_new_tab_notices( string $block_content ): string {
-		return (string) preg_replace_callback(
-			'/(<a\b[^>]*?target=["\']_blank["\'][^>]*>)(.*?)(<\/a>)/is',
-			static function ( array $matches ): string {
-				// Only skip when the marker is present as the exact class
-				// attribute this plugin injects, not as arbitrary text anywhere
-				// in the anchor HTML. A calendar URL, visible link text, or
-				// nested non-class attribute carrying the marker text must not
-				// suppress the required warning.
-				if (
-					str_contains(
-						$matches[2],
-						'class="screen-reader-text gatherpress-new-tab-notice"'
-					)
-				) {
-					return $matches[0];
+		$processor = new WP_HTML_Tag_Processor( $block_content );
+		$inside    = false;
+
+		// First pass: the parser makes every decision, recorded as temporary
+		// markers written with known, uniform quoting.
+		while ( $processor->next_tag( array( 'tag_closers' => 'visit' ) ) ) {
+			if ( 'A' === $processor->get_tag() ) {
+				if ( $processor->is_tag_closer() ) {
+					$inside = false;
+				} elseif ( '_blank' === $processor->get_attribute( 'target' ) ) {
+					$inside = true;
+					$processor->set_attribute( 'data-gatherpress-new-tab', '1' );
 				}
 
-				$notice = sprintf(
-					'<span class="screen-reader-text gatherpress-new-tab-notice"> %1$s</span>',
-					esc_html__( '(opens in a new tab)', 'gatherpress' )
-				);
+				continue;
+			}
 
-				return $matches[1] . $matches[2] . $notice . $matches[3];
-			},
-			$block_content
+			if (
+				$inside
+				&& 'SPAN' === $processor->get_tag()
+				&& ! $processor->is_tag_closer()
+				&& $processor->has_class( 'gatherpress-new-tab-notice' )
+			) {
+				$processor->set_attribute( 'data-gatherpress-has-notice', '1' );
+			}
+		}
+
+		$html = $processor->get_updated_html();
+
+		// Second pass: pure splicing on the markers the parser wrote. Anchors
+		// cannot nest, so the next </a> after a marked opener closes it.
+		$marker = ' data-gatherpress-new-tab="1"';
+		$notice = sprintf(
+			'<span class="screen-reader-text gatherpress-new-tab-notice"> %1$s</span>',
+			esc_html__( '(opens in a new tab)', 'gatherpress' )
+		);
+
+		$offset   = 0;
+		$position = strpos( $html, $marker, $offset );
+
+		while ( false !== $position ) {
+			$close = stripos( $html, '</a>', $position );
+
+			if ( false !== $close ) {
+				$range = substr( $html, $position, $close - $position );
+
+				if ( ! str_contains( $range, 'data-gatherpress-has-notice' ) ) {
+					$html = substr_replace( $html, $notice, $close, 0 );
+				}
+			}
+
+			$offset   = $position + 1;
+			$position = strpos( $html, $marker, $offset );
+		}
+
+		return str_replace(
+			array( $marker, ' data-gatherpress-has-notice="1"' ),
+			'',
+			$html
 		);
 	}
 }
--- a/includes/core/classes/blocks/class-add-to-calendar.php
+++ b/includes/core/classes/blocks/class-add-to-calendar.php
@@ -142,7 +142,17 @@ final class Add_To_Calendar {
 		return (string) preg_replace_callback(
 			'/(<a\b[^>]*?target=["\']_blank["\'][^>]*>)(.*?)(<\/a>)/is',
 			static function ( array $matches ): string {
-				if ( str_contains( $matches[0], 'gatherpress-new-tab-notice' ) ) {
+				// Only skip when the marker is present as the exact class
+				// attribute this plugin injects, not as arbitrary text anywhere
+				// in the anchor HTML. A calendar URL, visible link text, or
+				// nested non-class attribute carrying the marker text must not
+				// suppress the required warning.
+				if (
+					str_contains(
+						$matches[2],
+						'class="screen-reader-text gatherpress-new-tab-notice"'
+					)
+				) {
 					return $matches[0];
 				}
 
--- a/includes/core/classes/blocks/class-add-to-calendar.php
+++ b/includes/core/classes/blocks/class-add-to-calendar.php
@@ -186,8 +186,8 @@ final class Add_To_Calendar {
 		// cannot nest, so the next </a> after a marked opener closes it.
 		$marker = ' data-gatherpress-new-tab="1"';
 		$notice = sprintf(
-			'<span class="screen-reader-text gatherpress-new-tab-notice"> %1$s</span>',
-			esc_html__( '(opens in a new tab)', 'gatherpress' )
+			'<span class="screen-reader-text gatherpress-new-tab-notice">%1$s</span>',
+			esc_html__( ' (opens in a new tab)', 'gatherpress' )
 		);
 
 		$offset   = 0;
--- a/includes/core/classes/blocks/class-add-to-calendar.php
+++ b/includes/core/classes/blocks/class-add-to-calendar.php
@@ -65,6 +65,7 @@ final class Add_To_Calendar {
 		$render_block_hook = sprintf( 'render_block_%s', self::BLOCK_NAME );
 
 		add_filter( $render_block_hook, array( $this, 'replace_calendar_placeholders' ), 10, 2 );
+		add_filter( $render_block_hook, array( $this, 'inject_new_tab_notices' ), 11 );
 	}
 
 	/**
@@ -120,5 +121,39 @@ final class Add_To_Calendar {
 		}
 
 		return $tag->get_updated_html();
+	}
+
+	/**
+	 * Inject a screen-reader new-tab notice into each targeted _blank link.
+	 *
+	 * Walks the block content and appends a visually hidden "( opens in a new
+	 * tab )" warning to every anchor that opens in a new tab, so screen-reader
+	 * users get the same cue sighted users see. Anchors that already carry the
+	 * marker class are left untouched, keeping the transform idempotent when
+	 * the filter runs more than once against the same content.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @param string $block_content The block content to parse.
+	 *
+	 * @return string The block content with new-tab notices injected.
+	 */
+	public function inject_new_tab_notices( string $block_content ): string {
+		return (string) preg_replace_callback(
+			'/(<a\b[^>]*?target=["\']_blank["\'][^>]*>)(.*?)(<\/a>)/is',
+			static function ( array $matches ): string {
+				if ( str_contains( $matches[0], 'gatherpress-new-tab-notice' ) ) {
+					return $matches[0];
+				}
+
+				$notice = sprintf(
+					'<span class="screen-reader-text gatherpress-new-tab-notice"> %1$s</span>',
+					esc_html__( '(opens in a new tab)', 'gatherpress' )
+				);
+
+				return $matches[1] . $matches[2] . $notice . $matches[3];
+			},
+			$block_content
+		);
 	}
 }
--- a/src/blocks/add-to-calendar/template.js
+++ b/src/blocks/add-to-calendar/template.js
@@ -40,7 +40,7 @@ const TEMPLATE = [
 					[
 						'gatherpress/dropdown-item',
 						{
-							text: `<a href="#gatherpress-google-calendar" rel="noreferrer noopener nofollow" target="_blank">${ __( 'Google Calendar', 'gatherpress' ) }<span class="screen-reader-text"> ${ __( '(opens in a new tab)', 'gatherpress' ) }</span></a>`,
+							text: `<a href="#gatherpress-google-calendar" rel="noreferrer noopener nofollow" target="_blank">${ __( 'Google Calendar', 'gatherpress' ) }</a>`,
 							metadata: {
 								name: __( 'Google Calendar', 'gatherpress' ),
 							},
@@ -67,7 +67,7 @@ const TEMPLATE = [
 					[
 						'gatherpress/dropdown-item',
 						{
-							text: `<a href="#gatherpress-yahoo-calendar" rel="noreferrer noopener nofollow" target="_blank">${ __( 'Yahoo Calendar', 'gatherpress' ) }<span class="screen-reader-text"> ${ __( '(opens in a new tab)', 'gatherpress' ) }</span></a>`,
+							text: `<a href="#gatherpress-yahoo-calendar" rel="noreferrer noopener nofollow" target="_blank">${ __( 'Yahoo Calendar', 'gatherpress' ) }</a>`,
 							metadata: {
 								name: __( 'Yahoo Calendar', 'gatherpress' ),
 							},
--- a/src/blocks/add-to-calendar/template.js
+++ b/src/blocks/add-to-calendar/template.js
@@ -40,7 +40,7 @@ const TEMPLATE = [
 					[
 						'gatherpress/dropdown-item',
 						{
-							text: `<a href="#gatherpress-google-calendar" rel="noreferrer noopener nofollow" target="_blank">${ __( 'Google Calendar', 'gatherpress' ) }</a>`,
+							text: `<a href="#gatherpress-google-calendar" rel="noreferrer noopener nofollow" target="_blank">${ __( 'Google Calendar', 'gatherpress' ) }<span class="screen-reader-text"> ${ __( '(opens in a new tab)', 'gatherpress' ) }</span></a>`,
 							metadata: {
 								name: __( 'Google Calendar', 'gatherpress' ),
 							},
@@ -67,7 +67,7 @@ const TEMPLATE = [
 					[
 						'gatherpress/dropdown-item',
 						{
-							text: `<a href="#gatherpress-yahoo-calendar" rel="noreferrer noopener nofollow" target="_blank">${ __( 'Yahoo Calendar', 'gatherpress' ) }</a>`,
+							text: `<a href="#gatherpress-yahoo-calendar" rel="noreferrer noopener nofollow" target="_blank">${ __( 'Yahoo Calendar', 'gatherpress' ) }<span class="screen-reader-text"> ${ __( '(opens in a new tab)', 'gatherpress' ) }</span></a>`,
 							metadata: {
 								name: __( 'Yahoo Calendar', 'gatherpress' ),
 							},
--- a/src/blocks/online-event-link/render.php
+++ b/src/blocks/online-event-link/render.php
@@ -64,7 +64,7 @@ $gatherpress_context_json = wp_json_encode(
 	<?php if ( $gatherpress_has_link ) : ?>
 		<a class="gatherpress-online-event__text" href="<?php echo esc_url( $gatherpress_online_event_link ); ?>" target="_blank" rel="noopener noreferrer">
 			<?php echo wp_kses_post( $gatherpress_link_text ); ?>
-			<span class="screen-reader-text"> <?php echo esc_html__( '(opens in a new tab)', 'gatherpress' ); ?></span>
+			<span class="screen-reader-text"> <?php esc_html_e( '(opens in a new tab)', 'gatherpress' ); ?></span>
 		</a>
 	<?php else : ?>
 		<span class="gatherpress-online-event__text">
--- a/src/blocks/online-event-link/render.php
+++ b/src/blocks/online-event-link/render.php
@@ -50,7 +50,7 @@ $gatherpress_context_json = wp_json_encode(
 	array(
 		'postId'        => $gatherpress_current_post_id,
 		'linkText'      => $gatherpress_link_text,
-		'newTabWarning' => __( '(opens in a new tab)', 'gatherpress' ),
+		'newTabWarning' => __( ' (opens in a new tab)', 'gatherpress' ),
 	),
 	JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP
 );
@@ -64,7 +64,7 @@ $gatherpress_context_json = wp_json_encode(
 	<?php if ( $gatherpress_has_link ) : ?>
 		<a class="gatherpress-online-event__text" href="<?php echo esc_url( $gatherpress_online_event_link ); ?>" target="_blank" rel="noopener noreferrer">
 			<?php echo wp_kses_post( $gatherpress_link_text ); ?>
-			<span class="screen-reader-text gatherpress-new-tab-notice"> <?php esc_html_e( '(opens in a new tab)', 'gatherpress' ); ?></span>
+			<span class="screen-reader-text gatherpress-new-tab-notice"><?php esc_html_e( ' (opens in a new tab)', 'gatherpress' ); ?></span>
 		</a>
 	<?php else : ?>
 		<span class="gatherpress-online-event__text">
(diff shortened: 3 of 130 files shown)


src/utility.scss
```diff
@@ -1,6 +1,30 @@
...
+// visually hidden announcements such as the "(opens in a new tab)"
+// warning on links that open a new tab, so screen readers announce it
+// while sighted users see no change. `clip` was removed in WordPress 6.8
+// (make.wordpress.org/core/2025/03/20) but is retained here alongside
+// `clip-path` and `word-wrap` for backwards compatibility with older
+// WordPress and themes that expect the legacy shape.
+/* stylelint-disable property-no-deprecated -- clip + word-wrap are
+   deprecated but retained for legacy compatibility; dropping them breaks
+   older themes and WordPress versions that expect the pre-6.8 shape. */
+.screen-reader-text {
```
mauteri: This is a core-owned class (`wp-includes/css/dist/block-library/common.css`), so redefining it here overrides the theme wherever our stylesheet loads last. Can we drop this block and the `stylelint-disable` and just let the markup pick up core's rule?

The pre-6.8 note doesn't apply either, since `clip` was removed in 6.8 and we require WP 7.0.

mauteri (changes requested):

src/blocks/online-event-link/render.php
```diff
@@ -63,6 +64,7 @@
 	<?php if ( $gatherpress_has_link ) : ?>
 		<a class="gatherpress-online-event__text" href="<?php echo esc_url( $gatherpress_online_event_link ); ?>" target="_blank" rel="noopener noreferrer">
 			<?php echo wp_kses_post( $gatherpress_link_text ); ?>
+			<span class="screen-reader-text"> <?php echo esc_html__( '(opens in a new tab)', 'gatherpress' ); ?></span>
```
mauteri: Use `esc_html_e` rather than `echo esc_html__`

mauteri (changes requested):

src/utility.scss
```diff
@@ -1,6 +1,30 @@
...
+	border: 0;
+	clip: rect(1px, 1px, 1px, 1px);
```
mattcowan: This clip is going to hide the theme's skip link while it has focus. WordPress dropped clip from its screen-reader-text pattern in 6.8, so core's :focus rules only reset clip-path now - nothing resets clip any more. Because this stylesheet is enqueued from render_block whenever any gatherpress/* block renders, the rule lands on every front-end page with a GatherPress block and applies to the theme's .screen-reader-text elements too, not just ours.

I compared the same page on both branches. On develop the skip link appears normally when you Tab to it. On this branch it takes focus but stays clipped to 1x1 and is not visible, which is a 2.4.7 failure. Screenshots below.

Dropping the clip line fixes it, and the rest of the block may be redundant anyway - core's wp-block-library stylesheet already defines .screen-reader-text and loads in wp_head, which is what was hiding these spans correctly before this rule existed

Current Develop screenshot:

Image

Current PR screenshot: 

Image

mattcowan (commented):
I am in an airport and have a few other lower-level findings I need to verify, but I wanted to post this comment I was able to verify real quick.

src/utility.scss
faisalahammad: You are right that this is core-owned. I deleted the whole .screen-reader-text rule from src/utility.scss and both stylelint disables. The markup still uses the class, and core wp-block-library CSS loaded in wp_head already hides it. The theme keeps its own definition.

src/utility.scss
faisalahammad: Confirmed the 2.4.7 failure. I deleted the entire .screen-reader-text block, so no clip rule reaches the theme skip link. The focused skip link now renders as on develop. Because core wp-block-library.css already ships the class from wp_head, the warning stays hidden without our rule. Thanks for verifying on both branches.

src/blocks/online-event-link/render.php
faisalahammad: Updated the span to esc_html_e. The newTabWarning value in the context payload stays esc_html__ because it is encoded with wp_json_encode, not echoed. I did not change the similar calls in venue-detail and venue-map since they sit inside sprintf/printf format strings, where esc_html_e would let void corrupt the output ordering.

faisalahammad:
Addressed the review feedback on this branch:

- Deleted the plugin-defined `.screen-reader-text` rule from `src/utility.scss` (and both stylelint disables). It was core-owned and the `clip` property kept a focused theme skip link clipped to 1x1px (WCAG 2.4.7). The visually-hidden class is provided by core via wp-block-library, so the new-tab warning stays hidden without the plugin rule.
- Switched the online event link new-tab warning span to `esc_html_e`. The `newTabWarning` context payload stays `esc_html__` because it is passed through `wp_json_encode`.
- Left the similar calls in venue-detail and venue-map as `esc_html__`: they sit inside `sprintf`/`printf` format strings, where `esc_html_e` would print early and corrupt the output order.

Lint (php, phpstan, css), JS unit tests (1008 pass), and CodeRabbit are clean. Please re-review when you have a chance.

src/blocks/add-to-calendar/template.js
```diff
@@ -40,7 +40,7 @@ const TEMPLATE = [
 					[
 						'gatherpress/dropdown-item',
 						{
-							text: `<a href="#gatherpress-google-calendar" rel="noreferrer noopener nofollow" target="_blank">${ __( 'Google Calendar', 'gatherpress' ) }</a>`,
+							text: `<a href="#gatherpress-google-calendar" rel="noreferrer noopener nofollow" target="_blank">${ __( 'Google Calendar', 'gatherpress' ) }<span class="screen-reader-text"> ${ __( '(opens in a new tab)', 'gatherpress' ) }</span></a>`,
```
mattcowan: Here, and on line 70: The dropdown-item pipeline is text-only, so this span turns into visible text. dropdown-item/edit.js:85 does parsedDoc.body.textContent.trim(), which strips the markup and keeps the words. I ran that exact transformation on this string and got "Google Calendar (opens in a new tab)" - so the first time anyone edits the item, the warning is baked into the visible label permanently. The inner blocks are not locked, so an edit is a normal thing for an author to do.

Reproduced in the editor: insert the block, then type a character into the Google Calendar item and delete it again - a no-op edit. The item's markup goes from Google Calendar<span class="screen-reader-text"> (opens in a new tab)</span> to plain Google Calendar (opens in a new tab), and the label, the sidebar block title and the breadcrumb all show it. The Yahoo item, untouched, keeps its span - so it is the edit that does this, not anything global. Screenshots below.

Worth noting the same applies to any dropdown with actAsSelect turned on, since dropdown/view.js:86 copies textContent into the trigger. Add-to-calendar defaults to false, so that path is not live here.

Separately, this is a template change, so it only affects newly inserted blocks. I loaded an existing event and its calendar links still render without the warning. The changelog entry says the fix covers "the Google and Yahoo calendar options", which will not be true for existing content.

Screenshot before edit:  

Image

Screenshot after edit: 

Image

src/blocks/online-event-link/render.php
```diff
@@ -63,6 +64,7 @@
 	<?php if ( $gatherpress_has_link ) : ?>
 		<a class="gatherpress-online-event__text" href="<?php echo esc_url( $gatherpress_online_event_link ); ?>" target="_blank" rel="noopener noreferrer">
 			<?php echo wp_kses_post( $gatherpress_link_text ); ?>
+			<span class="screen-reader-text"> <?php esc_html_e( '(opens in a new tab)', 'gatherpress' ); ?></span>
```
mattcowan: This span needs the gatherpress-new-tab-notice class that view.js looks for. As it stands view.js:57 removes .gatherpress-new-tab-notice, which never matches the server-rendered span, so the span survives every swap.

I replayed the two branches on this exact markup. When the link is removed, the warning stays attached to the plain text: "Join online (opens in a new tab)" on a <span> that is no longer a link. When the link stays and the href changes, hadWarning is false so a second span is appended: "Join online (opens in a new tab) (opens in a new tab)".

Adding the class to this span fixes both.

Stranded span case:

Image

Duplicate on link case:

Image

Once it matches, hadWarning at view.js:98 is always false anyway, because the unconditional remove() above it has already stripped the span - so that branch could just append unconditionally.

src/blocks/online-event-link/render.php
```diff
@@ -48,8 +48,9 @@
 
 $gatherpress_context_json = wp_json_encode(
 	array(
-		'postId'   => $gatherpress_current_post_id,
-		'linkText' => $gatherpress_link_text,
+		'postId'        => $gatherpress_current_post_id,
+		'linkText'      => $gatherpress_link_text,
+		'newTabWarning' => esc_html__( '(opens in a new tab)', 'gatherpress' ),
```
mattcowan: esc_html__() here escapes into a JSON payload, and view.js:108 assigns it with textContent, which does not decode entities. Any locale whose translation contains an apostrophe announces the entity. I ran the French string through it: (s'ouvre dans un nouvel onglet) becomes (s&#039;ouvre dans un nouvel onglet) and survives wp_json_encode intact. Plain __() is right here - wp_json_encode with the JSON_HEX_* flags already handles the escaping for this context. English is unaffected, which is why it may not have shown up in initial testing.

Image

mattcowan (commented):
@faisalahammad thank you for uploading the previous fixes. I did find 3 more issues, sorry I was not able to post these sooner.

src/blocks/online-event-link/render.php
faisalahammad: Fixed. newTabWarning now uses __(). wp_json_encode still escapes the JSON. The echoed span still uses esc_html_e.

src/blocks/add-to-calendar/template.js
faisalahammad: I removed the span from the template. inject_new_tab_notices now adds the warning on render for every _blank calendar link. Existing events get the warning. Editor labels stay as Google Calendar and Yahoo Calendar.

src/blocks/online-event-link/render.php
faisalahammad: Server span now has gatherpress-new-tab-notice. view.js removes that class first, then appends the span again when the link stays a link. No stranded warning. No duplicate.

includes/core/classes/blocks/class-add-to-calendar.php
```diff
@@ -121,4 +122,38 @@ public function replace_calendar_placeholders( string $block_content, array $blo
...
+	 * @param string $block_content The block content to parse.
+	 *
+	 * @return string The block content with new-tab notices injected.
+	 */
+	public function inject_new_tab_notices( string $block_content ): string {
+		return (string) preg_replace_callback(
+			'/(<a\b[^>]*?target=["\']_blank["\'][^>]*>)(.*?)(<\/a>)/is',
+			static function ( array $matches ): string {
+				if ( str_contains( $matches[0], 'gatherpress-new-tab-notice' ) ) {
+					return $matches[0];
```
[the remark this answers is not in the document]
faisalahammad: Changed the idempotency check to only match the exact class attribute this plugin injects. A calendar URL, visible link text, or nested non-class attribute carrying the marker text no longer suppresses the warning. Added regression tests for the marker in the href and in the visible link text.

includes/core/classes/blocks/class-add-to-calendar.php
```diff
@@ -65,6 +65,7 @@ protected function setup_hooks(): void {
 		$render_block_hook = sprintf( 'render_block_%s', self::BLOCK_NAME );
 
 		add_filter( $render_block_hook, array( $this, 'replace_calendar_placeholders' ), 10, 2 );
+		add_filter( $render_block_hook, array( $this, 'inject_new_tab_notices' ), 11 );
```
mauteri, line 71: Please leave comment why this is set to priority `11` so next developer knows why.

includes/core/classes/blocks/class-add-to-calendar.php
```diff
@@ -121,4 +122,48 @@ public function replace_calendar_placeholders( string $block_content, array $blo
...
+	 * marker class are left untouched, keeping the transform idempotent when
+	 * the filter runs more than once against the same content.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @param string $block_content The block content to parse.
+	 *
+	 * @return string The block content with new-tab notices injected.
+	 */
+	public function inject_new_tab_notices( string $block_content ): string {
```
mauteri, line 155: Can we rework this with `WP_HTML_Tag_Processor` instead of a regex? The pattern here has a few failure modes:

- `[^>]*?target=["\']_blank["\']` matches the string anywhere in the tag, so `data-x='target="_blank"'` false-positives, and unquoted target=_blank (valid HTML) is missed.
- The idempotence check is an exact string match on `class="screen-reader-text gatherpress-new-tab-notice"`. Single quotes or a reordered class list defeat it and you get duplicate notices.\
- `(.*?)(<\/a>)` pairs with the first </a> anywhere, including inside a comment or attribute value.

The Tag Processor can't insert markup (attributes and plain text only, and WP_HTML_Processor on our WP floor has no public insertion API either), so a full swap isn't possible. But we can make the parser own every decision and leave only a mechanical splice:

```
public function inject_new_tab_notices( string $block_content ): string {
	$processor = new WP_HTML_Tag_Processor( $block_content );
	$inside    = false;

	// First pass: the parser makes every decision, recorded as temporary
	// markers written with known, uniform quoting.
	while ( $processor->next_tag( array( 'tag_closers' => 'visit' ) ) ) {
		if ( 'A' === $processor->get_tag() ) {
			if ( $processor->is_tag_closer() ) {
				$inside = false;
			} elseif ( '_blank' === $processor->get_attribute( 'target' ) ) {
				$inside = true;
				$processor->set_attribute( 'data-gatherpress-new-tab', '1' );
			}
			continue;
		}

		if (
			$inside
			&& 'SPAN' === $processor->get_tag()
			&& ! $processor->is_tag_closer()
			&& $processor->has_class( 'gatherpress-new-tab-notice' )
		) {
			$processor->set_attribute( 'data-gatherpress-has-notice', '1' );
		}
	}

	$html = $processor->get_updated_html();

	// Second pass: pure splicing on the markers the parser wrote. Anchors
	// cannot nest, so the next </a> after a marked opener closes it.
	$marker = ' data-gatherpress-new-tab="1"';
	$notice = sprintf(
		'<span class="screen-reader-text gatherpress-new-tab-notice"> %s</span>',
		esc_html__( '(opens in a new tab)', 'gatherpress' )
	);

	$offset = 0;
	while ( false !== ( $position = strpos( $html, $marker, $offset ) ) ) {
		$close = stripos( $html, '</a>', $position );

		if ( false !== $close ) {
			$range = substr( $html, $position, $close - $position );

			if ( ! str_contains( $range, 'data-gatherpress-has-notice' ) ) {
				$html = substr_replace( $html, $notice, $close, 0 );
			}
		}

		$offset = $position + 1;
	}

	return str_replace(
		array( $marker, ' data-gatherpress-has-notice="1"' ),
		'',
		$html
	);
}
```

test/unit/js/src/blocks/online-event-link/view.test.js
```diff
@@ -0,0 +1,225 @@
...
+	let callbacks;
+
+	beforeEach( () => {
+		( { state, callbacks } = store( 'gatherpress' ) );
+
+		// Reset the shared registry state between tests.
+		delete state.posts;
+	} );
+
+	/**
```
mauteri, line 69: Need to fix below too.

mauteri (changes requested):

mauteri:
Also need to add any missing code coverage. Make sure all workflows are green. Thx!

mattcowan (dismissed):
This approval is just for comments I had made: all four earlier findings from me are fixed and verified. The calendar fix is better than what I suggested.  Moving it to a render filter means existing events get the warning too, which the template approach couldn't do. The added test coverage is nice as well. Thank you for the good work here.

includes/core/classes/blocks/class-add-to-calendar.php
faisalahammad: Reworked. `inject_new_tab_notices` now uses a `WP_HTML_Tag_Processor` first pass that owns every decision and records it as temporary `data-*` markers, then a second pass splices the notice before each marked anchor's `</a>`. The target check reads the real attribute, so unquoted `target=_blank` matches and a `data-x='target="_blank"'` no longer false-positives. The existing-notice check runs through `has_class`, so single quotes or a reordered class list still suppress a duplicate.

Added regression tests for the unquoted target, the mimic-in-other-attribute case, and a reordered/single-quoted marker class. Full PHPUnit suite passes (1775 tests, 7203 assertions), PHPCS and PHPStan clean.

includes/core/classes/blocks/class-add-to-calendar.php
faisalahammad: Added a comment above the filter explaining priority 11: the notice is injected after `replace_calendar_placeholders` (priority 10) has built the final per-service hrefs onto the anchors, and after any third-party render filter at the default priority.

test/unit/js/src/blocks/online-event-link/view.test.js
faisalahammad: JSDoc alignment on setupElement was already fixed in d597f260. Extra spaces after {string} are gone.

src/blocks/online-event-link/render.php
```diff
@@ -63,6 +64,7 @@
 	<?php if ( $gatherpress_has_link ) : ?>
 		<a class="gatherpress-online-event__text" href="<?php echo esc_url( $gatherpress_online_event_link ); ?>" target="_blank" rel="noopener noreferrer">
 			<?php echo wp_kses_post( $gatherpress_link_text ); ?>
+			<span class="screen-reader-text gatherpress-new-tab-notice"> <?php esc_html_e( '(opens in a new tab)', 'gatherpress' ); ?></span>
```
mauteri: remove space before `<?php` and apply to translation, ie, `' (opens in...`

src/blocks/online-event-link/view.js
```diff
@@ -68,15 +77,34 @@ const { state } = store( 'gatherpress', {
...
 				linkElement.target = '_blank';
 				linkElement.rel = 'noopener noreferrer';
-				linkElement.innerHTML = currentHTML;
+				// Mirror the server-rendered markup in render.php: the
+				// screen-reader warning must survive the client-side swap.
+				// The string travels in the context payload because script
+				// modules cannot import @wordpress/i18n.
+				linkElement.innerHTML =
+					currentHTML +
+					`<span class="screen-reader-text gatherpress-new-tab-notice"> ${ context?.newTabWarning ?? '' }</span>`;
```
mauteri: Remove space before `${`. Space should be applied to `newTabWarning`.

src/blocks/online-event-link/view.js
```diff
@@ -68,15 +77,34 @@ const { state } = store( 'gatherpress', {
...
+				// dropped above, so appending unconditionally is
+				// idempotent.
+				if ( currentElement.href !== onlineEventLink ) {
+					currentElement.href = onlineEventLink;
+				}
+
+				if ( context?.newTabWarning ) {
+					const sr = document.createElement( 'span' );
+					sr.className = 'screen-reader-text gatherpress-new-tab-notice';
+					sr.textContent = ' ' + context.newTabWarning;
```
mauteri: Space should be applied to `newTabWarning`

src/blocks/venue-detail/render.php
```diff
@@ -72,12 +72,18 @@
 
 		$gatherpress_target_attr = '_blank' === $gatherpress_link_target ? ' target="_blank" rel="noopener noreferrer"' : '';
 
+		// Warn screen-reader users before the context change of a new tab.
+		$gatherpress_sr_text = '_blank' === $gatherpress_link_target
+			? sprintf( '<span class="screen-reader-text"> %s</span>', esc_html__( '(opens in a new tab)', 'gatherpress' ) )
```
mauteri: Space should be applied to translation

src/blocks/venue-map/render.php
```diff
@@ -268,6 +268,16 @@
 		);
 	}
 
+	// Warn screen-reader users before the context change of a new tab. It
+	// follows the image so the link's accessible name reads "map, opens in
+	// a new tab" rather than leading with the warning.
+	if ( '' !== $gatherpress_href && '_blank' === $gatherpress_target ) {
+		printf(
+			'<span class="screen-reader-text"> %s</span>',
```
mauteri: Space should be applied to translation.

test/unit/js/src/blocks/online-event-link/view.test.js
```diff
@@ -0,0 +1,260 @@
...
+ * Internal dependencies
+ */
+import '@src/blocks/online-event-link/view';
+
+/**
+ * The screen-reader warning this block injects on new-tab links. Mirrors the
+ * context payload from `render.php` so the client-side swap behaves like the
+ * server render.
+ */
+const NEW_TAB_WARNING = '(opens in a new tab)';
```
mauteri: apply space here. fix here and everywhere where it applies.

mauteri (changes requested):

src/blocks/online-event-link/render.php
faisalahammad: Done. Space moved into the translation string.

src/blocks/online-event-link/view.js
faisalahammad: Done. Space now travels inside the newTabWarning value from PHP, not in the JS template.

src/blocks/online-event-link/view.js
faisalahammad: Done. sr.textContent uses the value as is. The space comes from the translation.

src/blocks/venue-detail/render.php
faisalahammad: Done. Space moved into the esc_html__ string.

src/blocks/venue-map/render.php
faisalahammad: Done. Space moved into the esc_html__ string.

test/unit/js/src/blocks/online-event-link/view.test.js
faisalahammad: Done. Constant and fixtures now carry the leading space, markup no longer adds it.

faisalahammad:
Addressed the spacing feedback on this branch.

Moved the leading space into the translated string at every emit site, so the warning no longer depends on markup for the separator:

- online-event-link render.php (context payload + server span)
- online-event-link view.js (template literal + appendChild)
- venue-detail render.php
- venue-map render.php
- add-to-calendar inject_new_tab_notices
- view.test.js constant and fixtures

Rendered HTML output is unchanged. Lint (php, phpstan, js, css), Jest tests, and build pass. All six review threads resolved.

@mauteri please re-review when you have a chance.

mauteri:
@faisalahammad please fix PHPStan issues. Make sure all workflows are green before requesting a review. Thx!

mauteri (changes requested):
Requesting changes. See previous comment.

faisalahammad:
## CI Fix Summary — 3 failures resolved

| # | File | Error | Fix |
|---|------|-------|-----|
| 1 | includes/core/classes/event/class-abilities.php:174 | iterable `array` return missing value type | added `@phpstan-type EventUpcomingEvent` alias; narrowed `@return` to `array<int, EventUpcomingEvent>` |
| 2 | includes/core/classes/event/class-abilities.php:181 | `Cannot cast int\|WP_Post to int` | replaced `(int) ` with `is_int(  )\continue` guard |
| 3 | includes/core/classes/rsvp/class-abilities.php:185 | `Offset 'count' on left of ?? always exists and is not nullable` | dropped redundant `?? 0`; `RsvpResponseGroup` shape guarantees `count: int` |

Tests ✅ · Verification ✅ · CodeRabbit ✅ · PHP 8.1 ✅

mauteri:
@faisalahammad see coverage comment above for missing PHPUnit test coverage. Thx.

faisalahammad:
@mauteri both blockers are fixed, and they were the same root cause.

Commit `d96505ea` added an `is_int()` guard in `includes/core/classes/event/class-abilities.php` to clear PHPStan. Develop then fixed the same errors independently with a `@var int[]` annotation. That collision caused both failures: the guard's `continue` branch is unreachable (`fields => ids` always returns `int[]`), which failed the coverage gate at line 184, and the merge conflict was in that same file.

I merged develop and took its version. No runtime branch, so no line left to cover. `rsvp/class-abilities.php` needed no resolution since develop already has the identical change. Both files now match develop and drop out of the diff, taking the PR from 10 changed files to 8. No accessibility code changed.

Green locally and in CI, including the coverage gate: `class-add-to-calendar.php` at 100% (65/65), now the only changed file needing a check.

One thing to flag. CodeRabbit raised a minor point on `inject_new_tab_notices`: the closing `str_replace` removes `data-gatherpress-new-tab` globally, so markup already carrying that attribute would lose it. That cleanup is the implementation you supplied, and reaching the case needs an author to hand-add the attribute inside the block, so I left your design alone. Say the word and I will harden it with a per-invocation marker plus a regression test.

Its other three findings ask for render tests on `online-event-link`, `venue-detail` and `venue-map`, which the coverage config excludes, so I skipped those.

All review threads are resolved. Please re-review when you get a chance.

mauteri:
Thanks for this @faisalahammad, and for working through the review rounds on it.

I've opened #2258 as an alternative and I'd like to take that direction, so I want to lay out the reasoning rather than just swap one for the other.

The behavior in this PR is right. What I want to change is where it lives. This one adds the markup in each block that needs it: three render templates, plus a dedicated filter on add-to-calendar. That works today. The cost is that every block we add later with an outbound link has to remember to do the same thing, and nothing tells us when one doesn't.

#2258 does it in a single `render_block` filter for anything under the `gatherpress/` prefix. No block participates, and blocks we add later are covered without anyone thinking about it. It isn't smaller, roughly 250 lines of implementation against 130 here. The trade is that the per-block work drops to zero.

Two things your PR surfaced that carried straight over. The online event link builds its anchor in the browser, so a PHP filter alone can't reach it and a small script is still needed. And a link can already contain screen-reader text from the tooltip format, which an "is it already announced" check will trip over unless the notice carries its own class.

I'll leave this open for now and close it once #2258 is approved and merged.

mauteri:
Thanks for this, @faisalahammad. #2141 shipped in #2258, so this is superseded rather than rejected.

Same fix, different placement: yours adds the notice per block, #2258 puts it behind one `render_block` filter so blocks added later are covered without touching them.

Two things you had right that #2258 reached the long way round: visiting tag closers to track anchor state, and clearing the stale notice when the online event link swaps back to plain text.

You're on the props line in #2258.

PR #2203, opened 2026-08-21
Add i18n src-to-build translation mapping

### LINKED ISSUE #2011: Add i18n mapping from src to build files

The WP Glotpress at https://translate.wordpress.org/projects/wp-plugins/gatherpress shows the source for each string. It does work well for all php classes, but not for build files like render.php or any JS coming from a block src.

The current behaviour shows only the build file, which is almost useless, as it is unreadable.

Image

Image

### Designs

"GatherPress should add a mapping list to the i18n command" would be my recommendation, if GatherPress would handle its own `/languages` folder. But due to the fact, that all translations are handled on wp.org, I'm not quite sure how to get this done nicely.

I usually have sth. like this in `package.json`:

```json
"i18n:json:map": "wp i18n make-json languages --pretty-print --no-purge '--use-map={ \"src/index.js\": \"build/index.js\", \"src/save.js\": \"build/index.js\", \"src/edit.js\": \"build/index.js\", \"src/block.json\": \"build/block.json\" }'"

```

_No response_

### Code of Conduct

faisalahammad:
Fixes #2011

## Proposed changes:

Adds i18n tooling so every release zip ships a translation template with readable source references and correctly named JSON translation files.

- New `npm run i18n` script generates `languages/gatherpress.pot` with references pointing at `src/` files (build output is excluded from extraction), so translators see real source lines instead of minified build paths.
- New `.github/scripts/i18n/build-map.js` derives a src-to-build path map automatically from `webpack.config.js` entries plus a static import walk. Nothing is maintained by hand; adding a block or component updates the map on the next run.
- `wp i18n make-json --use-map` then splits translations into per-script JSON files whose names hash-match the actual `build/` script paths WordPress looks up at runtime in `load_script_translation_file()`.
- `npm run plugin-zip` runs the i18n step, and the wp.org deploy jobs in `release.yml` generate the pot before rsync so SVN trunk (and GlotPress) sees it. WP-CLI is installed in those jobs since `wp` is not on the runner image.
- New `Setup::load_textdomain()` hooked on `init` registers the plugin's own `languages/` directory, so self-hosted installs (GitHub zip) and tools like Loco Translate can find the shipped template and locally generated translations.
- The intermediate `languages/map.json` is excluded from both distribution paths (`package.json` files field, `.distignore`), and `languages/` is gitignored like `build/`.

One honest limit: this cannot change what translate.wordpress.org itself indexes from the deployed SVN code. What it does fix is every self-hosted and local translator path, and it ships a correct src-referenced POT with each zip.

### Other information:

New tests in `class-test-setup.php` cover the `init` hook registration and the textdomain registry path.

## Testing instructions:

1. Run `npm run i18n`.
2. Result: no errors. `Success:` from make-json may say `Created 0 files.` while only the `.pot` template exists; that is expected. JSON files appear per locale once translated `.po` files are present.
3. Run `grep -c "src/" languages/gatherpress.pot` and `grep -c "build/" languages/gatherpress.pot`.
4. Result: hundreds of src references, zero build references.
5. Run `node .github/scripts/i18n/build-map.js | jq 'keys | length'`.
6. Result: 187 keys, shared components list every bundle that includes them.
7. Optional, proves the hash naming: copy the pot to a `de_DE.po`, add a msgstr, run `wp i18n make-json /tmp/gatherpress-de_DE.po /tmp/jsonout --use-map=languages/map.json --no-purge`, then compare `php -r "echo md5('build/panels.js');"` against the generated filename.
8. Run `npm run plugin-zip` and `unzip -l gatherpress.zip | grep languages`.
9. Result: `gatherpress/languages/gatherpress.pot` present, no `src/`, no `.github/scripts/i18n/`, no `map.json`.
10. On a test site, switch the site language and load an admin page.
11. Result: no fatals, strings fall back to English since no translations exist yet.

### Changelog entry

Changelog entry added manually with `composer changelog:add` at `.github/changelog/2011-i18n-src-to-build-map` (Significance: minor, Type: added), since the workflow job fails for fork PRs.

<details>

#### Significance

-   [x] Minor

#### Type

#### Message

Ship a src-referenced translation template and correctly named JSON translation files with each release zip, so translators see readable source lines instead of minified build paths.

</details>

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -66,6 +66,9 @@ docs
 wp-cron.sh
 patches
 
+# Generated translations (npm run i18n); only the pot ships.
+languages/map.json
+
 # Miscellaneous and SQL Archives
 behat.yml
 bitbucket-pipelines.yml
--- a/.github/changelog/2011-i18n-src-to-build-map
+++ b/.github/changelog/2011-i18n-src-to-build-map
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Ship a src-referenced translation template and correctly named JSON translation files with each release zip, so translators see readable source lines instead of minified build paths.
--- a/.github/scripts/i18n/build-map.js
+++ b/.github/scripts/i18n/build-map.js
@@ -1,0 +1,291 @@
+/**
+ * External dependencies
+ */
+const fs = require( 'fs' );
+const path = require( 'path' );
+
+// The project webpack config spreads the default config's entries, and the
+// default only exports the two-config array when the modules flag is set
+// (the same flag `npm run build` passes). Setting it here lets this script
+// require the real config instead of duplicating its entry list.
+process.env.WP_EXPERIMENTAL_MODULES = '1';
+
+/**
+ * WordPress dependencies
+ */
+const projectConfig = require( path.resolve( process.cwd(), 'webpack.config.js' ) );
+
+/**
+ * Generates the src-to-build file map used by `wp i18n make-json --use-map`.
+ *
+ * The translation POT is extracted from `src/` so its references stay readable,
+ * but WordPress loads translations for `build/` script paths at runtime. This
+ * script bridges the two by deriving the mapping from the same webpack entry
+ * points the build uses, then following static imports to find every bundled
+ * source file, so it never has to be maintained by hand.
+ *
+ * Output goes to stdout as a JSON object keyed by source path. Values are
+ * arrays because webpack bundles shared modules into every entry that imports
+ * them, and `wp i18n make-json` emits one Jed JSON per build path:
+ * { "src/blocks/rsvp/edit.js": [ "build/blocks/rsvp/index.js" ] }.
+ *
+ * @since 0.36.0
+ */
+
+/**
+ * Collects the entry points of the classic webpack config.
+ *
+ * Mirrors how `npm run build` resolves entries: the project's webpack.config.js
+ * spreads the default `script` config's entries and adds its own.
+ *
+ * @since 0.36.0
+ *
+ * @return {Object<string, string>} Entry names mapped to absolute source paths.
+ */
+function getEntries() {
+	const configs = Array.isArray( projectConfig )
+		? projectConfig
+		: [ projectConfig ];
+
+	const scriptConfig =
+		configs.find(
+			( config ) => ! config.experiments?.outputModule
+		) ?? configs[ 0 ];
+
+	return typeof scriptConfig.entry === 'function'
+		? scriptConfig.entry()
+		: scriptConfig.entry;
+}
+
+/**
+ * Collects PHP render files copied verbatim from src to build.
+ *
+ * The block build copies `render.php` next to each block's build output without
+ * touching the code inside, so a translation reference in `src/.../render.php`
+ * resolves to the identical file under `build/`.
+ *
+ * @since 0.36.0
+ *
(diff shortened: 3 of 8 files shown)


carstingaxion:
@faisalahammad I've had this PR open now for a couple of times, thinking through it more and more. Just to let you know.

> One honest limit: this cannot change what translate.wordpress.org itself indexes from the deployed SVN code.

This limit is unfortunately exactly and 100% the need, so I'm still looking for a working idea besides or based on your PR. I asked in core polyglots slack channel today about this particular topic. Feel free to chime in at https://wordpress.slack.com/archives/C02RP50LK/p1788262223144529

carstingaxion:
OK, this needs a cleanup-decision now:

1. With the help of @wppoland (in the [linked polyglots-chat](https://wordpress.slack.com/archives/C02RP50LK/p1788262223144529)) I came to the understanding, that my initial question was asking for an impossible thing.
2. Because the `/src` folders are not part of SVN trunk (and will never be), GlotPress would have nothing to show "mapped".
3. I feel bad for that oversight and am sorry for every "useless" work being done

For that PR here, I would suggest closing it at all, because the major idea can never be done.

The tooling to create the mapping itself is interesting and maybe worth keeping as sep. repo for other to use. But keeping it in GatherPress without a clear job, seems to be not ok.

What do you think @mauteri ?

faisalahammad:
Thanks for checking with polyglots, carstingaxion. Good catch, src/ never reaches SVN trunk so GlotPress cannot use it. Agree that goal does not work.

One note for context: the rest of the PR does not depend on that goal. load_textdomain() fixes translations not loading at all on self-hosted and GitHub zip installs. build-map.js still produces correctly hash-named JSON files for the build output WordPress loads at runtime. Both work regardless of the GlotPress question.

Leaving the keep-vs-close call to the maintainers.

mauteri:
Thanks for the work on this, @faisalahammad, and thanks @carstingaxion for running it down with polyglots. I read through the thread and the diff and I'm going to close this one.

The original goal is off the table, as you both concluded: GlotPress reads the deployed SVN code, `src/` never reaches SVN, so there is nothing for a mapping to point at. No fault on anyone, it just took the polyglots conversation to see it.

On the parts that were pitched as standing on their own, I looked at what each does in our distribution model and I don't think either has a job here:

- `load_textdomain()` registers the plugin's own `languages/` folder, but that folder ships only the `.pot`. There are no `.mo` files in it to load, on a GitHub zip or anywhere else. wp.org language packs install to `wp-content/languages/plugins/` by slug however the plugin was installed, and core has loaded those without a `load_plugin_textdomain()` call since 4.6. That's why `develop` has never had one and translations work today.
- The `build-map.js` and hash-named JSON output only matter if we ship our own translation JSON, and we don't. wp.org generates the JS language packs from the deployed `build/` files with the right hashes already. So `map.json` is produced and then excluded from the zip with nothing consuming it.

What the PR does add is WP-CLI installation and an i18n step in both wp.org deploy jobs, to put a `.pot` into trunk that GlotPress doesn't use. That's a new place for a release to fail in exchange for nothing, and we've been working to keep that workflow boring.

The mapping tool itself is a nice piece of work. I'd second Carsten's suggestion of publishing it as its own repo, where it can help plugins that do ship their own translations. Happy to link to it from our contributor docs if you do.

You'll get props on #2011 for this regardless.

PR #2205, opened 2026-08-22
Add filter for shadow-source queries by event activity

### LINKED ISSUE #1906: Filter shadow-source queries by upcoming / past event activity
## Summary

When a `core/query` block is configured to list posts of a shadow-source post type (venues, productions, seasons, …), there is currently no way to filter that list by whether those posts have **upcoming** or **past** events attached to them. This issue proposes two nested toggles in the Query Loop sidebar that enable exactly that.

---

## Motivation

The existing `ShadowSourceFilterControls` toggle works in the **opposite direction**: you are on a shadow-source page (e.g. a Venue single) and you filter an *event* query down to only the events tagged with that venue's shadow term.

The missing direction is: you are listing *shadow-source posts themselves* (e.g. "all venues", "all productions") and you want to filter that list to only the venues / productions that have at least one upcoming event — or conversely, only the ones that are "closed" (all events are in the past).

Concrete use-cases:
- "Currently playing" production listing — only productions whose next performance has not yet happened.
- "Active venues" widget — venues that have at least one upcoming event.
- "Archive" view — productions / seasons with no future events remaining.

## Proposed UX

Two toggles, rendered in the existing GatherPress sidebar panel of the `core/query` block, below the other controls:

```
┌─────────────────────────────────────────┐
│ ☑  Filter by event activity             │  ← Toggle 1 (primary)
│    Only show [Productions] that have    │
│    events linked to them.               │
│                                         │
│   ☐  Upcoming events only              │  ← Toggle 2 (secondary, visible only
│      Hides [Productions] whose events  │    when Toggle 1 is on)
│      have all already ended.            │
└─────────────────────────────────────────┘
```

**Toggle 1 — "Filter by event activity"** (`has_events_filter: 1 | 0`)
- Visible when **all** of the following are true:
  1. `query.postType` is a post type that declares `gatherpress-shadow-source` support.
  2. The shadow taxonomy for that post type (`_<post_type>`) is registered on `gatherpress_event` (i.e. is a `gatherpress_shadow_taxonomy_object_types` member, exactly as `_gatherpress_venue` is wired onto events today).
- When enabled: the query is limited to shadow-source posts that have **at least one** event term in their shadow taxonomy — regardless of timing.
- Labels use `usePostTypeLabel` so a productions archive reads *"Filter by event activity — Only show Productions that have events linked to them."*

**Toggle 2 — "Upcoming events only"** (`upcoming_events_only: 1 | 0`)
- Visible only when Toggle 1 is `on`.
- Default: `on` (the most common case is "active / currently running").
- When `on`: restrict further to shadow-source posts that have at least one **upcoming** event term (i.e. an event whose `datetime_end` is in the future).
- When `off`: restrict to shadow-source posts whose event terms point **only to past events** — the "archive" / "closed" view.
- Labels adapt via `usePostTypeLabel`: *"Upcoming events only — Hides Productions whose events have all already ended."* / *"Past events only — Shows Productions whose events have all already ended."*

---

### JavaScript — visibility gate

The gate mirrors `showShadowSourceFilterControl` in `EventQueryControlsSlotFill`, but checks the **queried** post type (not the editor host):

```js
// In EventQueryControlsSlotFill, alongside the existing gates:
const queryPostTypeSupports = useSelect(
    ( wpSelect ) =>
        queryPostType
            ? wpSelect( 'core' ).getPostType( queryPostType )?.supports
            : null,
    [ queryPostType ]
);

const queryPostTypeIsShadowSource =
    !! queryPostTypeSupports?.[ 'gatherpress-shadow-source' ];

// Additionally verify the shadow taxonomy is wired to events.
// The REST endpoint (or a new gatherpress_shadow_event_types endpoint)
// could expose which shadow taxonomies are attached to gatherpress_event.
// For a first pass, `queryPostTypeIsShadowSource` alone is sufficient.

const showHasEventsFilter = queryPostTypeIsShadowSource;
```

Two new query attributes written by the toggles:

```js
// Toggle 1
has_events_filter: value ? 1 : 0

// Toggle 2 (only meaningful when Toggle 1 is on)
upcoming_events_only: value ? 1 : 0   // default 1
```

A new `HasEventsFilterControls` component, following the exact pattern of `ShadowSourceFilterControls` and `EventListTypeControls`, renders both toggles.

### PHP — REST collection params

In `Event_Query::rest_collection_params()`, register the two new params:

```php
$query_params['has_events_filter'] = [
    'description' => __( 'Filter shadow-source posts by event activity.', 'gatherpress' ),
    'type'        => 'integer',
    'enum'        => [ 0, 1 ],
];

$query_params['upcoming_events_only'] = [
    'description' => __( 'When filtering by event activity, limit to posts with upcoming events.', 'gatherpress' ),
    'type'        => 'integer',
    'enum'        => [ 0, 1 ],
    'default'     => 1,
];
```

Pass them through in `rest_query()` and `query_loop_block_query_vars()`, exactly as `shadow_filter` is handled today.

### PHP — query logic (`pre_get_posts` / `Shadow_Source`)

The heavy lifting belongs in a new method on `Shadow_Source` (or in a `pre_get_posts` callback in a new `Shadow_Source_Query` class), invoked when `has_events_filter = 1`:

**Algorithm:**

1. Determine the shadow taxonomy for the queried post type:
   `$shadow_taxonomy = '_' . $queried_post_type;` — e.g. `_gatherpress_venue`, `_gatherpress_play`.

2. Verify the shadow taxonomy is registered on `gatherpress_event` via `get_taxonomy( $shadow_taxonomy )->object_type`. If not, bail — the filter cannot be applied.

3. Build a sub-query for events that are *upcoming* or *past* depending on `upcoming_events_only`:

```php
$event_query_type = $upcoming_events_only ? 'upcoming' : 'past';

$event_query = new WP_Query( [
    'post_type'               => 'gatherpress_event', // or all gatherpress-event-date types
    'posts_per_page'          => -1,
    'fields'                  => 'ids',
    'no_found_rows'           => true,
    'update_post_meta_cache'  => false,
    'update_post_term_cache'  => true,
    'gatherpress_event_query' => $event_query_type,
] );
```

4. Collect the shadow taxonomy term slugs for those event IDs:

```php
$term_slugs = [];
foreach ( $event_query->posts as $event_id ) {
    $terms = wp_get_object_terms( $event_id, $shadow_taxonomy, [ 'fields' => 'slugs' ] );
    $term_slugs = array_merge( $term_slugs, $terms );
}
$term_slugs = array_unique( $term_slugs );
```

5. Resolve term slugs back to shadow-source post slugs (strip leading `_`) and inject a `post_name__in` (or a `tax_query` against a separate helper taxonomy) to constrain the outer query to only those posts.

> **Note:** The round-trip through term slugs → post slugs can be avoided if `Shadow_Source` exposes a helper like `post_ids_from_shadow_terms( string $taxonomy, array $term_slugs ): int[]` that does the reverse of `term_slug_from_post_name()`. That helper would be useful beyond this feature (e.g. for REST endpoints listing "venues with upcoming events").

---

## Visibility rules, summarised

| Editor context | Toggle 1 visible? | Toggle 2 visible? |
|---|---|---|
| Querying a `gatherpress-shadow-source` post type whose shadow taxonomy is wired to events | ✅ | Only when Toggle 1 is on |
| Querying `gatherpress_event` or any event-date post type | ❌ | ❌ |
| Querying any other post type | ❌ | ❌ |
| FSE template / template part (no concrete `queryPostType`) | ❌ | ❌ |

---

## Relationship to existing features

| Existing feature | Direction | This issue |
|---|---|---|
| `ShadowSourceFilterControls` (`shadow_filter`) | Querying **events**, filter by the current shadow-source page | Querying **shadow-source posts**, filter by event activity |
| `EventListTypeControls` (`gatherpress_event_query`) | Upcoming vs past **events** | Upcoming vs past **events attached to** the shadow-source posts |

The two features are complementary and can coexist in the same template: a Productions archive page might use this new filter to list only active productions, and each production's single template uses `shadow_filter` to show only that production's upcoming performances.

---

## Open questions

1. **Performance**: the sub-query for all upcoming event IDs with `-1` posts per page could be expensive on large sites. Should the implementation use a `DISTINCT` raw SQL query against `wp_term_relationships` instead, or gate the feature behind a site-scale warning?

2. **Multiple event post types**: should `upcoming_events_only` check all post types declaring `gatherpress-event-date`, or only `gatherpress_event`? The `event_query_types` pattern in `Event_Query` already handles multi-type queries; the sub-query here should follow the same convention.

3. **Expose `post_ids_from_shadow_terms()` as a public `Shadow_Source` API?** This reverse-lookup helper would make the PHP implementation clean and reusable; it seems worth adding regardless of this feature.

### Code of Conduct

faisalahammad:
Fixes #1906

## Proposed changes:

* Add a new `Shadow_Source::get_source_post_ids_by_event_activity()` helper that resolves, for a shadow-source post type, the source post IDs whose shadow terms sit on an upcoming or past event. It costs two queries no matter how many events exist: a new `Event\Query::get_active_shadow_term_slugs()` runs one indexed join from the term relationships into the events table and selects only the distinct shadow term slugs, then the helper resolves those slugs to source posts with a single `post_name__in` lookup. No event rows are materialized in PHP, so the work is bounded by the number of source posts rather than by the size of the event table. The events-table SQL lives in `Event\Query` because that class already owns every events-table fragment and the upcoming/past column choice in `get_datetime_comparison_column()`.
* Add a "Filter by event activity" toggle to the GatherPress query panel (`HasEventsFilterControls`) plus an "Upcoming events only" sub-toggle (on by default). The toggle is gated on the *queried* post type declaring `gatherpress-shadow-source` and actually wiring its shadow taxonomy onto event post types, so unwired shadow sources don't show a filter that would match nothing.
* Apply the filter in `Event\Query::prepare_event_query_before_execution()` by pinning `post__in` to the resolved source IDs. A `null` sentinel from the helper means the filter is inapplicable (taxonomy not wired to events), so the query keeps its own scope; an empty array means the filter ran and matched nothing, so `post__in` is pinned to an impossible ID to return an empty set. When the query already carries a `post__in`, the resolved IDs are intersected with it rather than merged, so the filter narrows the caller's scope instead of widening it.
* Expose `has_events_filter` and `upcoming_events_only` through the REST API. Each support group owns its own REST hooks: `Event_Query` registers its `rest_query`/`rest_collection_params` only on event-date post types (unchanged from develop), while `Shadow_Source::maybe_register_post_type_hooks()` registers its own callbacks on shadow-source post types to pass the activity params through. `query_loop_block_query_vars` and `aql_query_vars` copy the event vars and the activity vars in separate support-gated branches so the two param groups never cross.
* Add `hasEventActivityFilterSupport()` to the JS event helper so the editor can check whether a shadow-source type's taxonomy is wired onto events.

### Other information:

## Testing instructions:

* In a site with the venue post type, create two venues: one with an upcoming event attached, one with only a past event (or no event).
* Add an event `core/query` block and set its post type to `gatherpress_venue`.
* In the GatherPress query panel, enable "Filter by event activity". With "Upcoming events only" on, confirm only the venue with the upcoming event appears. Turn "Upcoming events only" off, confirm the venue with only a past event appears (archive view).
* Confirm the same filtering holds in the editor preview (REST) and on the front end.

### Changelog entry

- [x] Automatically create a changelog entry from the details below.

<details>

#### Significance

- [ ] Patch
- [x] Minor
- [ ] Major

#### Type

- [x] Added - for new features
- [ ] Changed - for changes in existing functionality
- [ ] Deprecated - for soon-to-be removed features
- [ ] Removed - for now removed features
- [ ] Fixed - for any bug fixes
- [ ] Security - in case of vulnerabilities

#### Message

Added the ability to filter a Query Loop of shadow-source posts (venues, productions, seasons) by whether those sources have upcoming or past events attached. A "Filter by event activity" toggle in the GatherPress query panel limits the list to sources carrying at least one event term, and an "Upcoming events only" toggle (on by default) narrows it to sources with upcoming events, or off to sources whose events are all past for an archive view. Supported over the REST API for editor previews.

</details>

## Summary by CodeRabbit

* **New Features**
  * Query Loop blocks now support shadow-source post types, such as venues and productions.
  * Added filters to show shadow-source posts with upcoming or past events.
  * Query settings now display relevant event-activity controls based on the selected post type.
  * REST and front-end queries apply event and shadow-source filters appropriately.

* **Bug Fixes**
  * Improved filtering accuracy when combining event activity filters with existing query constraints.

* **Tests**
  * Added comprehensive coverage for shadow-source queries, activity filters, and Query Loop controls.


includes/core/classes/class-shadow-source.php
```diff
@@ -605,4 +607,91 @@ public function build_tax_query_clause( WP_Post $source_post ): array {
...
+			|| ! in_array( Event::POST_TYPE, $taxonomy->object_type, true )
+		) {
+			return null;
+		}
+
+		$args = array(
+			'post_type'               => get_post_types_by_support( 'gatherpress-event-date' ),
+			'fields'                  => 'ids',
+			'no_found_rows'           => true,
+			'posts_per_page'          => -1,
```
mauteri: Flagging: This strikes me as a query that will have trouble scaling as it does a full table scan with no upper limit and then also does a join on the GatherPress event table with `gatherpress_event_query`.

faisalahammad: Good catch. Fixed in c8ecacf3.

It was `1 + N + 2M` queries: an unbounded `WP_Query` over all event post types, then `wp_get_object_terms()` per event and `get_term()` plus `get_page_by_path()` per term. Now it is 2 queries, flat.

New `Event\Query::get_active_shadow_term_slugs()` joins the term relationships straight into the events table and selects only `DISTINCT t.slug`, so no event rows load into PHP and the rows returned are bounded by the venue count, not the event table. Sentinels are filtered by a `LIKE` on the slug. `Shadow_Source` then resolves those slugs with one `post_name__in` lookup. Every index it needs already exists.

Left it uncached on purpose: the result depends on "now", so a cache would go stale as events cross the boundary. Happy to add one if you prefer.

Also found that `post__in` used `array_merge`, which widened the set, so the filter added posts instead of restricting them. Changed to intersect.

All existing tests for this feature pass unchanged.

carstingaxion:
@faisalahammad @mauteri I followed the testing instructions on this PR using the generated PR playground (which has the full gatherpress-demo-data):

1. Neither a `core/query` block, nor the gatherpress-variation showed any *new* toggles after switching to `gatherpress_venue`
2. The browser console revealed a 404 on https://playground.wordpress.net/scope:curious-sunny-valley/wp-json/wp/v2/taxonomies/_gatherpress_event?context=edit&_locale=user, which looks very strange to me, because their shouldn't be any _gatherpress_event taxonomy at all. Havent looked closer, where the request is made.
3. Even without the new toggle, a normal query for venues should work and bring up all venues, but the result is zero. No venues at all.

includes/core/classes/blocks/class-event-query.php
```diff
@@ -116,13 +116,44 @@ protected function setup_hooks(): void {
 	 * @return void
 	 */
 	public function register_existing_event_date_post_types(): void {
-		foreach ( get_post_types_by_support( 'gatherpress-event-date' ) as $post_type ) {
+		foreach ( $this->get_query_rest_post_types() as $post_type ) {
 			$this->maybe_register_event_date_rest_hooks( $post_type );
 		}
 	}
```
carstingaxion, line 123: This may be technical correct, which I'm still unsure, but it got harder to understand now, that the method not only "registers existing event date post types" but also venue-information types.

includes/core/classes/blocks/class-event-query.php
```diff
@@ -131,7 +162,7 @@ public function register_existing_event_date_post_types(): void {
 	 * @return void
 	 */
 	public function maybe_register_event_date_rest_hooks( string $post_type ): void {
-		if ( ! post_type_supports( $post_type, 'gatherpress-event-date' ) ) {
+		if ( ! in_array( $post_type, $this->get_query_rest_post_types(), true ) ) {
 			return;
 		}
```
carstingaxion, line 212: Same as above: feels awkward to broaden the scope of the methods like this. (But dont have a better idea, yet)

includes/core/classes/blocks/class-event-query.php
```diff
@@ -567,6 +638,17 @@ public function aql_query_vars( array $query_args, array $block_query ): array {
...
 
+		// Pass through source-post event activity filtering.
+		if ( ! empty( $block_query['has_events_filter'] ) ) {
+			$query_args['has_events_filter'] = $block_query['has_events_filter'];
+		}
+
+		// Pass through the upcoming-only flag; only read when the activity
+		// filter is on, and it defaults to upcoming.
+		if ( isset( $block_query['upcoming_events_only'] ) ) {
+			$query_args['upcoming_events_only'] = $block_query['upcoming_events_only'];
```
carstingaxion, line 794: I guess the docs say more, than the check does. Or am I misunderstanding that?

includes/core/classes/event/class-query.php
```diff
@@ -275,6 +275,53 @@ static function () use ( $page_id ) {
...
+			$post_type = $query->get( 'post_type' );
+
+			if (
+				is_string( $post_type )
+				&& post_type_supports( $post_type, 'gatherpress-shadow-source' )
+			) {
+				// The block writes upcoming_events_only as an integer (0/1), default
+				// on. Coerce the string form too so both gate paths agree.
+				$upcoming = 'upcoming' === $query->get( 'upcoming_events_only', 'upcoming' )
+					|| 1 === (int) $query->get( 'upcoming_events_only' );
```
carstingaxion: Why do you check both variants? Where could this 'upcoming' string being set, so we could check it here?

src/helpers/event.js
```diff
@@ -102,6 +102,33 @@ export function isEventPostType( postType = null ) {
...
+ * @return {boolean} True if the source's shadow taxonomy attaches to events.
+ */
+export function hasEventActivityFilterSupport( sourcePostType ) {
+	if ( ! sourcePostType ) {
+		return false;
+	}
+
+	const taxonomy = select( 'core' ).getTaxonomy( `_${ sourcePostType }` );
+
+	return !! taxonomy?.types?.includes( 'gatherpress_event' );
```
carstingaxion: `gatherpress_event` shouldn't be hardcoded here. I guess we have similar post_type_support checks already around like `isEventPostType` or `isPostTypeSupporting`

includes/core/classes/blocks/class-event-query.php
faisalahammad: Resolved in 253d57a. The shared registration path now intentionally covers both post types that support `gatherpress-event-date` and post types that support `gatherpress-venue-information`/`gatherpress-shadow-source`. Venue and custom shadow-source Query Loops need their own REST collection/query hooks for `has_events_filter`; the helper deduplicates overlapping post types.

includes/core/classes/blocks/class-event-query.php
faisalahammad: Resolved in 253d57a. Scope is intentionally broadened for the same reason: shadow-source collections must receive REST schema and query handling even when they are not event-date post types. Support-based resolution keeps this extensible for custom event and shadow-source post types.

includes/core/classes/blocks/class-event-query.php
faisalahammad: Resolved in 253d57a. Documentation now describes both gates: the queried type must declare `gatherpress-shadow-source`, and its shadow taxonomy must be attached to at least one post type declaring `gatherpress-event-date`. The implementation checks both conditions before exposing the activity filter.

includes/core/classes/event/class-query.php
faisalahammad: Resolved in 253d57a. Normal block attributes write integer `0`/`1`, but AQL and direct query integrations can provide literal `upcoming`/`past` strings. The coercion preserves both representations, defaults missing values to upcoming, and keeps `past` on the past branch.

src/helpers/event.js
faisalahammad: Resolved in 253d57a. The editor now uses support-aware helpers and a reactive `useHasEventActivityFilterSupport` hook. It checks the queried type's `gatherpress-shadow-source` support and verifies taxonomy wiring against all post types declaring `gatherpress-event-date`, so `gatherpress_event` is no longer hardcoded.

faisalahammad:
## CI Fix Summary — 2 failures resolved

| # | File | Error | Fix |
|---|------|-------|-----|
| 1 | `test/unit/js/src/helpers/event.test.js:235` | Spell Check with Typos — `recognises` should be `recognizes` | Corrected to US English spelling |
| 2 | `src/helpers/event.js:136,150` | Check Test Coverage on Changed Files — 98.53% (134/136 statements) | Added 2 tests to `describe('hasEventActivityFilterSupport', ...)` covering both previously-untested early-return branches |

Tests ✅ (1038/1038) · Coverage ✅ (`event.js` 100%) · Lint ✅ · CodeRabbit ✅ · Independent verification ✅

### Also addressed

@carstingaxion reported the "Filter by event activity" toggle never appearing, a 404 on a `_gatherpress_event` taxonomy REST endpoint, and a plain venue query returning zero results.

- **Toggle never appearing** — found and fixed. `showHasEventsFilterControl` in `src/variations/core/query/components.js` required the queried post type to differ from the editing host's post type (or be inside a Site Editor template) — a requirement copied from the unrelated "Filter by Current Venue" toggle that doesn't apply to this one. Dropped it, and added matching test coverage in `components.test.js`.
- **404 on `_gatherpress_event`** — no code path found anywhere in the diff that would ever request this taxonomy (the only place a `_<postType>` taxonomy name is built is gated behind `gatherpress-shadow-source` support, which `gatherpress_event` never has). Most likely a stale cached editor bundle from iterative testing in the shared Playground session — worth a re-test in a fresh instance before digging further.
- **Zero venues returned** — no default in the REST schema or code path zeroes this out unconditionally; most likely the tested query block still carried a stale `has_events_filter: 1` attribute left over from earlier testing (hiding the toggle doesn't clear an already-set attribute). Worth re-testing with a freshly-inserted query block.

carstingaxion:
>1. Toggle never appearing — found and fixed. showHasEventsFilterControl in src/variations/core/query/components.js required the queried post type to differ from the editing host's post type (or be inside a Site Editor template) — a requirement copied from the unrelated "Filter by Current Venue" toggle that doesn't apply to this one. Dropped it, and added matching test coverage in components.test.js.

Yes, this worked now. The toggle appeared within the Venue Query Settings panel and did its job. The second one, to switch upcoming or past events, appeared and worked also. 

>2. 404 on _gatherpress_event — no code path found anywhere in the diff that would ever request this taxonomy (the only place a _<postType> taxonomy name is built is gated behind gatherpress-shadow-source support, which gatherpress_event never has). Most likely a stale cached editor bundle from iterative testing in the shared Playground session — worth a re-test in a fresh instance before digging further.

I dont know why, but its gone, too. Maybe that kind of in browser caching something, you mentioned.

>3. Zero venues returned — no default in the REST schema or code path zeroes this out unconditionally; most likely the tested query block still carried a stale has_events_filter: 1 attribute left over from earlier testing (hiding the toggle doesn't clear an already-set attribute). Worth re-testing with a freshly-inserted query block.

Ok, this one is solved or fixed or disappeared now, too. A regular core/query, querying for venues, listed all existing regardless of their event-activity. Done.

<hr>

Unfortunately, I found another thing that might be a problem from that PR. The "Event Query settings" panel is empty, when I select "Events" as post type to query. I'm not 100% sure, this bug is caused by your PR, but I haven't seen it before so let us investigate.

[Bildschirmaufzeichnung vom 04.09.2026, 10:57:25.webm](https://github.com/user-attachments/assets/dc441bbd-f497-4d20-a1a8-d21c5bb57615)

faisalahammad:
Thanks for testing and for the recording. That report was real, and it was caused by this PR.

The query controls refactor gated the event controls on the post type of the page the editor is open on, instead of the post type the query loop asks for. So a query loop on a page, querying Events, ended up with an empty panel.

Fixed in 98f7ad11. The five event controls (list type, include unfinished, per page, offset, order) now show when the queried post type supports event dates. Exclude current event keeps its host rule, and a query without a post type set no longer falls back to the host check. A regression test covers the page host with Events queried.

mauteri:
@faisalahammad heads up, #2270 just merged and it renames the post type supports from string literals to class constants:

| Literal | Constant |
| --- | --- |

This branch adds 55 lines using the literal form, 29 of `gatherpress-shadow-source` and 26 of `gatherpress-event-date`, which is by some way the most of any open PR. Could you merge `develop` in and swap them over? The values are unchanged, so it is a mechanical find and replace plus the `use` statements.

No rush on my account, and nothing here is wrong today. It is just that the codebase now has one way of naming these and it would be a shame for this to land as the exception.

faisalahammad:
@mauteri done, thanks for flagging it.

I merged `origin/develop` (now at `9e630081`) and swapped the 55 production PHP literals to the new constants. The only places the string still appears in `includes/` are the two `const SUPPORT = '…'` definitions and a few docblock/comment lines. Imports added where needed: `GatherPress\Core\Shadow_Source` in `Blocks\Event_Query`, `GatherPress\Core\Event` in `Shadow_Source`.

A small note on test fixtures: I left the string-literal fixtures in the PHP and JS tests as-is. That matches the convention that #2270 set in the develop tests (its own `class-test-event.php` keeps 28 `gatherpress-event-date` literals and the new constants are guarded by `assertSame` value tests). I would rather not churn the test fixtures in the same PR that introduced the feature. Happy to do a separate cleanup pass if you would prefer.

`npm run lint:php` and `npm run lint:phpstan` are clean, and `npm run test:unit:php` plus the multisite suite still pass with the three PR production files at 100% statements. Coverage bot should be happy again on the next push.

faisalahammad:
@carstingaxion heads up: I just merged `origin/develop` into the PR (`9e630081`), and a follow-up commit `3878b167` tightens try/finally cleanup in the new PHPUnit tests. The five event controls fix you re-tested earlier (commit `98f7ad11`) is in the merge, so the "Event Query settings" panel behaviour should still hold.

Could you re-run your earlier test pass on the new build, especially the "Events as post type to query" scenario from your recording? The develop merge brought in `aecb529b` (#2270) which renames `gatherpress-event-date` and `gatherpress-shadow-source` to `Event::SUPPORT` / `Shadow_Source::SUPPORT`; the public behaviour of the panel is unchanged but I would like a fresh confirmation that the controls still appear.

carstingaxion (dismissed):
Tested again, manually from the PR Playground. This looks good and seems to works as it should. I re-tested all from my previous tests and didn't encounter any issues.

Thank you very much for your work @faisalahammad !

Up to you @mauteri

mauteri:
@carstingaxion @faisalahammad This is a large PR, so I'm going to start with a CodeRabbit Review and then I'll get to reviewing it and testing it myself. @faisalahammad once CodeRabbit completes, please review the feedback.

src/variations/core/query/controls.js
```diff
@@ -201,7 +221,12 @@ export const EventQueryControlsPanel = ( props ) => {
 		__( 'Event', 'gatherpress' )
 	);
 
-	if ( ! queryPostTypeSupportsEvents ) {
+	// The panel stays open when the queried post type declares either
+	// event-date or shadow-source support. The inner SlotFill renders
+	// event-only controls only for event-date support, activity controls
+	// only for shadow-source support, and nothing otherwise. For unrelated
+	// post types the inner content is empty so we collapse the whole panel.
+	if ( ! queryPostTypeSupportsEvents && ! queryPostTypeSupportsShadowSource ) {
```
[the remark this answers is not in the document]
carstingaxion: Oh, yes. Encountered exactly this while testing, but forgot to mention.

includes/core/classes/blocks/class-event-query.php
```diff
@@ -381,46 +498,80 @@ public function rest_query( array $args, WP_REST_Request $request ): array {
...
+			$custom_args['orderby'] = $request->get_param( 'orderby' );
+
+			$shadow_filter = $request->get_param( 'shadow_filter' );
+			if ( null !== $shadow_filter ) {
+				$custom_args['shadow_filter'] = $shadow_filter;
+			}
 		}
 
-		$custom_args['orderby'] = $request->get_param( 'orderby' );
+		if ( $query_shadow_supports ) {
```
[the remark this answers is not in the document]
faisalahammad: Fixed in a5af63b7. The context params now also pass through on event queries. A new branch `if ( $query_shadow_supports || $query_event_supports )` reads them and forwards them to the merged query args. Added a regression test where an event REST request with the params gets them forwarded.

src/variations/core/query/controls.js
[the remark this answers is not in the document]
faisalahammad: Fixed in a5af63b7. The strip effect now removes `has_events_filter` and `upcoming_events_only` when the queried post type loses shadow-source support. The context params are only removed when the type supports neither feature, so the strip does not fight the backfill effect on event queries. Added tests for both cases.

test/unit/php/includes/tests/core/classes/event/class-test-query.php
```diff
@@ -1774,6 +2160,310 @@ static function ( $key, $value ) use ( &$set_called_with_tax_query ) {
...
+		// An administrator doesn't have the custom cap by default, so the
+		// private production event must NOT be picked up yet.
+		$admin_id = $this->factory->user->create( array( 'role' => 'administrator' ) );
+		wp_set_current_user( $admin_id );
+
+		$this->assertNotContains(
+			'_slugs-custom-cap-venue',
+			$instance->get_active_shadow_term_slugs( Venue::TAXONOMY, true ),
+			'A private custom-cap event should not resolve without the post-type read_private_posts cap.'
+		);
```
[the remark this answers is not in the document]
faisalahammad: Fixed in a5af63b7. The custom post type test now wraps the support check, the admin assertion, the cap grant and the final assertion in one try block. The finally restores the post type and the user. Same pattern applied in the shadow-source taxonomy test.

faisalahammad:
All CodeRabbit feedback is now addressed in a5af63b7:

1. **Event query context forwarding (major)** - The shadow-source context params now also pass through on event queries. A new branch `if ( $query_shadow_supports || $query_event_supports )` reads them and forwards them to the merged query args. Added a regression test where an event REST request with the params gets them forwarded.

2. **Clear activity attrs on support removal (minor)** - The strip effect now removes `has_events_filter` and `upcoming_events_only` when the queried post type loses shadow-source support. The context params are only removed when the type supports neither feature, so the strip does not fight the backfill effect and the block does not ping-pong between two states.

3. **try/finally in tests (minor)** - The custom post type test and the shadow-source taxonomy test now wrap their mutations in try blocks. The finally always restores the post type, the taxonomy wiring and the current user.

Also folded in the nitpicks: single sprintf help text, a shared `resolveEventActivityFilterSupport` helper for the sync and hook checks, extra JS tests for untested branches, and a behavioral render test instead of a call-count assertion.

Verified locally: PHPCS, PHPStan, ESLint, Jest (64 suites, 1175 tests) and the touched PHPUnit suites all pass. @mauteri could you take another look?

src/variations/core/query/controls.js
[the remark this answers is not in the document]
carstingaxion: @coderabbit re-review this.

includes/core/classes/blocks/class-event-query.php
```diff
@@ -116,13 +117,88 @@ protected function setup_hooks(): void {
...
+	 * vice versa) would 400 the request because the param isn't on the
+	 * matching collection's schema.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @param array<string, mixed> $args WP_Query args built by the REST layer.
+	 *
+	 * @return string[] Post type slugs referenced by `$args['post_type']`.
+	 */
+	protected function get_requested_post_types_from_args( array $args ): array {
```
mauteri: This can be one path instead of three. Start with an empty array, let the two branches decide what goes in it as an `if`/`elseif`, and do the `array_values( array_filter() )` once at the end, so nothing is repeated and there's a single return:

```php
protected function get_requested_post_types_from_args( array $args ): array {
	$post_type  = $args['post_type'] ?? '';
	$post_types = array();

	if ( is_array( $post_type ) ) {
		$post_types = array_map( 'strval', $post_type );
	} elseif ( is_string( $post_type ) && '' !== $post_type ) {
		$post_types = array_map( 'trim', explode( ',', $post_type ) );
	}

	return array_values( array_filter( $post_types ) );
}
```

Behavior is identical, the empty case comes out as `array()` through the same last line, and the existing `test_get_requested_post_types_from_args_*` cases pass against it as written.

includes/core/classes/blocks/class-event-query.php
```diff
@@ -116,13 +117,88 @@ protected function setup_hooks(): void {
...
+	 * `gatherpress-shadow-source`) get the collection filters. Shadow-source
+	 * queries need them so the "filter by event activity" params
+	 * (`has_events_filter`, `upcoming_events_only`) are accepted when a query
+	 * loop lists the source posts themselves.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @return string[] Post types whose REST collections accept GatherPress query params.
+	 */
+	protected function get_query_rest_post_types(): array {
```
mauteri: Stepping back from the small fix I left on `get_requested_post_types_from_args()`: this class is doing two jobs through the same hooks, and that's what makes it hard to follow. Let's separate them.

**Two concerns, one pipe.** The event-loop params (`gatherpress_event_query`, `include_unfinished`, `orderby=rand|datetime`) belong to `Event::SUPPORT` post types. The activity params this PR adds (`has_events_filter`, `upcoming_events_only`) belong to `Shadow_Source::SUPPORT` post types. Right now both register the same `rest_{type}_query` and `rest_{type}_collection_params` callbacks on the union of both supports, and every callback then has to work out which kind of post type it's on: `$is_event_date` / `$is_shadow` in `add_gatherpress_collection_params()`, `$query_event_supports` / `$query_shadow_supports` in the query path, and the whole helper that parses `post_type` back out of the args so an event-only param doesn't 400 a shadow listing.

**The split.** `Blocks\Event_Query` keeps `maybe_register_event_date_rest_hooks()` gated on `Event::SUPPORT`, as it was on develop, wiring only the event params. The activity params move to `Shadow_Source::maybe_register_post_type_hooks()`, which already runs on `registered_post_type` for every shadow-source type and already hangs that type's term-sync hooks there. Give it its own two callbacks for `rest_{type}_query` and `rest_{type}_collection_params` carrying only `has_events_filter` and `upcoming_events_only`. Each domain class then registers REST behavior for its own post types, and the block just reads it.

**What that removes.** `get_query_rest_post_types()` and its union. The `$is_event_date` / `$is_shadow` pair. The `$query_event_supports` / `$query_shadow_supports` pair. `get_requested_post_types_from_args()` entirely, so ignore my earlier comment on it. Each callback only ever sees the post types it was registered for, and a post type that somehow declared both supports just gets both sets of hooks, which is correct without any gating.

The block-side branch that copies block query vars into query args has the same shape, so it splits the same way: event vars on event loops, activity vars on shadow loops, each in its own place.

includes/core/classes/event/class-query.php
```diff
@@ -662,6 +710,131 @@ protected function get_datetime_comparison_column( string $type, bool $inclusive
...
+	 * deliberately do not.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @param string $taxonomy Shadow taxonomy to read term slugs from.
+	 * @param bool   $upcoming Whether to match upcoming events rather than past ones.
+	 *
+	 * @return string[] Distinct shadow term slugs, empty when nothing matched.
+	 */
+	public function get_active_shadow_term_slugs( string $taxonomy, bool $upcoming ): array {
```
mauteri, line 734: This method is too much to follow as one piece, and the shape of it is the reason: the whole statement is assembled from string fragments and then every value is bound in one `array_merge` at the very bottom, so a reader has to count placeholders back to arguments. On top of that, `$post_type_placeholders` at 789 and `$private_type_condition` at 791 are variables spliced into SQL text. Both happen to be safe here, since one is a generated run of `%s` and the other is literals, but nobody should have to prove that on review, and it's the exact pattern that turns into an injection when someone later passes a real value through it.

The single indexed join itself stays. It's the only shape whose result is bounded by the number of venues rather than the number of events, which is what the scaling comment earlier asked for, and nothing in `WP_Query` or `WP_Term_Query` can express it without materializing the event set.

Four changes, all verified against your existing tests:

1. **Split it.** The public method reads top to bottom: get the event types, bail on none, build, run, return. The statement lives in a builder; the status condition and the `IN ( … )` list get their own small helpers.
2. **Prepare each condition where it's written.** Every `WHERE` term is a `$wpdb->prepare()` call with its value beside its placeholder, and the builder only joins finished fragments.
3. **No variables in SQL text.** The one place a placeholder run is generated is `prepare_in_list()`, which returns an already-prepared fragment; callers write the column as a literal and join the fragment to it. The private-status condition comes back from `prepare_readable_status_clause()` already prepared.
4. **One `phpcs` block per concern**, around the specific line that needs it, instead of four suppressions wrapping the whole thing.

```php
public function get_active_shadow_term_slugs( string $taxonomy, bool $upcoming ): array {
	global $wpdb;

	$event_post_types = get_post_types_by_support( Event::SUPPORT );

	// With no event post type registered there is nothing the filter
	// could match, and an empty IN () list is invalid SQL anyway.
	if ( empty( $event_post_types ) ) {
		return array();
	}

	$sql = $this->build_active_shadow_term_sql( $taxonomy, $upcoming, $event_post_types );

	// Every fragment of $sql was prepared where it was written. The result
	// is time-relative, so a persistent cache would go stale as events
	// cross the upcoming/past boundary.
	// phpcs:disable WordPress.DB.PreparedSQL.NotPrepared
	// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
	// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
	$slugs = $wpdb->get_col( $sql );
	// phpcs:enable WordPress.DB.DirectDatabaseQuery.NoCaching
	// phpcs:enable WordPress.DB.DirectDatabaseQuery.DirectQuery
	// phpcs:enable WordPress.DB.PreparedSQL.NotPrepared

	return array_map( 'strval', (array) $slugs );
}

/**
 * Build the statement behind get_active_shadow_term_slugs().
 *
 * The join runs from the term relationships straight into the events
 * table, so the result is bounded by the source posts that carry events
 * rather than by the size of the event table. Each condition is prepared
 * beside the value it binds, and no variable is ever spliced into SQL
 * text: the statement only joins fragments prepare() already produced.
 *
 * @since 0.36.0
 *
 * @param string   $taxonomy         Shadow taxonomy slug.
 * @param bool     $upcoming         True for upcoming events, false for past.
 * @param string[] $event_post_types Event post type slugs, never empty.
 *
 * @return string The prepared statement.
 */
protected function build_active_shadow_term_sql(
	string $taxonomy,
	bool $upcoming,
	array $event_post_types
): string {
	global $wpdb;

	// Same column and clock the posts_clauses filters use, so this query
	// and a regular upcoming/past event query always agree on the boundary.
	$column = $this->get_datetime_comparison_column( $upcoming ? 'upcoming' : 'past', $upcoming );
	$now    = gmdate( Event::DATETIME_FORMAT, time() );

	// Identifier placeholders use the %i form phpcs does not yet recognize.
	// phpcs:disable WordPress.DB.PreparedSQL.NotPrepared
	// phpcs:disable WordPress.DB.PreparedSQLPlaceholders.UnsupportedIdentifierPlaceholder
	$from = $wpdb->prepare(
		'FROM %i AS tr'
		. ' INNER JOIN %i AS tt ON tt.term_taxonomy_id = tr.term_taxonomy_id'
		. ' INNER JOIN %i AS t ON t.term_id = tt.term_id'
		. ' INNER JOIN %i AS e ON e.post_id = tr.object_id'
		. ' INNER JOIN %i AS p ON p.ID = tr.object_id',
		$wpdb->term_relationships,
		$wpdb->term_taxonomy,
		$wpdb->terms,
		sprintf( Event::TABLE_FORMAT, $wpdb->prefix ),
		$wpdb->posts
	);

	$where = array(
		$wpdb->prepare( 'tt.taxonomy = %s', $taxonomy ),
		'p.post_type ' . $this->prepare_in_list( $event_post_types ),
		$this->prepare_readable_status_clause( $event_post_types ),
		// Real shadow term slugs carry the leading underscore that
		// Shadow_Source::term_slug_from_post_name() adds; sentinels such
		// as the venue subsystem's `online-event` deliberately do not.
		$wpdb->prepare( 't.slug LIKE %s', $wpdb->esc_like( '_' ) . '%' ),
		$upcoming
			? $wpdb->prepare( 'e.%i >= %s', $column, $now )
			: $wpdb->prepare( 'e.%i < %s', $column, $now ),
	);

	$sql = 'SELECT DISTINCT t.slug ' . $from . ' WHERE ' . implode( ' AND ', $where );
	// phpcs:enable WordPress.DB.PreparedSQLPlaceholders.UnsupportedIdentifierPlaceholder
	// phpcs:enable WordPress.DB.PreparedSQL.NotPrepared

	return $sql;
}

/**
 * The post status condition: published events, plus private ones for
 * each event post type whose read_private_posts capability the current
 * user holds.
 *
 * @since 0.36.0
 *
 * @param string[] $event_post_types Event post type slugs, never empty.
 *
 * @return string The prepared condition, parenthesized.
 */
protected function prepare_readable_status_clause( array $event_post_types ): string {
	global $wpdb;

	$private_post_types = array_values(
		array_filter(
			$event_post_types,
			static function ( string $post_type ): bool {
				$post_type_object = get_post_type_object( $post_type );

				return current_user_can( $post_type_object->cap->read_private_posts ?? 'read_private_posts' );
			}
		)
	);

	$clause = $wpdb->prepare( 'p.post_status = %s', 'publish' );

	if ( ! empty( $private_post_types ) ) {
		$clause .= $wpdb->prepare( ' OR ( p.post_status = %s AND p.post_type ', 'private' )
			. $this->prepare_in_list( $private_post_types )
			. ' )';
	}

	return '( ' . $clause . ' )';
}

/**
 * A prepared `IN ( ... )` list for string values.
 *
 * The only place a placeholder run is generated. Its length is the count
 * of the values and every value is bound by prepare(), so a caller writes
 * its column as a literal and joins this fragment to it, never a variable.
 *
 * @since 0.36.0
 *
 * @param string[] $values The values to match, never empty.
 *
 * @return string The prepared fragment, including the IN keyword.
 */
protected function prepare_in_list( array $values ): string {
	global $wpdb;

	// phpcs:disable WordPress.DB.PreparedSQL.NotPrepared
	// phpcs:disable WordPress.DB.PreparedSQLPlaceholders.ReplacementsWrongNumber
	$fragment = $wpdb->prepare(
		'IN (' . implode( ', ', array_fill( 0, count( $values ), '%s' ) ) . ')',
		array_values( $values )
	);
	// phpcs:enable WordPress.DB.PreparedSQLPlaceholders.ReplacementsWrongNumber
	// phpcs:enable WordPress.DB.PreparedSQL.NotPrepared

	return $fragment;
}
```

All of your tests for this method and for `get_source_post_ids_by_event_activity()` pass against this unchanged, and phpcs, PHPStan and PHPMD are clean on the file.

One thing worth knowing while you rework it: `%i` wraps its value in a single pair of backticks, so it's for bare identifiers only. `e.%i` with the column name is fine, which is why the boundary condition keeps it. Passing `p.post_type` through `%i` produces one column literally named `p.post_type` and the query silently matches nothing, which is why the `IN` helper takes only the values and leaves the column to the caller.

src/helpers/event.js
```diff
@@ -102,6 +102,100 @@ export function isEventPostType( postType = null ) {
...
+ * Shared four-step resolution used by both `hasEventActivityFilterSupport`
+ * (synchronous) and `useHasEventActivityFilterSupport` (reactive).
+ *
+ * @param {Function} selectFn       Data store selector (select or wpSelect).
+ * @param {string}   sourcePostType Shadow-source post type to check.
+ *
+ * @return {boolean} True if the source's shadow taxonomy attaches to an
+ *                   event-supporting post type.
+ */
+function resolveEventActivityFilterSupport( selectFn, sourcePostType ) {
```
mauteri, line 118: Five returns here, and two of the guards aren't doing anything the last line doesn't already do: `some()` on an empty array is false, and a not-yet-loaded `getPostTypes()` becomes an empty list once you default it. The first two guards fold into one boolean, which also keeps the taxonomy lookup from firing a fetch for a post type that isn't a shadow source.

```js
function resolveEventActivityFilterSupport( selectFn, sourcePostType ) {
	// The taxonomy is only worth looking up for a post type that declares
	// shadow-source support; asking for a bogus one would start a fetch.
	const isShadowSource =
		!! sourcePostType &&
		!! selectFn( 'core' ).getPostType( sourcePostType )?.supports?.[
			'gatherpress-shadow-source'
		];
	const taxonomy = isShadowSource
		? selectFn( 'core' ).getTaxonomy( `_${ sourcePostType }` )
		: undefined;

	if ( ! taxonomy?.types?.length ) {
		return false;
	}

	// Nothing loaded yet reads as no event post types, and `some()` on an
	// empty list is false, so both former guards fold into the result.
	const eventPostTypes = (
		selectFn( 'core' ).getPostTypes( { per_page: -1, context: 'edit' } ) ??
		[]
	)
		.filter( ( type ) => type?.supports?.[ 'gatherpress-event-date' ] )
		.map( ( type ) => type.slug );

	return eventPostTypes.some( ( slug ) => taxonomy.types.includes( slug ) );
}
```

Behavior is the same, and your tests in `event.test.js` and `components.test.js` pass against it as written. Same shape we hold PHP to: one guard chain, one result.

mauteri (changes requested):

includes/core/classes/blocks/class-event-query.php
faisalahammad: This method is gone now. The REST hook split made it unnecessary: Event_Query registers its hooks only on event post types, so each callback knows its own group and nothing has to parse the post type back out of the query args.

includes/core/classes/blocks/class-event-query.php
faisalahammad: Done in d2ecaaf1. Event_Query registers its rest hooks only on event-date post types again, same as develop. Shadow_Source::maybe_register_post_type_hooks() now registers its own rest_query and rest_collection_params callbacks on shadow-source types, carrying only has_events_filter and upcoming_events_only. The is_event_date/is_shadow and query_event_supports/query_shadow_supports pairs are gone, and get_query_rest_post_types() and get_requested_post_types_from_args() are deleted. query_loop_block_query_vars and aql_query_vars copy each param group in its own support-gated branch, with a test asserting the groups never cross. A post type that declared both supports would get both callbacks, which is the correct behavior.

includes/core/classes/event/class-query.php
faisalahammad: Done in d2ecaaf1. get_active_shadow_term_slugs() now reads top to bottom: it gets event types, bails when none, builds the prepared SQL, runs it, and returns the slugs. The SQL builder keeps the indexed join. prepare_readable_status_clause() and prepare_in_list() prepare each condition beside its value, so no values are spliced into SQL text. I also added direct reflection tests for all three helpers.

src/helpers/event.js
faisalahammad: Done in d2ecaaf1. resolveEventActivityFilterSupport() now uses one isShadowSource check, one taxonomy guard, getPostTypes() with an empty fallback, and one final some(). Existing event helper and Query Loop control tests pass.

faisalahammad:
Implemented all four requested changes in d2ecaaf1.

- REST hooks are split by support. Event_Query owns event-date params. Shadow_Source owns activity params.
- The unused post-type parsing helpers are deleted.
- Event activity SQL is split into prepared helper methods with direct reflection coverage.
- The event activity support helper is simplified.

Checks passed:
- PHPCS
- PHPStan
- ESLint
- Jest: 1175 tests passed
- PHPUnit: 44 + 86 + 77 + 122 tests passed
- CodeRabbit review: 0 findings

faisalahammad (commented):
## CI fix summary for c6739106

| Check | Before | Fix | After |
| --- | --- | --- | --- |
| PHPMD | NPathComplexity 26316 on `query_loop_block_query_vars()` (limit 9217) | Moved the event branch and the shadow-source branch into protected helpers. No logic change, merge order and casts stay the same. | 0 violations |
| Coverage gate | 98.41% on `class-event-query.php`, lines 361-366 uncovered | Added a test for the event branch context passing, plus direct tests for the helpers | 100% |

Ran the CI commands locally before pushing:

- PHPMD: pass
- PHPCS: pass
- PHPStan level 8: pass
- PHPUnit single site: 2043 tests pass
- PHPUnit multisite: 352 tests pass
- Jest: 1175 tests pass
- Changed files coverage: 100% on all three files

The coverage gap came from this PR itself. It changed the postType in an existing test to `gatherpress_venue`. That post type does not support event date, so the test stopped covering the event branch copy of the context params. The new test sets postType to the event type and covers that path again.

The helper tests call the methods directly on purpose. xdebug does not trace same class helpers called through short delegation, so the public method tests alone report the helper bodies as uncovered even though they run. The direct calls keep the coverage report true.

PR #2224, opened 2026-08-31
Omit end time for all-day events

### LINKED ISSUE #966: Omit end time for all-day events
All-day events should omit the start/end **time** in the editor UI and in the rendered output, and store a full-day span so downstream features stay correct.

## Data

- Register a `gatherpress_is-all-day` postmeta (boolean, default `false`) for every post type that supports `gatherpress-event-date`.
- When enabled, set the event start to the very beginning of the day and the event end to the very end of the day. The stored start/end then actually span the whole day, which keeps calendar exports (ICS), duration, and date queries correct rather than hiding a real time (e.g. an end that's still 3pm under the hood).

## UI

- Add an "All day" toggle, modeled on Google Calendar's all-day option in the date picker.
- Lives in the **Event Settings** sidebar panel and carries through to the **Event Date** block in the context of the event.
- When it's on, hide the time portion of the date/time picker.

## Display

- Use the existing [`gatherpress_datetime_format`](https://github.com/GatherPress/gatherpress/blob/develop/docs/developer/hooks/gatherpress_datetime_format.md) filter to strip the time from the rendered event date (singular, archive, and query-block contexts) when the postmeta is `true`.

## Acceptance

- Toggling "All day" on stores `gatherpress_is-all-day = true` and sets start/end to the day boundaries.
- The time fields are hidden in the picker while all-day is on.
- Rendered event dates omit the time everywhere (block, singular, archive, query) for all-day events.
- Behavior is unchanged for non-all-day events.

---

_Originally discussed in meeting Dec 13, 2024: block does not display start/end time for all-day events._

mauteri:
Fixes #966

An all-day event stores a real full-day span rather than hiding a time that is still 3pm underneath, so exports, duration and date queries stay correct — the point @mauteri made on the issue.

## Data

`gatherpress_is_all_day` is registered for every post type supporting `gatherpress-event-date`. Snake case, matching every other meta key in the plugin.

`Event::save_datetimes()` snaps the start and end to their own day's boundaries. That is the one place every writer funnels through, so the editor, REST, imports and WP-CLI all land on them rather than only the editor.

## Display

The site already keeps its date and time formats separately, so an all-day event simply uses the date one and never reaches for the time. Nothing has to take a format apart to remove the time from it. A one-day event's end drops out along with the separator before it.

```
timed:   August 29, 2026 6:00 pm to 8:00 pm UTC
all day: August 29, 2026
```

## Timezone

This is the part worth reviewing.

An all-day event carries no timezone, the way iCalendar's `DTSTART;VALUE=DATE` and Google Calendar's `start.date` do not. It is floating: August 29 is August 29 in Tokyo and in Los Angeles.

Converting our stored GMT would break that. A Tokyo event snapped to `2026-08-29 00:00:00` local has a GMT start of `2026-08-28 15:00:00` — the day before. So all-day formats from the local column, parsed and rendered in the event's own zone. Nothing crosses a meridian, the day cannot move, and a zone-bearing format still names the zone the day belongs to rather than GMT.

| timezone | display | rendered as GMT |
| --- | --- | --- |
| America/New_York | August 29, 2026 EDT | August 29, 2026 |
| Asia/Tokyo | August 29, 2026 JST | August 29, 2026 |
| UTC | August 29, 2026 UTC | August 29, 2026 |

The GMT columns stay a real instant, because ordering upcoming against past genuinely wants one.

## Editor

The toggle sits after Time zone in Event settings. Turning it on snaps the stored values, hides the time in both pickers and hides Duration, which is a length in hours an all-day event does not have. Turning it back off restores the times it replaced, onto whatever dates are selected by then, so flipping the toggle to look at it costs nothing.

Those times are remembered for the session only. An event saved as all-day no longer has other times to come back to, which is the point of storing a real span.

The Event Date block matches: date-only preview, date-only format placeholders, and Append time zone switching off when all-day is turned on — you can switch it back and it sticks.

## Known behaviour, not a bug in this PR

The block preview does not update the instant the toggle is flipped; it catches up when the block is selected. That is Gutenberg's `AsyncModeProvider` deferring updates for unselected blocks — a `core/editor` meta change alone does not re-render them. It affects any block reacting to post meta.

## Tests

PHP covers the data layer (the meta registration, the day-boundary snap, and that a timed event is untouched), the display (date only, and a multi-day event still naming the day it ends on), and the floating date across UTC, New York and Tokyo — the case that breaks under GMT conversion. `class-event.php` and `class-meta.php` are both at 100%.

JS covers the new `datetime.js` helpers, taking that file to 100%.

`src/components/**/*` and `src/blocks/*/edit.js` are excluded from the coverage gate by `.github/coverage-config.json`, so the editor components and the block's `edit.js` do not need to reach 100% to land.

## Follow-up

Calendar export is filed separately as #2225. Every exporter builds its timestamps from the GMT columns, so an all-day event serializes as a timed event running to `23:59:59`, and in a zone far enough from UTC it lands on the wrong day. The standard form is `VALUE=DATE` with an exclusive next-day end, read from the local columns. Storage stays as it is here — `23:59:59` is what keeps `datetime_end_gmt >= now` queries working; only the serialization changes.

## Testing

- Toggle All day on an event; the pickers lose their time and Duration disappears
- Save and check the stored span is `00:00:00` to `23:59:59`
- Change the event's timezone to Asia/Tokyo and confirm the rendered date does not move
- Toggle back off and confirm the previous times return

## Summary by CodeRabbit

* **New Features**
  * Added support for marking events as all-day.
  * All-day events use date-only pickers, formatting, and end-date controls.
  * All-day dates remain consistent across time zones.
  * Added controls for displaying the time zone by default, always, or never.
  * Event date blocks recognize and display all-day events.

* **Documentation**
  * Added guidance for creating all-day events and configuring time-zone display.

* **Tests**
  * Added coverage for all-day saving, formatting, metadata, time zones, and date helpers.

### DIFF
--- a/.github/changelog/966-all-day-events
+++ b/.github/changelog/966-all-day-events
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Add an all day toggle to events, which stores a real full-day span and renders the date without a time.
--- a/docs/developer/event-duration.md
+++ b/docs/developer/event-duration.md
@@ -16,6 +16,10 @@ When the editor shows the event date and time, it renders either the
 matches one of the duration options; picking "Set an end time…" switches to the
 absolute picker.
 
+An all-day event has no length in hours, so the Duration select is hidden while
+the **All day** toggle is on and the end date picker takes its place. Both
+filters still run; their result is simply not shown for that event.
+
 ## `gatherpress.durationOptions`
 
 Filters the array of duration presets shown in the Duration select.
@@ -35,7 +39,7 @@ addFilter(
 	( options ) => [
 		{ label: '3 hours', value: 3 },
 		{ label: '6 hours', value: 6 },
-		{ label: 'All day', value: 24 },
+		{ label: '12 hours', value: 12 },
 		// Keep a "Set an end time…" entry if you still want the absolute
 		// end-time picker to be reachable.
 		{ label: 'Set an end time…', value: false },
--- a/docs/developer/post-type-supports/README.md
+++ b/docs/developer/post-type-supports/README.md
@@ -10,13 +10,18 @@ These supports are declared on post types that act as **events**.
 
 The core identifier for event post types. Enables event datetime storage and display. This includes:
 
-- Registration of datetime meta fields (`gatherpress_datetime`, `gatherpress_datetime_start`, `gatherpress_datetime_end`, `gatherpress_timezone`, etc.) — GatherPress also auto-adds WordPress's `custom-fields` support to the post type so the REST controller actually attaches the `meta` field to the schema (without it, `register_post_meta()` quietly registers the keys but the editor's PUT silently strips them)
+- Registration of datetime meta fields (`gatherpress_datetime`, `gatherpress_datetime_start`, `gatherpress_datetime_end`, `gatherpress_timezone`, `gatherpress_is_all_day`, `gatherpress_show_timezone`, etc.) — GatherPress also auto-adds WordPress's `custom-fields` support to the post type so the REST controller actually attaches the `meta` field to the schema (without it, `register_post_meta()` quietly registers the keys but the editor's PUT silently strips them)
 - Storage in the `gatherpress_events` database table
 - Date-based query ordering (upcoming/past)
 - Event Date block rendering
 - Add to Calendar block rendering
 - RSS feed enrichment with event date information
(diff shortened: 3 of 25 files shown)


mauteri:
@coderabbitai full review

PR #2227, opened 2026-08-31
Fix tooltip accessibility: keyboard focus, touch toggle, screen reader text and client navigation (#2217)

### LINKED ISSUE #2217: Tooltip is mouse-only: not focusable, no touch, invisible to assistive tech
The tooltip format at `src/formats/tooltip/` renders its bubble entirely in CSS: `content: attr(data-gatherpress-tooltip)` on a `::after`, shown on `:hover, :focus`. That works for a mouse user on a desktop and for nobody else.

## What's missing

**Not focusable.** `style.scss:80` has `&:hover, &:focus`, but the format registers a plain `<span>` with three data attributes and no `tabindex`. A span isn't in the tab order, so the `:focus` half of that rule never fires. It reads as keyboard coverage and provides none.

**No touch support.** `:hover` on touch is unreliable, so tapping either does nothing or leaves a sticky hover state. On an events site this is likely most of the traffic.

**Invisible to assistive technology.** CSS-generated content is announced inconsistently at best, and there is no `role="tooltip"` or `aria-describedby` wiring. Note that `aria-describedby` cannot point at a `::after` pseudo-element, so the practical fix is a visually hidden `.screen-reader-text` span carrying the same string. Core ships that class on the frontend, so there is nothing new to define.

**No affordance.** Nothing signals the tooltip exists: no underline, no cursor change. A viewer who does not already suspect there is more information never hovers to find it. Undiscoverable help is close to no help.

**Stale after client navigation.** `view.js` sets its colour custom properties on `DOMContentLoaded`, so a tooltip inserted by a Query Loop page change keeps the defaults. Same pattern flagged during review of #2065.

## Scope

- Make the tooltip target focusable
- Toggle on tap for touch devices
- Emit a `.screen-reader-text` span alongside the data attribute
- Give the target a visual affordance
- Re-run or bind the colour setup so it survives client-side navigation

## Why now

#2065 wants to put an event's time in the visitor's timezone behind a tooltip. Under the current implementation that information would be reachable only by mouse hover, which is the wrong outcome for a feature whose whole point is telling people something they had not realised. That PR is blocked on this one.

Related: #2216 (language consistency), #2201 (screen reader announcements on new-tab links).

wppoland:
Fixes #2217.

### Summary of Changes
Addresses all 5 deficiencies outlined in #2217:
1. **Focusable targets:** Registered and applied `tabindex="0"` on `.gatherpress-tooltip` elements, enabling keyboard users to focus tooltips and navigate with `Tab` / `Shift+Tab`.
2. **Keyboard focus affordance:** Added `:focus-visible` outline styles in `style.scss` for clear visual cues when focused via keyboard.
3. **Touch support:** Added click/tap handling with `.gatherpress-tooltip--is-active` toggle state, outside-click dismissal, and `Escape` key dismissal.
4. **Assistive technology support:** Automatically emits a `.screen-reader-text` span alongside tooltip content so screen readers announce the text reliably.
5. **Client-side navigation resilience:** Added `focusin` / `mouseenter` delegation and a `MutationObserver` in `view.js` so tooltips rendered dynamically or via client navigation remain fully initialized.

### Verification
- **JS Unit Tests:** 79/79 unit tests pass with 100% statement, function, and line coverage across `src/formats/tooltip/`.
- **Linting:** `npm run lint:js` and `npm run lint:css` both pass cleanly with 0 errors and 0 warnings.
- **Full Test Suite:** All 62 test suites (1117 tests) in `npm run test:unit:js` pass.
- **Build:** Webpack production build compiles cleanly without errors.

## Summary by CodeRabbit

* **New Features**
  * Tooltips are now keyboard accessible and include screen-reader text.
  * Added click/tap toggling, Escape-key dismissal, and improved focus and hover behavior.
  * Tooltips dynamically inserted into the page are initialized automatically.
  * Added visible keyboard-focus outlines and touch-friendly interactions.

* **Tests**
  * Expanded coverage for tooltip accessibility, interactions, initialization, and dynamic behavior.

### DIFF
--- a/src/formats/tooltip/view.js
+++ b/src/formats/tooltip/view.js
@@ -111,19 +111,27 @@ export function closeAllTooltips() {
  * @param {MouseEvent|TouchEvent} event The click or tap event.
  */
 export function handleDocumentClick( event ) {
-	const target = event?.target instanceof Element ? event.target : null;
-	const tooltip = target
+	const rawTarget = event?.target instanceof Node ? event.target : null;
+	const target = rawTarget instanceof Element ? rawTarget : rawTarget?.parentElement;
+	const tooltip = target?.closest
 		? target.closest( '.gatherpress-tooltip[data-gatherpress-tooltip]' )
 		: null;
 
 	if ( tooltip ) {
 		initTooltip( tooltip );
-		const isCurrentlyActive = tooltip.classList.contains(
-			'gatherpress-tooltip--is-active'
-		);
+		const doc = tooltip.ownerDocument;
+		const isCurrentlyActive =
+			tooltip.classList.contains( 'gatherpress-tooltip--is-active' ) ||
+			( doc?.activeElement === tooltip &&
+				! tooltip.classList.contains( 'gatherpress-tooltip--is-dismissed' ) );
+
 		closeAllTooltips();
+
 		if ( ! isCurrentlyActive ) {
+			tooltip.classList.remove( 'gatherpress-tooltip--is-dismissed' );
 			tooltip.classList.add( 'gatherpress-tooltip--is-active' );
+		} else {
+			tooltip.classList.add( 'gatherpress-tooltip--is-dismissed' );
 		}
 	} else {
 		closeAllTooltips();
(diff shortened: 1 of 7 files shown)


src/formats/tooltip/view.js
```diff
@@ -1,46 +1,200 @@
...
+
+/**
+ * Handle Escape key press to dismiss open tooltips.
+ *
+ * @param {KeyboardEvent} event The keydown event.
+ */
+export function handleDocumentKeyDown( event ) {
+	if ( 'Escape' === event?.key ) {
+		closeAllTooltips();
+	}
```
carstingaxion, line 165: Doesn't seem to work.

Tested in the PR Playground with the tooltip of a past Online Event

wppoland: Good catch, @carstingaxion. Pushed a fix in 4ac8d8edf.

Two reasons were causing this in the preview:
1. **Focus override on toggle:** When clicking or tapping to dismiss, the element still held focus (`document.activeElement`), and CSS `:focus` was unconditionally forcing the tooltip visible. The same issue prevented the `Escape` key from dismissing the tooltip while focused. Added a `.gatherpress-tooltip--is-dismissed` state (cleared on `focusout` / `mouseleave`) and wired it to `:focus:not(.gatherpress-tooltip--is-dismissed)`, satisfying WCAG 1.4.13 dismissal without losing focus.
2. **Touch affordance:** Added `touch-action: manipulation` and pointer styling for touch devices so taps bubble and register consistently.

Unit tests and lint checks updated and passing.

carstingaxion: Tested again in the PR Playground and the ESC key works as expected. Thanks for the fix @wppoland !

carstingaxion (dismissed):

src/formats/tooltip/view.js
```diff
@@ -1,46 +1,266 @@
...
+	) {
+		const existingSrText = tooltip.querySelector(
+			':scope > .screen-reader-text'
+		);
+		if ( ! existingSrText ) {
+			const srText = document.createElement( 'span' );
+			srText.className = 'screen-reader-text';
+			srText.textContent = ` (${ tooltipText })`;
+			tooltip.appendChild( srText );
+		}
```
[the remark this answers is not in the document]
mauteri: Addressed in 50630b39.

For the record, the "theme-provided in most setups" premise isn't quite right: core ships `.screen-reader-text` on every frontend through the block-library stylesheet, which `wp_common_block_scripts_and_styles()` enqueues unconditionally, and the Theme Handbook requires themes to carry it too. The only site without it is one whose theme dequeues block styles. But the span sits inside the trigger, so on that site it would render mid-sentence, which is ugly enough to be worth ten lines.

The rule is scoped to `.gatherpress-tooltip > .screen-reader-text` rather than declared globally, so it never overrides what a theme or core defined, and its declarations match core's so behavior is identical whichever wins. `overflow-wrap` stands in for core's legacy `word-wrap` alias, which our stylelint rejects.

mauteri (approved):

PR #2228, opened 2026-08-31
Export all-day events as floating dates in calendar feeds (#2225)

### LINKED ISSUE #2225: Export all-day events as floating dates in calendar feeds
Follow-up to #966 / #2224, which added an "All day" toggle to events.

An all-day event is stored as a real full-day span in local time (`00:00:00` to `23:59:59`) so that existing `datetime_end_gmt >= now` queries keep working. That storage is right and should stay. What needs to change is how those events are **serialized** for calendar clients.

## What happens now

Every calendar export builds its timestamps the same way, from the GMT columns:

```php
$date_start = $this->event->get_formatted_datetime( 'Ymd', 'start', false );
$time_start = $this->event->get_formatted_datetime( 'His', 'start', false );
```

So an all-day event exports as a timed event running to `23:59:59`. Two problems fall out of that:

1. **It doesn't read as all-day.** Calendar clients render a block spanning the day rather than the all-day banner, and the event carries a timezone it shouldn't have. RFC 5545 all-day events are floating: no zone, no time.
2. **It can land on the wrong day.** Because the exports read the GMT columns, an all-day event on 2026-08-29 in `Asia/Tokyo` is stored GMT as `2026-08-28 15:00:00` to `2026-08-29 14:59:59` and exports with a start date of **2026-08-28**. Any event whose local zone is far enough from UTC exports a day off.

## What it should emit

Per RFC 5545 §3.6.1, an all-day event uses `VALUE=DATE` with an **exclusive** end (`DTEND` is the day *after* the last day):

```
DTSTART;VALUE=DATE:20260829
DTEND;VALUE=DATE:20260830
```

Both dates come from the **local** columns, not GMT, since all-day is floating.

## Where

- `Calendar::get_ics_file_content()` — `includes/core/classes/calendar/class-calendar.php:294-295`. Also covers Outlook, which downloads the same ICS template (`Calendar\Setup` line 156).
- `Calendar::get_google_destination_url()` — same file, ~line 154. Google's `dates` param takes the same `YYYYMMDD/YYYYMMDD` exclusive-end form for all-day.
- `Calendar::get_yahoo_destination_url()` — same file, ~line 203. Yahoo uses `st` plus a duration; its all-day form is `allday=true`.

## Notes

- Nothing about storage, queries, or `Event::get_display_datetime()` changes here. This is serialization only.
- `Event::is_all_day()` already exists and reads the `gatherpress_is_all_day` meta.
- Worth a test per exporter, including one in a non-UTC zone to pin the day-boundary case in problem 2.

wppoland:
Fixes #2225.

### Summary of Changes
Addresses calendar client export/serialization for all-day events:
1. **RFC 5545 iCal / ICS (`get_ical_event_string`):** All-day events now emit `DTSTART;VALUE=DATE:YYYYMMDD` and `DTEND;VALUE=DATE:YYYYMMDD` with an exclusive end date (the day after the last day of the event), sourced from local date values rather than GMT. This prevents timezone offset drift (e.g. in Asia/Tokyo) and ensures calendar clients render the event as an all-day banner without a timezone.
2. **Google Calendar (`get_google_destination_url`):** All-day events serialize the `dates` query parameter as `YYYYMMDD/YYYYMMDD` with exclusive end date (no time component `T...Z`).
3. **Yahoo! Calendar (`get_yahoo_destination_url`):** All-day events serialize `st=YYYYMMDD` and duration in 24-hour intervals (`dur=2400` for 1 day, `4800` for 2 days).

### Verification
- **PHP Unit Tests:** Added 3 unit tests in `Test_Calendar`:
  - `test_get_google_destination_url_with_all_day_event` (1-day and multi-day exclusive end date)
  - `test_get_yahoo_destination_url_with_all_day_event` (1-day and 2-day duration)
  - `test_get_ical_event_string_with_all_day_event` (verifies local floating date without GMT boundary drift on non-UTC timezone Asia/Tokyo)
- **Coding Standards:** `phpcs` passes cleanly with 0 errors and 0 warnings on `class-calendar.php` and `class-test-calendar.php`.
- **Static Analysis:** `phpstan` passes with 0 errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved calendar exports for all-day events across Google Calendar, Yahoo Calendar, and iCal.
  * All-day events now use date-only values, correct exclusive end dates, and accurate whole-day durations.
  * Prevented timezone-related date shifts in iCal exports.

* **Tests**
  * Added coverage confirming correct all-day event formatting for all supported calendar formats.

### DIFF
--- a/includes/core/classes/calendar/class-calendar.php
+++ b/includes/core/classes/calendar/class-calendar.php
@@ -211,15 +211,28 @@ final class Calendar {
 		}
 
 		if ( $this->event->is_all_day() ) {
-			// Yahoo takes two plain dates and its own flag, not a duration.
-			// `dur` is an `hhmm` field, and a single all-day event already
-			// needs more hours than the two digits hold. Local dates, since
-			// an all-day date is floating and must not shift with the zone.
-			$timing = array(
-				'st'     => $this->event->get_formatted_datetime( 'Ymd', 'start' ),
-				'et'     => $this->event->get_formatted_datetime( 'Ymd', 'end' ),
-				'allday' => 'true',
-			);
+			// Yahoo has two shapes for this and they cannot be combined: it
+			// ignores `dur` the moment `et` is present. A single day says
+			// `dur=allday` and sends no end at all; a span sends the end and
+			// no duration. That end is exclusive, so it names the day after
+			// the last one, as `DTEND` does in the feed above.
+			//
+			// Local dates throughout, an all-day date being floating.
+			$date_start = $this->event->get_formatted_datetime( 'Ymd', 'start' );
+
+			if ( $this->event->is_same_date() ) {
+				$timing = array(
+					'st'  => $date_start,
+					'dur' => 'allday',
+				);
+			} else {
+				$end_date = $this->event->get_formatted_datetime( 'Y-m-d', 'end' );
+
+				$timing = array(
+					'st' => $date_start,
+					'et' => gmdate( 'Ymd', (int) strtotime( $end_date . ' +1 day' ) ),
+				);
+			}
 		} else {
 			$date_start = $this->event->get_formatted_datetime( 'Ymd', 'start', false );
 			$time_start = $this->event->get_formatted_datetime( 'His', 'start', false );
--- a/includes/core/classes/calendar/class-calendar.php
+++ b/includes/core/classes/calendar/class-calendar.php
@@ -192,7 +192,7 @@ final class Calendar {
 	 * Off-site destination URL for the Yahoo! Calendar redirect.
 	 *
 	 * Opens Yahoo! Calendar's event-creation form pre-filled with this
-	 * event's title, start time, duration, location, and description.
+	 * event's title, start and end time, location, and description.
 	 * Called from `Calendar\Setup::queried_event_yahoo_url()` to produce
 	 * the 302 target for the `/event/<slug>/yahoo-calendar/` endpoint —
 	 * front-end code should use `get_yahoo_url()` (the on-site URL) instead.
@@ -234,21 +234,15 @@ final class Calendar {
 				);
 			}
 		} else {
-			$date_start = $this->event->get_formatted_datetime( 'Ymd', 'start', false );
-			$time_start = $this->event->get_formatted_datetime( 'His', 'start', false );
-
-			// Figure out duration of event in hours and minutes: hhmm format.
-			$diff_start = $this->event->get_formatted_datetime( $this->event::DATETIME_FORMAT, 'start', false );
-			$diff_end   = $this->event->get_formatted_datetime( $this->event::DATETIME_FORMAT, 'end', false );
-			$duration   = ( ( strtotime( $diff_end ) - strtotime( $diff_start ) ) / 60 / 60 );
-			$full       = intval( $duration );
-			$fraction   = ( $duration - $full );
-			$hours      = str_pad( strval( $duration ), 2, '0', STR_PAD_LEFT );
-			$minutes    = str_pad( strval( $fraction * 60 ), 2, '0', STR_PAD_LEFT );
-
+			// Yahoo has no timezone parameter and does not honor a trailing
+			// Z: whatever it is handed is shown verbatim as local wall-clock
+			// time. So the event's own local time goes out, right for anyone
+			// in its zone and the closest available for anyone else. The end
+			// is sent rather than a duration: `dur` is an `hhmm` field that
+			// tops out at 99 hours, and Yahoo ignores it once `et` is present.
 			$timing = array(
-				'st'  => sprintf( '%sT%sZ', $date_start, $time_start ),
-				'dur' => $hours . $minutes,
+				'st' => $this->event->get_formatted_datetime( 'Ymd\THis', 'start' ),
+				'et' => $this->event->get_formatted_datetime( 'Ymd\THis', 'end' ),
 			);
 		}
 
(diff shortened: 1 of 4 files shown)


includes/core/classes/calendar/class-calendar.php
```diff
@@ -259,13 +279,24 @@ public function get_ical_event_string(): string {
...
-		$time_start     = $this->event->get_formatted_datetime( 'His', 'start', false );
-		$date_end       = $this->event->get_formatted_datetime( 'Ymd', 'end', false );
-		$time_end       = $this->event->get_formatted_datetime( 'His', 'end', false );
-		$datetime_start = sprintf( '%sT%sZ', $date_start, $time_start );
-		$datetime_end   = sprintf( '%sT%sZ', $date_end, $time_end );
-		$modified_gmt   = strtotime( $post->post_modified_gmt );
+		if ( $this->event->is_all_day() ) {
+			$date_start   = $this->event->get_formatted_datetime( 'Ymd', 'start' );
+			$end_date_str = $this->event->get_formatted_datetime( 'Y-m-d', 'end' );
+			$date_end     = gmdate( 'Ymd', (int) strtotime( $end_date_str . ' +1 day' ) );
```
carstingaxion, line 297: Will "End date" + 1 day not be too late?
Shouldn't that be "Start date" + 1 day?

carstingaxion (commented):
~~Only found little thing to look at.~~

In general, this looks good.

includes/core/classes/calendar/class-calendar.php
wppoland: Under RFC 5545 Section 3.6.1, `DTEND` with `VALUE=DATE` (all-day events) is non-inclusive (exclusive): it specifies the calendar date following the event.

- For a single-day all-day event (e.g. October 5): `DTSTART:20261005` and `DTEND:20261006` (`end_date + 1 day`).
- For a multi-day all-day event (e.g. October 5 to October 7): `DTSTART:20261005` and `DTEND:20261008` (`end_date + 1 day`).

If we used `start_date + 1 day`, any multi-day all-day event would be cut short to 1 day on the calendar.

carstingaxion: Thanks for the update @wppoland .

That makes sense and as such should not be changed. Thanks for providing the facts in an understandable way.

carstingaxion (dismissed):

mauteri:
@wppoland @carstingaxion heads up on two things here.

**First, I pushed a commit to this branch** (41b08655) addressing CodeRabbit's note about the Yahoo all-day flag. The all-day branch now sends `allday=true` with `st` and `et` as plain local dates, and no longer sends `dur` at all.

The missing flag was the smaller half. `dur` is an `hhmm` field, so `days * 24` was emitting `dur=2400` for a one-day event and would have asked for `12000` on a five-day one, which is more hours than the two digits hold. Yahoo takes a start and an end date for this instead, and note that its `et` is the last day itself rather than the day after, unlike the exclusive end the ICS feed correctly uses. I updated the all-day test to match and added a `dur` assertion to the timed test, which did not have one.

**Second, and separately: the timed Yahoo duration is broken on `main` and always has been.** This is not something this PR introduced, but it lives about ten lines from what you are both looking at, so it is worth knowing before someone builds on it.

```php
$hours = str_pad( strval( $duration ), 2, '0', STR_PAD_LEFT );
```

That pads `$duration`, which is a float, rather than `$full`, which is the integer hour count computed on the line above it. Whole-hour events come out right, which is why nobody noticed. Anything fractional does not:

```
1.5 hour event  ->  dur=1.530     (expected 0130)
2.5 hour event  ->  dur=2.530     (expected 0230)
```

1.5 hours is one of our default duration presets, so this is not an edge case. A long event compounds it, since the two-digit pad also stops helping past 99 hours. A real five-day timed event on my local site produces:

```
dur=119.9997222222259.983333333334
```

I have deliberately left that alone in this PR. It is a pre-existing bug in released code, it affects the timed path rather than the all-day one, and folding it in here would mix an unrelated fix into a change that is already about serialization. Happy for it to be its own issue.

One more thing on verification, and this applies to all three providers rather than just Yahoo. Every test we have asserts the URL or the ICS string we built, which only proves we emitted what we meant to emit. It does not prove Google or Yahoo accept it. Before this merges, somebody should click through both links once and confirm the event lands as an all-day banner on the correct dates, ideally starting from a non-UTC timezone, since day drift is the whole reason #2225 exists. That check is quick and it is the only part of this that a unit test genuinely cannot cover.

mauteri:
Correction to my comment above: the two timed-path Yahoo bugs are now fixed **in this PR** rather than left for a separate one, since they're the same export and the same parser.

What landed (ec7cf4d8):

- Timed events send `st` and `et` in the event's own local time with no `Z`. Yahoo has no timezone parameter and stores the value verbatim as local wall clock, so a 19:00 Tokyo event used to go out as `10:00Z` and land at 10:00.
- `dur` is gone from the timed path. Yahoo ignores it once `et` is present, which also removes the `dur=1.530` float bug and the 99-hour cap.

Both are read from Yahoo's own query parser (`qm()` in `bundle_epoch_calendar.js`), not inferred. One thing that parser makes unavoidable: `DUR=allday` is the only value that sets the all-day flag, and it is ignored whenever `ET` is present, so a **multi-day** all-day event cannot be pre-flagged in Yahoo's composer. It opens with the correct dates and the toggle off. A one-day all-day event opens with the toggle on.

@wppoland @carstingaxion

mauteri (approved):

PR #2229, opened 2026-08-31
Add event operational status support: scheduled, cancelled, postponed, rescheduled, moved online (#2056)

### LINKED ISSUE #2056: No way to cancel or postpone an event

wppoland:
Fixes #2056.

### Summary of Changes

Adds support for setting and displaying an event's operational status (`scheduled`, `cancelled`, `postponed`, `rescheduled`, `moved-online`):

1. **Storage & Data Model:**
   - Registers `gatherpress_status` post meta in `Event\Meta` with REST API support and enum validation (`scheduled`, `cancelled`, `postponed`, `rescheduled`, `moved-online`).
   - Adds status constants and helper methods to `Event` class (`get_status()`, `is_cancelled()`, `is_postponed()`, `get_status_label()`, `get_schema_event_status()`, `get_ical_status()`).

2. **Calendar Feeds (RFC 5545):**
   - Emits `STATUS:<CONFIRMED|CANCELLED|TENTATIVE>` property in `Calendar::get_ical_event_string()`. Subscribed calendar clients (Apple Calendar, Google Calendar, Outlook) receive standard cancellation and tentative indicators.

3. **Editor UI:**
   - Adds `<EventStatus />` select control component in the Event Settings sidebar panel under Date & Time.

4. **Blocks & Frontend Display:**
   - Adds new `gatherpress/event-status` block with `badge` (default) and `notice` style variations, with option to hide when status is `scheduled`.
   - Adds status-aware CSS classes (`gatherpress-event-date--is-cancelled`, `gatherpress-event-date--is-postponed`) and subtle strikethrough styling to the `gatherpress/event-date` block.

### Verification
- **PHPUnit Tests:** Added unit tests covering all status methods, constants, and block validation (`Test_Event`, `Test_Event_Status`).
- **Jest Unit Tests:** Added unit tests covering `EventStatus` component and `gatherpress/event-status` Edit block component (100% passing across all 64 JS suites).
- **Coding Standards:** `phpcs`, `phpstan`, `lint:js`, and `lint:css` all pass with 0 errors and 0 warnings.

## Summary by CodeRabbit

* **New Features**
  * Added event statuses: Scheduled, Cancelled, Postponed, Rescheduled, and Moved Online.
  * Added an Event Status selector to event settings.
  * Added an Event Status block with badge and notice display styles.
  * Added options to hide the status when an event is scheduled.
  * iCalendar exports now include event status information.

* **UI Improvements**
  * Cancelled and postponed event dates display with strikethrough styling and reduced opacity.
  * Status badges use distinct colors and animations, with reduced-motion support.


carstingaxion:
Thanks for your contribution here @wppoland !

Just from reading through the description, I like that this PR has a lot from my gatherpress-event-status experiment.

But I'm asking myself, why you choose postmeta over a taxonomy, like I did. What are your reasons to use meta here? It seems to be the wrong way, given that *I think* events might be queried by that status.

mauteri:
> Thanks for your contribution here @wppoland !
> 
> Just from reading through the description, I like that this PR has a lot from my gatherpress-event-status experiment.
> 
> But I'm asking myself, why you choose postmeta over a taxonomy, like I did. What are your reasons to use meta here? It seems to be the wrong way, given that _I think_ events might be queried by that status.

We should use taxonomy here as meta queries will have a hard time scaling for larger sites. Thx!

wppoland:
@carstingaxion Fair challenge, and you were right. Done in 95aeadff5.

My reason for meta was narrow: status is a single-valued scalar on the event, the same shape as `gatherpress_is_all_day`, and `register_post_meta()` with an `enum` in `show_in_rest` gives validation and an editor binding for free. That is a real convenience, but it is a convenience on the write path, and it optimizes the wrong side.

The read path is what matters, and you named it. There was no query by status in the PR, which is the problem rather than a defence: `Event\Query::get_events_list()` already filters by Topic and Venue through `tax_query`, so the moment anyone wants "hide cancelled events from the list", or a status filter on the events block, meta forces either a `meta_query` against a non-indexed `meta_value` or a second store. Better to be in the right place before that lands than to migrate afterwards. It is also the same call @mauteri made on #2064 for RSVP check-in.

## What changed

`_gatherpress_event_status` on `Event::POST_TYPE`, hidden the way the RSVP taxonomies are: `public => false`, `show_ui => false`, `publicly_queryable => false`, `rewrite => false`. There is no useful `/event-status/cancelled/` archive.

`default_term => scheduled`, so every saved event carries a status term and filtering for one is an `IN` query rather than a `NOT EXISTS`. `get_status()` keeps a no-term fallback, since `default_term` only applies on insert and says nothing about events stored before the taxonomy existed.

`Event::set_status()` passes a plain string to `wp_set_object_terms()`, which replaces rather than appends. That is what enforces the statuses being mutually exclusive, rather than a guard I have to remember to write. `STATUSES`, `get_status_label()`, `get_schema_event_status()` and `get_ical_status()` are pure functions of the slug, so they carried over untouched.

## On the editor question I raised

I asked which shape you expected and then found a third one that avoids the tradeoff, so the question is moot.

The taxonomy is kept **out of REST** deliberately. A taxonomy reaches the block editor as an array of term ids, which is the round trip I did not want for what is really a `SelectControl`. Instead the editor reads and writes a plain slug through a `gatherpress_status` REST field, registered in `Event\Rest_Api`, whose `update_callback` checks `edit_post` on the event and routes through `Event::set_status()`. So the editor keeps the simple contract meta gave it, the taxonomy is the query surface, and there is exactly one write path with no second store to drift.

Shout if you would rather the taxonomy were exposed over REST as well; it is a one-line change, I just could not see who would consume it.

## Verification

- PHP single site: 1924 tests, 7509 assertions, OK
- PHP multisite: 325 tests, 869 assertions, OK
- JS: 1108 tests, OK
- Coverage on every changed PHP file, merging both clover runs the way the CI check does: 100%
- `phpcs` against `phpcs.ruleset.xml`, PHPStan, and `lint:js`: clean

wppoland:
Fixed in 178c0c67c.

The one red check was coverage: `class-event.php` at 99.73%, a single uncovered line 657. That is the guard at the top of `get_status()`:

```php
public function get_status(): string {
	if ( ! $this->post ) {
		return self::STATUS_SCHEDULED;
	}
```

`Event::__construct()` only assigns `$this->post` when the id belongs to a post type supporting `gatherpress-event-date`, so the line is reached through `new Event( 0 )`, which is already how the other no-post guards in this class are covered (`is_all_day()`, `get_timezone_preference()`, and two more).

Verified locally with a clover run rather than by eye: line 657 now has a hit count of 1 and `class-event.php` reports 335/335, 100%. Full suite green at 1921 tests, 7487 assertions.

includes/core/classes/event/class-event.php
```diff
@@ -589,6 +656,140 @@ public function is_all_day(): bool {
...
+		// and default_term only applies on insert. No term means scheduled.
+		$terms = get_the_terms( $this->post->ID, self::TAXONOMY_STATUS );
+
+		if ( ! is_array( $terms ) || empty( $terms ) ) {
+			return self::STATUS_SCHEDULED;
+		}
+
+		$status = (string) $terms[0]->slug;
+
+		return in_array( $status, self::STATUSES, true ) ? $status : self::STATUS_SCHEDULED;
```
carstingaxion: I like to discuss the "exclusivenes" of event statuses (_or stati ?_) in general.

While experimenting with `gatherpress-event-status`, I found out that it could make sense to assign multiple status terms.

> ## [Can an event have multiple statuses?](https://github.com/carstingaxion/gatherpress-event-status#can-an-event-have-multiple-statuses)
>Yes, the taxonomy supports multiple term assignments. For example, an event could be both "Rescheduled" and "Moved Online" if both apply.

Why did you decide for explicit singularity?

src/blocks/event-status/style.scss
```diff
@@ -0,0 +1,194 @@
...
+			border-color: #1a73e8;
+			color: #1a73e8;
+		}
+
+		&.gatherpress-event-status--is-moved-online {
+			background-color: #f3e8fd;
+			border-color: #7627bb;
+			color: #7627bb;
+		}
+	}
```
carstingaxion: Instead of hardcoding all the color values, I'd like to suggest to use a CSS custom prop for the primary color (maybe background-color) per status and derive all the other colors via CSS color functions from that primary-prop.

(just a note: We did something similar just recently for the map marker colors)

src/blocks/event-date/render.php
carstingaxion, line 1: Instead of changing one particular block, inject the CSS class into WP core post class filter. This allows to use it here on the event-date block but also for all others. And it will be available on archive views, on singualr views and inside query loops.

_The post class filter would use the taxonomy by default, if it where public. But as non-public, we need to add this by hand._

carstingaxion (changes requested):

src/blocks/event-status/style.scss
wppoland: Great suggestion! Refactored in 63e395bef to use CSS custom properties (`--gatherpress-status-color`, `--gatherpress-status-bg`, `--gatherpress-status-border`) per status modifier, removing all the duplicated style rules.

src/blocks/event-date/render.php
wppoland: Implemented in 63e395bef. Added `add_status_post_class()` hooked to the `post_class` filter in `Event\Setup`, so non-scheduled statuses append `gatherpress-event-status--is-{status}` to the post wrapper across archives, query loops, and single templates, and updated event-date styles and unit tests accordingly.

includes/core/classes/event/class-event.php
wppoland: Good question! Singularity was chosen primarily because:
1. Standards compatibility: Schema.org `eventStatus` (`EventStatusType`) and iCalendar (RFC 5545 `STATUS`) are strictly scalar properties representing one operational state.
2. Clear UX: In the inspector sidebar, a `SelectControl` dropdown prevents conflicting or contradictory selections (like Cancelled together with Scheduled).

However, because the underlying data is stored as a taxonomy (`gatherpress_event_status`), multiple term assignments are naturally supported by the taxonomy storage itself. If we want to allow combinations like Rescheduled + Moved Online in the UI, we can either expand the select control to multi-select / checkboxes or introduce a combined status. What do you think?

mauteri:
@coderabbitai review

includes/core/classes/event/class-setup.php
```diff
@@ -119,6 +120,79 @@ protected function setup_hooks(): void {
...
+	 */
+	public function add_status_post_class( array $classes, array $css_classes, int $post_id ): array {
+		if ( Event::POST_TYPE !== get_post_type( $post_id ) ) {
+			return $classes;
+		}
+
+		$event  = new Event( $post_id );
+		$status = $event->get_status();
+
+		if ( Event::STATUS_SCHEDULED !== $status ) {
```
carstingaxion: So the default status will never have a status post class, but the CSS (in this PR) references this class.

src/blocks/event-status/style.scss
```diff
@@ -0,0 +1,168 @@
...
+			background-color: var(--gatherpress-status-color);
+		}
+
+		&:hover {
+			transform: translateY(-1px);
+			box-shadow: 0 2px 6px rgb(0 0 0 / 8%);
+		}
+	}
+
+	&.gatherpress-event-status--is-scheduled {
```
carstingaxion: This CSS will never be used, because the CSS class is never written to any page.

carstingaxion (changes requested):

includes/core/classes/event/class-setup.php
wppoland: Good catch! Updated in commit ccc7f389d to unconditionally add the status post class for all statuses (including `gatherpress-event-status--is-scheduled`), ensuring every event carries its status class across loops, archives, and templates.

src/blocks/event-status/style.scss
wppoland: Removed the redundant `&.gatherpress-event-status--is-scheduled` declaration in commit ccc7f389d since the root `.wp-block-gatherpress-event-status` rule already defines those scheduled green variables as defaults.

mauteri:
@wppoland heads up, #2270 just merged and it renames the post type supports from string literals to class constants:

| Literal | Constant |
| --- | --- |

This branch has one line using the literal form, a `post_type_supports()` call with `gatherpress-event-date`. Could you merge `develop` in and swap it for `Event::SUPPORT`? The value is unchanged, so it is just the constant and a `use` statement if one is not already there.

wppoland:
@mauteri Merged `develop` in and swapped `gatherpress-event-date` for `Event::SUPPORT` in `Event_Status::validate_event()`. All tests and linters pass cleanly.

includes/core/classes/event/class-setup.php
wppoland: Good catch! Updated so that the status post class is now added for all statuses including scheduled (gatherpress-event-status--is-scheduled).

src/blocks/event-status/style.scss
wppoland: Updated in class-setup.php so the class is now always attached, matching the stylesheet selector.

includes/core/classes/event/class-event.php
```diff
@@ -589,6 +656,140 @@ public function is_all_day(): bool {
...
+	public function set_status( string $status ): bool {
+		if ( ! $this->post || ! in_array( $status, self::STATUSES, true ) ) {
+			return false;
+		}
+
+		// A plain string replaces every term in the taxonomy, which is what
+		// enforces the statuses being mutually exclusive.
+		$result = wp_set_object_terms( $this->post->ID, $status, self::TAXONOMY_STATUS );
+
+		return ! is_wp_error( $result );
```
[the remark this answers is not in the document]
mauteri: Fixed in d00b4fbbe. `set_status()` now touches the post row after writing the term, so the sequence advances and a client accepts the change. There is a regression test in the calendar suite that watches SEQUENCE and STATUS across a cancellation; I checked it fails without the fix.

includes/core/classes/event/class-event.php
```diff
@@ -589,6 +656,140 @@ public function is_all_day(): bool {
...
+	 * Returns the human-readable translated label for the current status.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @return string Localized status label.
+	 */
+	public function get_status_label(): string {
+		$labels = array(
+			self::STATUS_SCHEDULED    => __( 'Scheduled', 'gatherpress' ),
+			self::STATUS_CANCELLED    => __( 'Cancelled', 'gatherpress' ),
```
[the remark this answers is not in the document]
mauteri: Done, and then taken further. Every status is spelled the American way now, including the stored slug, which is `canceled`. Schema.org's `EventCancelled` and RFC 5545's `CANCELLED` keep their own spelling, since those are their words rather than ours.

src/blocks/event-status/edit.js
```diff
@@ -0,0 +1,94 @@
...
+ * @param {Object}   props               Component props.
+ * @param {Object}   props.attributes    Block attributes.
+ * @param {Function} props.setAttributes Function to set attributes.
+ * @param {Object}   props.context       Block context.
+ *
+ * @return {JSX.Element} The rendered component.
+ */
+const Edit = ( { attributes, setAttributes, context } ) => {
+	const { hideScheduled } = attributes;
+	const postId = context.postId;
```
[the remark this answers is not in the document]
mauteri: Fixed in d00b4fbbe, then overtaken: the custom block is gone. The status now displays through an Event Status variation of core's Post Terms block, which takes its post from core's own context handling.

src/blocks/event-status/edit.js
```diff
@@ -0,0 +1,94 @@
...
+				return (
+					select( 'core/editor' )?.getEditedPostAttribute(
+						'gatherpress_status'
+					) || 'scheduled'
+				);
+			}
+
+			const post = select( 'core' ).getEntityRecord(
+				'postType',
+				context.postType || 'gatherpress_event',
```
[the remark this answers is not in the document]
mauteri: The branches you listed were covered in d00b4fbbe, and the file has since been removed. Display moved to a variation of core's Post Terms block, so there is no GatherPress edit component left to branch.

test/unit/php/includes/tests/core/classes/blocks/class-test-event-status.php
```diff
@@ -0,0 +1,89 @@
...
+		$event_post = $this->mock->post( array( 'post_type' => Event::POST_TYPE ) )->get();
+		$block      = array(
+			'blockName' => Event_Status::BLOCK_NAME,
+			'attrs'     => array( 'postId' => $event_post->ID ),
+		);
+
+		$content = '<div class="wp-block-gatherpress-event-status">Cancelled</div>';
+		$result  = $instance->validate_event( $content, $block );
+
+		$this->assertSame( $content, $result );
```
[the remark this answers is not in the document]
mauteri: Partly done, partly overtaken. The non-viewable event case and the render branches were covered in d00b4fbbe. The event-status block has since been replaced by a variation of core's Post Terms block, so those tests went with it; the event-date wrapper branches you asked for are still covered.

test/unit/php/includes/tests/core/classes/event/class-test-event.php
```diff
@@ -2433,4 +2433,189 @@ public function data_timezone_precedence(): array {
...
+	 * @return void
+	 */
+	public function test_get_status(): void {
+		$post  = $this->mock->post( array( 'post_type' => Event::POST_TYPE ) )->get();
+		$event = new Event( $post->ID );
+
+		// Default status is scheduled.
+		$this->assertSame( Event::STATUS_SCHEDULED, $event->get_status() );
+		$this->assertFalse( $event->is_cancelled() );
+		$this->assertFalse( $event->is_postponed() );
```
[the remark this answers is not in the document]
mauteri: Pushing back on this one. The no-term fallback is already reached. `test_get_status` clears the taxonomy with `wp_delete_object_term_relationships()` and then asserts the fallback, a few lines below the assertion you flagged. The line you pointed at exercises the in-vocabulary path instead, which is also worth keeping.

test/unit/php/includes/tests/core/classes/event/class-test-rest-api.php
```diff
@@ -88,6 +94,83 @@ public function test_setup_hooks(): void {
...
+		$this->assertSame(
+			Event::STATUS_SCHEDULED,
+			call_user_func( $field['get_callback'], array( 'id' => $post->ID ) ),
+			'Failed to assert an untouched event reads as scheduled.'
+		);
+
+		// An editor of the event may set the status.
+		$this->mock->user( 'admin' );
+
+		call_user_func( $field['update_callback'], Event::STATUS_CANCELLED, $post );
```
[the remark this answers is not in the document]
mauteri: Done in d00b4fbbe. There is a test that dispatches authenticated read and write requests against the real route, asserts the stored status changes, and asserts an unknown status is refused with a 400 while the stored one is left alone.

src/blocks/event-status/style.scss
mauteri: This is how it works now, and it went further than one property per status. Each status declares a single color in `Event\Status`, and everything else is derived from it with `color-mix()`, the way the tooltip format does. The stylesheet names no status at all: `Event\Setup` writes one custom property per status from the registry, so a status a site registers through the filter gets its own color without shipping CSS.

src/blocks/event-date/render.php
mauteri: Done, and it turned out to be the thing that made the rest possible. `add_status_post_class()` puts the status on the post wrapper through the `post_class` filter, so it is there in archives, query loops and single templates. That class is now also what the status colors key off, which is why the stylesheet needs to know nothing about any particular status.

includes/core/classes/event/class-setup.php
mauteri: Fixed. Every event carries a status class, the default included.

src/blocks/event-status/style.scss
mauteri: Resolved by removing the file. The event-status block has been replaced by a variation of core's Post Terms block, and its stylesheet names no status, so there is no unreachable selector left to write.

mauteri:
@wppoland thanks for your patience on this. I've pushed a fair bit to your branch and want to be upfront that I'm still weighing the shape of it.

**What I changed and why**

The big one: the custom Event Status block is gone, replaced by a variation of core's Post Terms block. The status is a term on the event, and core already has a block that renders a post's terms. That took about 900 lines out, and we inherit every typography, color, spacing and border control core maintains, plus correct rendering if an event ever holds more than one status. For that to work the taxonomy is now publicly queryable and REST visible, because core refuses to render a taxonomy that isn't viewable. It stays out of the admin UI, since I want the vocabulary coming from code rather than from people adding terms.

Statuses now live in `Event\Status` behind a `gatherpress_event_statuses` filter, and nothing names a status in code anymore. An entry declares its label, description, color, Schema.org value, iCal value, and any post type support it depends on. I'd rather companion plugins and site owners extend this than us adding statuses forever.

Smaller: `moved-online` is just `moved` now, since an event can move venue and not only online. And we use American spelling, so the slug is `canceled`.

**What I'm still unsure about**

Whether this needs to be as much as it is. Stripped back, it's a taxonomy and something that displays it. The parts I'm confident are worth keeping are the calendar and structured data: the Schema.org mapping, the iCal STATUS, and the fix so a cancellation actually reaches people who already subscribed to the event.

**Open questions**

@carstingaxion your multi-status question is still open, and it got smaller. Display is free now that Post Terms renders a term list. What's left is whether `set_status()` stops replacing terms, and which single status the calendar and Schema.org hear when several apply, since both are scalar. I'm leaning toward each status declaring a priority so the vocabulary decides rather than us hardcoding an order.

I'm also wondering whether we're missing "tentative". iCal has it natively and Schema.org has no equivalent, so it's the one state the standards support that we don't currently offer.

wppoland:
Thanks @mauteri, this refactoring is a huge win. Replacing the bespoke block with a `core/post-terms` variation cuts down massive maintenance overhead, gives us block-theme design controls (typography, colors, borders) for free, and handles multi-term rendering naturally. Keeping the taxonomy hidden from the admin menu while exposing it to REST / queries hits the sweet spot for a code-driven vocabulary.

Here are thoughts on the open questions:

#### 1. Multi-status resolution & scalar consumers (iCal & Schema.org)
- **Priority key in status definition**: Adding a `priority` attribute (e.g. integer, where higher priority wins) to each status entry in `Status::all()` is much cleaner than hardcoding precedence. It allows companion plugins or custom statuses added via `gatherpress_event_statuses` to position themselves properly.
- **Default precedence order**:
  1. `canceled` (highest priority - if an event is canceled, calendar clients and search engines must know it is not taking place, regardless of any other status).
  2. `postponed` / `rescheduled`.
  3. `moved`.
  4. `scheduled` (lowest priority - base state).
- **API behavior**: `Event::get_statuses()` can return all assigned terms (for templates, badges, and Post Terms display), while `Event::get_status()` returns the highest-priority status for scalar consumers (iCal `STATUS` and Schema.org `eventStatus`).
- **`set_status()`**: Could accept a single slug to set as the primary status (replacing terms by default for the standard single-select UI), with an optional parameter or companion method (`add_status()`) if an organizer or plugin needs to attach secondary statuses.

#### 2. Adding "Tentative" status
- Adding **Tentative** makes a lot of sense:
  - Organizers frequently publish "save the date" announcements or preliminary meetups before speaker or venue contracts are finalized.
  - iCal has native support (`STATUS:TENTATIVE`).
  - For Schema.org, falling back to `EventScheduled` (or `EventPostponed` if unconfirmed) matches standard practice since Schema.org lacks an explicit tentative enum.
  - Proposed definition:
    ```php
    'tentative' => array(
        'label'       => __( 'Tentative', 'gatherpress' ),
        'description' => __( 'Event is planned provisionally and awaiting confirmation.', 'gatherpress' ),
        'color'       => '#ea8600',
        'schema'      => 'EventScheduled',
        'ical'        => 'TENTATIVE',
        'priority'    => 20,
    ),
    ```

#### 3. Vocabulary & Naming
- The switch to US spelling (`canceled`) and shortening `moved-online` to `moved` (coupled with the `supports` check on venue/online capabilities) makes the taxonomy significantly more versatile and consistent.

Let me know if you would like me to push a commit adding the `priority` keys and the `tentative` status definition with corresponding unit tests, or if you prefer to finalize that part directly.

mauteri:
@wppoland thanks, I like these ideas. Please continue with them.

wppoland:
@mauteri All set! I have implemented the proposed ideas and pushed the updates in commits 0973bc83 and 9eed9e07:

1. **Tentative Status**:
   - Added `tentative` to `Status::all()` mapped to iCal `STATUS:TENTATIVE`, Schema.org `EventScheduled`, color `#ea8600`, and priority 20.
2. **Priority-Based Scalar Resolution**:
   - Status definitions now declare an integer `priority` key in `Status::all()`.
   - `Event::get_status()` resolves the highest-priority status for scalar consumers (Schema.org `eventStatus` and iCal `STATUS`), following precedence: `canceled` (priority 100) > `postponed` (priority 80) > `rescheduled` (priority 60) > `moved` (priority 40) > `tentative` (priority 20) > `scheduled` (priority 10).
   - `Event::get_statuses()` returns all assigned status terms for multi-term display in templates and the Post Terms variation block.
3. **Tests & Coverage**:
   - Added comprehensive unit tests for multi-status priority resolution, the `tentative` status mapping, and verified 100% test coverage across all touched files.

All 18 CI checks are passing cleanly and green. Ready for your review whenever you have a moment!

carstingaxion (changes requested):
Wow, the changes you both made to the idea are amazing; thank you @wppoland and @mauteri !

I'm currently testing in the PR Playground

- The frontend styles differ from the backend styles, where the prefix and suffix retrieve the styling for the block
- in opposite, selected block styles (in the editor) are not respected, but overriden with the custom styles from the block Image Image

includes/core/classes/event/class-setup.php
```diff
@@ -120,6 +122,165 @@ protected function setup_hooks(): void {
...
+		);
+
+		// Registered here rather than through wp_enqueue_block_style(), which
+		// defers registration to render time and would drop the colors below.
+		wp_register_style(
+			$handle,
+			plugins_url( 'build/variations/core/post-terms/style-index.css', GATHERPRESS_CORE_FILE ),
+			array(),
+			$asset['version']
+		);
```
carstingaxion, line 255: This does not result in a selectable block style as you can see from my earlier screenshot.

includes/core/classes/event/class-setup.php
```diff
@@ -120,6 +122,165 @@ protected function setup_hooks(): void {
...
+			$rules .= sprintf(
+				'.gatherpress-event-status--is-%s{--gatherpress-status-color:%s}',
+				sanitize_html_class( (string) $gatherpress_slug ),
+				$color
+			);
+		}
+
+		if ( '' !== $rules ) {
+			wp_add_inline_style( $handle, $rules );
+		}
```
carstingaxion, line 283: And as a side-effect, the colors are never used anywhere. I was really looking for them ...

includes/core/classes/event/class-status.php
```diff
@@ -0,0 +1,374 @@
...
+				'priority'    => 30,
+			),
+			'moved'       => array(
+				'label'       => __( 'Moved', 'gatherpress' ),
+				'description' => __(
+					'Event is taking place somewhere else, online or at another venue.',
+					'gatherpress'
+				),
+				'color'       => '#7627bb',
+				'schema'      => 'EventScheduled',
```
carstingaxion, line 113: This seems to be wrong. 'Moved-online' which was the previous name here was derived from the schema.org notation, so this should not be `EventScheduled` and only maybe `CONFIRMED` here. Please review this one again!

includes/core/classes/event/class-status.php
```diff
@@ -0,0 +1,374 @@
...
+		 *
+		 * The first status in the list is what an event with no status of its
+		 * own reports, so order matters.
+		 *
+		 * @since 0.36.0
+		 *
+		 * @param array<string, array<string, mixed>> $statuses  The statuses, keyed by slug.
+		 * @param string                              $post_type Post type they are offered for, or empty for all.
+		 */
+		return (array) apply_filters( 'gatherpress_event_statuses', $statuses, $post_type );
```
carstingaxion, line 179: I'm wondering why phpstan is not yelling at this line. I thought it is not allowed to convert mixed to array, but ok ...

includes/core/classes/event/class-status.php
```diff
@@ -0,0 +1,374 @@
...
+	 *
+	 * @param string $slug The status slug.
+	 *
+	 * @return string The color, or an empty string when there is none to use.
+	 */
+	public static function color( string $slug ): string {
+		$color = (string) ( self::get( $slug )['color'] ?? '' );
+
+		if (
+			preg_match( '/^#[0-9a-f]{3,8}$/i', $color )
```
carstingaxion, line 271: could also use https://developer.wordpress.org/reference/functions/sanitize_hex_color/ or https://developer.wordpress.org/reference/functions/sanitize_hex_color_no_hash/

includes/core/classes/event/class-status.php
```diff
@@ -0,0 +1,374 @@
...
+	 * @since 0.36.0
+	 *
+	 * @param string $post_type Post type to offer statuses for, or empty for all of them.
+	 *
+	 * @return array<string, array<string, mixed>> The statuses, keyed by slug.
+	 */
+	public static function all( string $post_type = '' ): array {
+		$statuses = array(
+			'scheduled'   => array(
+				'label'       => __( 'Scheduled', 'gatherpress' ),
```
carstingaxion, line 72: Use _x and add a proper translator comment before

includes/core/classes/event/class-status.php
```diff
@@ -0,0 +1,374 @@
...
+				'description' => __( 'Event is planned and confirmed to take place.', 'gatherpress' ),
+				'color'       => '#137333',
+				'schema'      => 'EventScheduled',
+				'ical'        => 'CONFIRMED',
+				'priority'    => 0,
+			),
+			'canceled'    => array(
+				'label'       => __( 'Canceled', 'gatherpress' ),
```
carstingaxion, line 80: Use _x and add a proper translator comment before

includes/core/classes/event/class-status.php
```diff
@@ -0,0 +1,374 @@
...
+			'postponed'   => array(
+				'label'       => __( 'Postponed', 'gatherpress' ),
+				'description' => __( 'Event is delayed to a future unconfirmed date.', 'gatherpress' ),
+				'color'       => '#b06000',
+				'schema'      => 'EventPostponed',
+				'ical'        => 'TENTATIVE',
+				'priority'    => 40,
+			),
+			'rescheduled' => array(
+				'label'       => __( 'Rescheduled', 'gatherpress' ),
```
carstingaxion, line 99: Use _x and add a proper translator comment before

includes/core/classes/event/class-status.php
```diff
@@ -0,0 +1,374 @@
...
+					'Event will not take place. Calendar feeds will mark it as canceled.',
+					'gatherpress'
+				),
+				'color'       => '#c5221f',
+				'schema'      => 'EventCancelled',
+				'ical'        => 'CANCELLED',
+				'priority'    => 50,
+			),
+			'postponed'   => array(
+				'label'       => __( 'Postponed', 'gatherpress' ),
```
carstingaxion, line 91: Use _x and add a proper translator comment before

includes/core/classes/event/class-status.php
```diff
@@ -0,0 +1,374 @@
...
+			'rescheduled' => array(
+				'label'       => __( 'Rescheduled', 'gatherpress' ),
+				'description' => __( 'Event date and time have been changed.', 'gatherpress' ),
+				'color'       => '#1a73e8',
+				'schema'      => 'EventRescheduled',
+				'ical'        => 'TENTATIVE',
+				'priority'    => 30,
+			),
+			'moved'       => array(
+				'label'       => __( 'Moved', 'gatherpress' ),
```
carstingaxion, line 107: Use _x and add a proper translator comment before

includes/core/classes/event/class-status.php
```diff
@@ -0,0 +1,374 @@
...
+				),
+				'color'       => '#7627bb',
+				'schema'      => 'EventScheduled',
+				'ical'        => 'CONFIRMED',
+				'priority'    => 20,
+				// An event can only have moved if it can say where it is.
+				'supports'    => array( 'gatherpress-venue', 'gatherpress-online-event' ),
+			),
+			'tentative'   => array(
+				'label'       => __( 'Tentative', 'gatherpress' ),
```
carstingaxion, line 120: Use _x and add a proper translator comment before

src/utility.scss
```diff
@@ -35,6 +35,27 @@
...
+		box-shadow: 0 0 0 5px transparent;
+	}
+
+	100% {
+		transform: scale(0.95);
+		box-shadow: 0 0 0 0 transparent;
+	}
+}
+
 // Respect a reduced-motion preference without losing the loading state.
```
carstingaxion, line 59: Where is this used, I couldn't find any spot?!

carstingaxion (changes requested):

carstingaxion (changes requested):
Even when the isActive condition is met, the block DOES NOT identify as "Event Status" 

```html
<!-- wp:post-terms {"term":"_gatherpress_event_status","className":"gatherpress-event-status is-style-post-terms-1"} /-->
```

Image

carstingaxion:
If there will be no admin column showing the current Event Status, maybe GatherPress could leverage the [display_post_states](https://developer.wordpress.org/reference/hooks/display_post_states/) and highlight all non-default statuses in that way? (maybe using the colors from the registry, to match frontend and help editing users to understand easily)

carstingaxion:
In addition, IMHO this needs a documentation file explaining how to adjust the colors to match a custom theme.
And it needs a simple block documentation because wp.org will nobody tell about this block(-variation).

carstingaxion:
@wppoland @mauteri From my perpspective, using the block is especially interesting to highlight any changes, less about saying "97% of our events are scheduled". To help with focusing, I added a toggle to my experiments under [carstingaxion/gatherpress-event-status](https://github.com/carstingaxion/gatherpress-event-status) that allows the block to be hidden with the default status (scheduled) set. Maybe its worth implementing such a thing here, too?!

wppoland:
Thanks @carstingaxion! That is a good point regarding focus: for most readers, non-scheduled statuses (cancelled, postponed, rescheduled) are what need immediate visual emphasis.

Since this PR uses an Event Status variation of core's `core/post-terms` block (rather than a standalone custom block), it doesn't have custom block attributes by default. If we decide to support hiding default statuses, we could either:
1. Filter the block rendering (`render_block_core/post-terms`) when the class is `gatherpress-event-status` and the term matches the default status, or
2. Extend the variation via `blocks.registerBlockType` filters to add a toggle attribute.

Let's see what @mauteri thinks about the preferred direction here.

wppoland:
Thank you for the thorough review and testing in the Playground, @carstingaxion! All points have been addressed and resolved in commit 18093b1e:

### 1. Editor vs Frontend Prefix/Suffix & Theme Block Styles
- **Prefix/Suffix distortion fixed:** In the editor, `core/post-terms` renders prefix and suffix using `<span class="wp-block-post-terms__prefix">`, while terms render as `<a>` tags. Badge styles are now scoped specifically to `.gatherpress-event-status__term` (frontend) and `> a[rel="tag"]`, ensuring prefix/suffix spans in Gutenberg are untouched.
- **Theme block styles respected:** Scoped badge styling with `&:not([class*="is-style-"], .is-style-default), &.is-style-gatherpress-badge`. When a theme provides block styles (e.g. Twenty Twenty-Five styles), they are now respected without being overridden.
- **Selectable Block Style registered:** Registered the `gatherpress-badge` block style via `register_block_style()` on `core/post-terms` with label `"Badge"`.
- **Variation deduplication:** Added `unregisterBlockVariation( CORE_BLOCK, TAXONOMY_STATUS )` to prevent core's auto-generated variation from conflicting with GatherPress's customized variation, and added explicit `labels` to taxonomy registration.

### 2. Status Colors
- Fixed the issue where `--gatherpress-status-color: currentcolor;` on `.wp-block-post-terms.gatherpress-event-status` was overriding the status color custom property from parent containers (`.gatherpress-event-status--is-*`). Removed that override and used `var(--gatherpress-status-color, currentcolor)`.
- Added inline style rules in `Setup::register_status_style()` targeting both `.gatherpress-event-status--is-{slug}` and `.wp-block-post-terms.gatherpress-event-status` term links, ensuring colors render reliably in both editor and frontend.
- Validated status colors using WordPress's `sanitize_hex_color()`.

### 3. Restored `moved-online` & Schema.org Mapping
- Updated `moved` back to `moved-online` with Schema.org `EventMovedOnline`, RFC 5545 `CONFIRMED`, priority 20, and `supports: [ 'gatherpress-online-event' ]`.
- Updated all associated PHP and JS tests to assert `moved-online` and `EventMovedOnline`.

### 4. i18n & Code Quality
- **i18n:** Added `/* translators: Event operational status label. */` comments and `_x( ..., 'event status', 'gatherpress' )` across all status labels.
- **Type safety:** Removed unnecessary `(array)` cast on `apply_filters( 'gatherpress_event_statuses', ... )`; PHPStan infers the type cleanly with 0 errors.
- **Cleaned CSS:** Removed the unused `@keyframes gatherpress-pulse-ring` animation from `utility.scss`.
- **Validation:** 
  - `npm run test:unit:js` (all 66 suites, 1,143 tests pass)
  - `npm run lint:js` & `npm run lint:css` (0 errors)
  - `npm run build` (compiled cleanly)
  - `vendor/bin/phpcs` (0 errors, 0 warnings)
  - `composer test:phpstan` (0 errors across 144 files)

Ready for your re-review!

mauteri:
> Thanks @carstingaxion! That is a good point regarding focus: for most readers, non-scheduled statuses (cancelled, postponed, rescheduled) are what need immediate visual emphasis.
> 
> Since this PR uses an Event Status variation of core's `core/post-terms` block (rather than a standalone custom block), it doesn't have custom block attributes by default. If we decide to support hiding default statuses, we could either:
> 
> 1. Filter the block rendering (`render_block_core/post-terms`) when the class is `gatherpress-event-status` and the term matches the default status, or
> 2. Extend the variation via `blocks.registerBlockType` filters to add a toggle attribute.
> 
> Let's see what @mauteri thinks about the preferred direction here.

I like idea 2 of extending the variation.

wppoland:
Rebased and merged latest `develop` to resolve the conflict in `class-test-rest-api.php`, and implemented Idea 2:

### Event Status Variation Extension
- **`hideWhenScheduled` toggle attribute:** Extended the `core/post-terms` block via `blocks.registerBlockType` filter with a `hideWhenScheduled` boolean attribute (default `false`).
- **InspectorControls panel:** Added a "Hide when scheduled" toggle control to the block settings sidebar, allowing organizers to display the status badge only for non-scheduled events (canceled, postponed, rescheduled, moved online, tentative).
- **Server-side rendering:** Hooked `render_block_core/post-terms` to suppress block output when `hideWhenScheduled` is enabled and the event is in the default `scheduled` status.
- **Active variation matching:** Standardized `isActive` to `[ 'term' ]` so the block identifies cleanly as Event Status in breadcrumbs and toolbar.

PR #2258, opened 2026-09-05
Announce links that open in a new tab

### LINKED ISSUE #2141: Links that open in a new tab give no warning

mauteri:
Closes #2141. An alternative to #2201.

Not smaller than #2201: about 250 lines of implementation against its 130. The difference is where the code lives. #2201 puts the markup in each block that needs it, so three templates changed and a fourth block got its own filter. This puts it in one place and no block participates, including blocks we add later.

## Proposed changes:

Sighted users get a cue from the new tab itself. Screen-reader users get nothing unless the link says so.

* **One filter, every block.** `Blocks\General_Block::announce_new_tab_links()` runs on `render_block` for anything under the `gatherpress/` prefix and adds a visually hidden note inside links that target `_blank`. No block templates change.
* **Links created after render.** A PHP filter cannot see a link that a block's own JS builds, which is what `online-event-link` does when it swaps a span for an anchor. `src/helpers/new-tab-notice.js` watches GatherPress blocks and keeps the notice in step with what its links actually do: added where a new-tab link lacks one, withdrawn where one has outlived its link. It is registered once beside the utility stylesheet and enqueued with it, so it loads only when a GatherPress block is on the page, and translates its own string through `wp_set_script_translations`.

  The withdrawing half was missing until @wppoland caught it in #2274. `online-event-link` hides its link by swapping in a span and copying the content across verbatim, notice included, so plain text was left announcing that it opens a new tab. The same held for a link retargeted away from `_blank`. Fixed in 23f69add9.
* **Links in the block sidebar.** Neither the filter nor the observer sees the editor's inspector, which React builds outside any block. The two documentation links there, date formatting on `event-date` and autocomplete on `form-field`, now use core's `ExternalLink`, which carries the announcement, the external `rel` values and the arrow the rest of the editor shows. Pulled in from @wppoland's #2274 in 819d91a69.
* **Styling.** Every note we emit carries `gatherpress--screen-reader-text`, one utility class in the shared stylesheet that mirrors core's `.screen-reader-text`, so it hides wherever it lands: inside a block, or a tooltip in a core paragraph. The tooltip's own copy of those declarations is deleted. Core ships `.screen-reader-text` in the block-library stylesheet and most themes carry it, but neither is guaranteed, and without it the note renders as visible text next to every link.

The tag processor chooses the anchors and finds their closing tags, so attribute quoting, tag case, whitespace, comments and attribute text are its problem rather than a string search's. It cannot insert markup or say where a token sits, so a small `Tag_Processor` subclass reads each closer's offset through a bookmark, and the note is spliced in at those offsets from the end.

### Other information:

## Testing instructions:

Checked on a real page, not only in tests:

* **Server-rendered.** On an event with an Add to Calendar block, Google and Yahoo carry the note. iCal does not, correctly, because it is a download rather than a new tab.
* **Created after render.** A link added to a GatherPress block after load is announced, as is a link whose target changes to `_blank` afterwards.
* **Reverted after render.** Swap the online event link back to plain text (change your RSVP so the link hides), or retarget a link to `_self`: the notice goes with it. Static text must not claim to open a tab.
* **In the editor.** Select an Event Date block and open its settings: the formatting documentation link shows core's external-link arrow and is read as opening a new tab. Same for the Learn more link under a Form Field's autocomplete setting.
* **Not doubled** on links PHP already handled, and hidden at 1px rather than showing as visible text.

To try it yourself: open an event with an Add to Calendar block, inspect the Google link, and confirm the hidden span sits inside the anchor after the label. Then switch your screen reader on and tab to it.

Props @faisalahammad and @wppoland, whose #2201 and #2274 took earlier runs at #2141.

### Changelog entry

Added at `.github/changelog/2141-announce-new-tab-links`.

## Summary by CodeRabbit

- **New Features**
  - Added screen-reader notices to links that open in new tabs, including dynamically added or updated links.
  - Improved tooltip accessibility and isolation when multiple tooltips or screen-reader elements are present.
  - Documentation links in editor settings now use the standard external-link presentation.
  - Online event links now work with any content type supporting event functionality.

- **Bug Fixes**
  - Prevented duplicate or outdated new-tab notices.
  - Improved handling of varied link targets and dynamic content.


includes/core/classes/blocks/class-setup.php
```diff
@@ -349,4 +367,93 @@ public function enable_context_for_core_query_block( array $args, string $block_
...
+	 * @param array<string,mixed> $block         Parsed block.
+	 *
+	 * @return string Markup with a notice inside each new-tab link.
+	 */
+	public function announce_new_tab_links( string $block_content, array $block ): string {
+		if ( ! str_starts_with( (string) ( $block['blockName'] ?? '' ), 'gatherpress/' ) ) {
+			return $block_content;
+		}
+
+		if ( ! str_contains( $block_content, '_blank' ) ) {
```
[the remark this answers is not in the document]
mauteri: Good catch, fixed in 7edf0a30.

Target keywords beginning with an underscore are ASCII case-insensitive per the HTML spec, so `target="_BLANK"` does open a new tab and would have been left silent. Both the `stripos()` prefilter and the comparison now normalize, and the data provider covers `_BLANK` and `_Blank`.

The same bug was in the front-end script, which you didn't flag: its selector was `a[target="_blank"]` and CSS attribute values are case-sensitive, so links created after render had the identical gap. That now matches on `a[target]` and compares the lowercased value, with a test for it.

includes/core/classes/blocks/class-setup.php
```diff
@@ -349,4 +367,93 @@ public function enable_context_for_core_query_block( array $args, string $block_
...
+		$notice = sprintf(
+			'<span class="screen-reader-text %1$s">%2$s</span>',
+			esc_attr( self::NEW_TAB_CLASS ),
+			esc_html__( '(opens in a new tab)', 'gatherpress' )
+		);
+
+		$output = '';
+		$offset = 0;
+
+		$marked = strpos( $html, $marker );
```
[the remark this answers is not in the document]
mauteri: Valid, fixed in 7edf0a30, though I went a different way than a unique marker value.

The first pass now visits every tag rather than only anchors: qualifying links get the marker, and anything else carrying that attribute has it removed. So a marker already present in authored content is cleared before the splice runs, and the splice can only land on links this pass marked itself. That keeps the marker a fixed, greppable string rather than a per-request random one.

Regression test added with a same-tab link carrying `data-gatherpress-new-tab="1"` alongside a real new-tab link: only the real one is announced, and no marker survives in the output.

src/formats/tooltip/view.js
```diff
@@ -60,12 +67,15 @@ export function initTooltip( tooltip ) {
 		'function' === typeof tooltip.appendChild &&
 		'undefined' !== typeof document
 	) {
+		// Found by its own class rather than by position, so another
+		// feature's screen-reader text inside the same element is never
+		// mistaken for this one.
 		const existingSrText = tooltip.querySelector(
-			':scope > .screen-reader-text'
+			`.${ NOTICE_CLASS }`
 		);
```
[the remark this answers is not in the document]
mauteri: Right, and this was a regression I introduced on this branch. The tooltip used `:scope > .screen-reader-text` before, and when I gave it a marker class I switched to a descendant query and lost the direct-child constraint.

Fixed in 7edf0a30: it is `:scope > .${ NOTICE_CLASS }` now, so it is narrowed by both the class and the position. Nested-tooltip test added, where an outer tooltip containing an inner one still speaks its own text rather than adopting the child's.

mauteri:
@coderabbitai all three are addressed in 7edf0a30. Please take another look and confirm, particularly the marker approach, since I cleared stray markers in the first pass instead of generating a request-unique value.

src/helpers/new-tab-notice.js
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Announce every new-tab link inside an element, and the element itself.
+ *
+ * @param {Element} root The element to search.
+ *
+ * @return {void}
+ */
+const announceWithin = ( root ) => {
+	root.querySelectorAll?.( 'a[target="_blank"]' ).forEach( announce );
```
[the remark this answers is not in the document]
mauteri: Already covered by 7edf0a30. I spotted the same gap on the front end while fixing the PHP side and did both together.

The selector was `a[target="_blank"]`, and CSS attribute values are case-sensitive, so an uppercase target was missed there too. It now matches on `a[target]` and compares through a small `opensNewTab()` helper that lowercases the value, which the root-anchor branch uses as well.

Jest covers both paths you asked for: `_BLANK` present at load, and `_Blank` set on a link appended after render. There is also a case pinning that `_self` and a named target stay untouched.

src/utility.scss
```diff
@@ -1,5 +1,24 @@
 // Import tooltip format styles for frontend display.
 @use "formats/tooltip/style";
+// Hide screen-reader text anywhere inside a GatherPress block, so no block
+// has to style it itself. Core ships `.screen-reader-text` in the
+// block-library stylesheet and most themes carry it, but neither is
+// guaranteed, and without it these notes render as visible text.
+// Declarations mirror core's, with `overflow-wrap` standing in for its legacy
+// `word-wrap` alias. No `:focus` counterpart: nothing hidden this way is
+// focusable, and adding one would fight a theme's own skip-link styling.
+[class*="wp-block-gatherpress-"] .screen-reader-text {
```
mattcowan: The tooltip format is registered for every block, and `maybe_enqueue_tooltip_assets` loads this stylesheet for any block that carries it. The rule this replaces matched `.gatherpress-tooltip > .screen-reader-text` anywhere. This one does not reach a tooltip in a core paragraph, so without core's rule the note renders as visible text. Both notices have their own class now, so `.gatherpress-tooltip-notice` can join the selector.

```suggestion
[class*="wp-block-gatherpress-"] .screen-reader-text,
.gatherpress-tooltip-notice {
```

src/blocks/online-event-link/render.php
```diff
@@ -27,10 +27,23 @@
...
+// that uses it and stays updatable.
 $gatherpress_link_text = $attributes['linkText'] ?? '';
+
 if ( empty( $gatherpress_link_text ) ) {
-	$gatherpress_link_text = __( 'Online event', 'gatherpress' );
+	$gatherpress_link_text = esc_html__( 'Online event', 'gatherpress' );
+
+	// Only events hold the link back until someone is attending, so the
+	// caveat belongs to them and not to anything else using this block.
+	if ( Event::POST_TYPE === $gatherpress_current_post_type ) {
```
mattcowan: Is this intended? This checks Event::POST_TYPE, and edit.js:110 checks isEventPostType(), which is gatherpress-event-date support. A post type that declares the support gets the caveat in the editor and not on the front end. AGENTS.md asks for post_type_supports() here and usePostTypeSupports() in components.

src/blocks/online-event-link/edit.js
```diff
@@ -105,8 +105,18 @@ const Edit = ( { context, attributes, setAttributes } ) => {
...
 
-	// Use default text if linkText is empty.
-	const displayText = linkText || __( 'Online event', 'gatherpress' );
+	// Mirrors the default render.php builds when the attribute is empty, so the
+	// editor shows what the front end will render without saving it.
+	const defaultText = isEventPostType( postType )
+		? sprintf(
+			/* translators: %1$s: tooltip text, %2$s: label text */
+			'<span class="gatherpress-tooltip" data-gatherpress-tooltip="%1$s">%2$s</span>',
+			__( 'link available for attendees only', 'gatherpress' ),
```
mattcowan: render.php escapes this with esc_attr__(). Here the translation lands raw in an attribute, so a translation with a double quote breaks the preview markup. escapeAttribute() from @wordpress/escape-html matches the PHP side.

.github/changelog/2141-online-event-link-default
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: changed
+
+The online event link's tooltip, and the screen reader text it produces, is built when the block renders instead of being written into the post. Its wording can be corrected without editing anyone's content. The link text itself is editable exactly as before.
```
mattcowan: Posts saved before this change keep the seeded linkText, so the old wording stays until someone edits the block. On my site an event created on develop still shows "Link available for attendees only." with this branch. Does the second sentence need to be dropped, or limit it to new content?

includes/core/classes/class-assets.php
```diff
@@ -278,6 +279,39 @@ public function register_block_assets(): void {
...
+	 * @return string The block content, unchanged.
+	 */
+	public function maybe_enqueue_new_tab_notice( string $block_content, array $block ): string {
+		if ( ! str_starts_with( (string) ( $block['blockName'] ?? '' ), 'gatherpress/' ) ) {
+			return $block_content;
+		}
+
+		$asset = $this->get_asset_data( 'new_tab_notice' );
+
+		wp_enqueue_script(
```
mattcowan: This runs the full wp_enqueue_script() and wp_set_script_translations() path once per GatherPress block on the page. Register the script once in register_block_assets() beside the utility style, and enqueue it by handle from the matched branch of maybe_enqueue_styles(). That branch also honors gatherpress_asset_utility_style_block_prefixes, so companion blocks that share the stylesheet get the script too.

includes/core/classes/blocks/class-setup.php
```diff
@@ -349,4 +367,109 @@ public function enable_context_for_core_query_block( array $args, string $block_
...
+	public function announce_new_tab_links( string $block_content, array $block ): string {
+		if ( ! str_starts_with( (string) ( $block['blockName'] ?? '' ), 'gatherpress/' ) ) {
+			return $block_content;
+		}
+
+		// Target keywords beginning with an underscore are case-insensitive,
+		// so `_BLANK` opens a new tab just as `_blank` does.
+		if ( false === stripos( $block_content, '_blank' ) ) {
+			return $block_content;
+		}
```
mattcowan: Four `return` statements. AGENTS.md caps a method at three (php:S1142) and asks for guard chains to merge. The two leading guards can be one `if`.

```suggestion
		// Target keywords beginning with an underscore are case-insensitive,
		// so `_BLANK` opens a new tab just as `_blank` does.
		if (
			! str_starts_with( (string) ( $block['blockName'] ?? '' ), 'gatherpress/' )
			|| false === stripos( $block_content, '_blank' )
		) {
			return $block_content;
		}
```

src/helpers/new-tab-notice.js
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * PHP adds the notice when a block renders. This covers links that appear or
+ * change afterwards, so a block that swaps a span for a link at runtime gets
+ * the same treatment without knowing this exists.
+ *
+ * The notice follows the target rather than being added once. A block that
+ * hides a link by swapping in a span copies the content across verbatim,
+ * notice included, and would otherwise leave static text announcing that it
+ * opens a new tab. The same holds for a link retargeted away from `_blank`.
+ */
```
mattcowan, line 27: New symbols need @since 0.36.0 in the docblock, per the AGENTS.md docblock rule. Same for the other three functions in this file and NOTICE_CLASS in tooltip/view.js.

mattcowan (commented):
The filter and the observer do what the description says, and I confirmed the calendar links, the hidden rendering, and the editor preview on a real site. Two regressions need a fix before merge: the RSVP swap back to a span keeps the notice, and the tooltip's hidden text outside a GatherPress block lost its rule. The other comments are alignment with AGENTS.md conventions and questions about scope.

src/utility.scss
mauteri: Right, a tooltip in a paragraph never sat inside a GatherPress block. Rather than adding selectors, every note we emit now carries one utility class, `gatherpress--screen-reader-text`, and that is the only rule. Done in 84069b68d.

src/blocks/online-event-link/render.php
mauteri: Not intended. Both checks in render.php go through `post_type_supports()` with `Event::SUPPORT` now, and edit.js uses `usePostTypeSupports()`, so the editor and the front end agree. Done in 84069b68d.

src/blocks/online-event-link/edit.js
mauteri: Done in 84069b68d with `escapeAttribute()` from `@wordpress/escape-html`, now a listed dependency.

includes/core/classes/class-assets.php
mauteri: Done in 84069b68d. Registered once in `register_block_assets()` with its translations, and enqueued by handle from the matched branch of `maybe_enqueue_styles()`, so the prefix filter covers it as well. The separate filter and its tests are gone; the `maybe_enqueue_styles` tests now assert the script alongside the stylesheet.

includes/core/classes/blocks/class-setup.php
mauteri: Merged in 84069b68d, three returns.

src/helpers/new-tab-notice.js
mauteri: Added in 84069b68d: the module docblock, the constants and all five functions in the helper, and `NOTICE_CLASS` in tooltip/view.js.

mauteri:
Thanks @mattcowan. The swap regression was fixed in 23f69add9, pushed this morning before your review landed; if your checkout predates it, pull and the notice leaves with the span. The tooltip rule and the rest are in 84069b68d.

src/blocks/online-event-link/render.php
mauteri: Reworked in 55937f660: the render-time default is gone and the seeded link text stays in the template as editable content, so the caveat check went with it. The link lookup still goes through `post_type_supports()` with `Event::SUPPORT`.

src/blocks/online-event-link/edit.js
mauteri: Moot in 55937f660: the editor no longer builds the default, so no translated string lands in an attribute from JS. The dependency is dropped again.

.github/changelog/2141-online-event-link-default
mauteri: Dropped in 55937f660. The default is not built at render after all; the seeded text stays in the template and only the screen-reader notices are added on render. That changelog entry is deleted.

mauteri:
@coderabbitai review

includes/core/classes/blocks/class-general-block.php
```diff
@@ -326,4 +351,110 @@ public function process_anonymous_field( string $block_content, array $block ):
...
+			esc_html__( '(opens in a new tab)', 'gatherpress' )
+		);
+
+		$output = '';
+		$offset = 0;
+
+		$marked = strpos( $html, $marker );
+
+		while ( false !== $marked ) {
+			$close = strpos( $html, '</a>', $marked );
```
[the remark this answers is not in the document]
mauteri: Fixed in e9e35e39d. The processor now finds the closers itself: the announce pass visits tag closers, records each qualifying anchor's closing-tag offset, and splices from the end. The offset comes from a small `Tag_Processor` subclass that reads it through a bookmark and releases it at once. The marker attribute and the second string scan are gone. Regression cases added for `</A>`, `</a >`, a `</a>` inside a comment and one inside an attribute, with a placement test asserting the notice lands before the real closer in each.

test/unit/js/src/blocks/event-date/edit.test.js
```diff
@@ -29,6 +29,7 @@ jest.mock( '@wordpress/block-editor', () => ( {
 
 jest.mock( '@wordpress/components', () => ( {
 	__experimentalVStack: ( { children } ) => <div>{ children }</div>,
+	ExternalLink: ( { href, children } ) => <a href={ href }>{ children }</a>,
```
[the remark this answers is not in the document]
mauteri: Done in e9e35e39d. The `ExternalLink` stand-in now renders what core's does, anchor with `target`, `rel` and the labelled arrow, and the event-date test asserts the documentation link opens a new tab and carries the announcement in its accessible name. `form-field` gets its own Jest test asserting the same for its Learn more link, and that a hidden field has none.

Issue #1169, opened 2025-08-13
Allow to query by date range


carstingaxion:
### Is your enhancement related to a problem? Please describe.

To query events by a particular date range is a need:

1. For editors using the _event query block_ who want to show **“Events next month”**
2. For editors using the _event query block_ together with the `core/query-total` block for statistical queries, like **“We had 123 great Events in 2024”** (https://github.com/GatherPress/gatherpress-wp-org-integration/issues/13)
3. ~~In the admin view for events, a new select field could be added, allowing to “Filter by Event month or year”. Similar to the core filter dropdown, to filter by publication date, which could be removed in favour of the before mentioned. (#694)~~ Handled in #1702.

### Designs

Some adjustments in the `event-query `classes will be needed in general, while all three use cases will also need additional code to work. Issues should be created for each.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

carstingaxion:
I wanted to note, that I’m using a filter within the [gatherpress/calendar](https://github.com/carstingaxion/gatherpress-calendar) and [gatherpress/statistics](https://github.com/carstingaxion/gatherpress-statistics) blocks, that makes date base queries work.

Maybe this could be interesting to be implemented in GatherPress core, too.
While I’m using custom UI on the blocks, GatherPress could just re-enable the datequery filters on its variations the core/query block. 



Here is, what I’m doing:

```php
/**
 * Filter SQL WHERE clause to use GatherPress event dates instead of post dates.
 *
 * The exact same filter exists in GatherPress/Statistics to enable date queries.
 *
 * @since 0.1.0
 *
 * @global wpdb $wpdb WordPress database abstraction object.
 *
 * @param string   $where The WHERE clause of the query.
 * @param WP_Query $query The WP_Query instance.
 *
 * @return string Modified WHERE clause.
 */
function posts_where( string $where, WP_Query $query ): string {
	global $wpdb;

	if ( ! ( $wpdb instanceof wpdb ) ) {
		return $where;
	}

	if ( empty( $query->query_vars['date_query'] ) || ! is_array( $query->query_vars['date_query'] ) ) {
		return $where;
	}

	if ( 'gatherpress_event' !== $query->get( 'post_type' ) && ! in_array( 'gatherpress_event', (array) $query->get( 'post_type' ), true ) ) {
		return $where;
	}

	$date_filter_raw = $query->query_vars['date_query'][0] ?? array();
[...]

Issue #369, opened 2023-09-04
BLOCK Calendar


mauteri:
This block is a calendar view which has events appear in a calendar format. 

Two possible view options: Monthly and Weekly.

carstingaxion:
![](https://make.wordpress.org/training/files/2024/07/learn-online-workshop.png)

> On the development front, [a new block child theme](https://github.com/WordPress/Learn/tree/trunk/wp-content/themes/pub/wporg-learn-2024) was created. This builds on the parent theme used for all the previously redesigned sites and integrates blocks from the Sensei LMS (Learning Management System) plugin, which powers the courses and lessons on Learn WordPress, much like it did prior to this redesign. 

Maybe a piece of code to adopt or learn from, for the Calendar block, GatherPress needs.

MervinHernandez:
October 25: Discussed = it would be best if core has a calendar block we can leverage, as oppose to composing a full calendar block from scratch.

carstingaxion:
I created a plugin for myself, that provides a Calendar block, and for sure it works with GatherPress.
https://github.com/carstingaxion/gatherpress-calendar

https://github.com/user-attachments/assets/7089ef0c-e2aa-417a-a3fb-a5861315869b

mauteri:
Looks good! Will check it out. 😊

mauteri:
@carstingaxion this is looking like a pretty good solution. There are some bugs that need to be worked out with some blocks within the modal. Adding link to title or adding an RSVP block breaks the layout; looks like due to nested anchors. We should try to leverage the modal manager/modal block here too so we don't have 2 modal implementations. Hoping we can simplify the inclusion of the block somehow. I originally looked for it in blocks before I realized in the readme that it actually needs to be a child of the query loop block. I understand that limitation, but maybe we can add a variation of the Query loop block for an "Event Calendar" to make it easier for people to find. 

These are my initial thoughts, but overall, nice work! I'm fine with expanding on what you started here in this repo you setup until it is ready to integrate into the project, or if you want to put in a PR to GatherPress that's fine too. I'll leave that to you. Thanks!

carstingaxion:
> I do want to clarify, though, that I don’t think it’s accurate to say there was little or no feedback. There were periods where my attention was pulled to other parts of the project, and I may not always have responded as quickly or as deeply as you hoped. That’s on capacity, not lack of interest or appreciation for the work you put in.

That is ok.
[...]

PR #2266, opened 2026-09-06
Resolve date queries against event dates

mauteri:
Part of #1169, which is three separate things. This is the mechanism underneath all of them, and it closes the second bullet on its own.

## The problem

WordPress resolves `date_query` against `post_date`. For an event that is the day someone wrote the post, not the day the event happens, so "events in June" answers with whatever was published in June.

## The change

For a query made up entirely of post types supporting `gatherpress-event-date`, `date_query` is lifted out of the query vars during `pre_get_posts` and resolved into a window compared against the events table instead.

The interception happens before core builds any SQL for it: `pre_get_posts` fires at `WP_Query` line 1916, the `date_query` clause is built at 2144. That is worth calling out because the existing workaround in [gatherpress-calendar](https://github.com/carstingaxion/gatherpress-calendar) and gatherpress-statistics has to regex the generated `YEAR()`/`MONTH()` clause back out of the `WHERE` string. Nothing needs stripping here, because nothing is ever built.

**Overlap, not start.** An event belongs to the window when its start falls before the window closes and its end falls after the window opens, matching the admin filter that shipped in #2263. One running May 30 to June 2 answers to both months; one running September to November answers to October, though it neither starts nor ends there.

**Boundaries are reckoned in the site's timezone**, the way WordPress reckons `post_date`, then converted once to UTC and compared against `datetime_start_gmt` / `datetime_end_gmt`. Plain range comparisons, so `KEY datetime_start_gmt` stays usable. This is the only reading that holds up across events in different timezones, which is the wrinkle GatherPress has and core does not: core's `post_date` is one timezone for the whole site.

**Two escape hatches**, both shaped like core's own `column` key:

- `'column' => 'datetime_start'` buckets by each event's own clock instead. Not index-backed, and not comparable between timezones, but it is what a list table wants when the column beside the filter renders in that clock.
- `'column' => 'post_date'` (or any other core column) leaves the clause entirely to WordPress. That is how an event query keeps filtering by publish date, which the admin list relies on.

`Date_Query::resolve()` reads the range-shaped arguments only: `after`, `before`, `year`, `monthnum`/`month`, `day` and `inclusive`. `week`, `dayofweek`, `hour` and friends describe a repeating slice of the calendar rather than one stretch of it and have no single answer across mixed timezones, so they are left alone rather than answered wrongly.

## What this does and does not close

- **Bullet 2, the `core/query-total` statistics case, is closed by this.**
- **Bullet 1, "Events next month" in a query loop, is closed for Advanced Query Loop users.** `aql_query_vars()` already passes AQL's built args through untouched, and AQL ships its own date controls, so its date range starts resolving against event dates as soon as this lands. The core Query Loop still has no date control; that UI is a separate PR.
- **Bullet 3, the admin filter, shipped in #2263** and is deliberately unaffected: it filters by publish date through core's `m` parameter, and its event-date filter has its own parameter.

## Two things worth a look

- **A query mixing event and non-event post types is left to core.** Non-event rows have no events-table row, so hijacking would silently drop them from results.
- **`adjust_event_sql()` used to append the events-table join unconditionally**, which was fine when only the upcoming and past handlers called it. The date window handler needs the same join and may run without them (an "all" query joins nothing today), so both now go through an idempotent `ensure_events_join()`.

## Testing

1. Add events across several months, including one that runs from one month into the next, and one in a distant timezone near a month boundary.
2. In a query loop using **Advanced Query Loop** with the event post type, set a date range and confirm it filters on event dates rather than publish dates.
3. Confirm the range picks up an event that merely overlaps it.
4. **Events > All Events**: confirm both admin dropdowns still behave as they did, since the admin list is deliberately untouched.

## Verification

Full suite green at 2020 tests. `class-date-query.php` 78/78 and `class-query.php` 330/330 lines covered. Mutating the interception away, the column allow-list away, and the join's idempotency away each fails the tests that should catch it.

## Summary by CodeRabbit

* **New Features**
  * Event searches using date filters now match events that overlap the requested date range, including year, month, day, and custom date boundaries.
  * Date filtering respects the site’s configured timezone and supports inclusive date ranges.

* **Bug Fixes**
  * Corrected event date filtering so it evaluates event dates rather than standard post publication dates.

* **Tests**
  * Added coverage for date-range filtering, timezone handling, boundary conditions, and mixed query types.

### DIFF
--- a/includes/core/classes/event/class-date-query.php
+++ b/includes/core/classes/event/class-date-query.php
@@ -66,6 +66,29 @@ final class Date_Query {
 	const EVENT_COLUMNS = array( self::DEFAULT_COLUMN, 'datetime_start' );
 
 	/**
+	 * Arguments a clause may carry and still be resolved.
+	 *
+	 * A clause carrying anything else is refused outright rather than
+	 * resolved in part. Honoring `year` while quietly dropping `week` would
+	 * hand back a window a whole year wide, and the caller lifts the original
+	 * `date_query` out of the query once a window comes back, so there is
+	 * nothing left to apply the dropped argument.
+	 *
+	 * @since 0.36.0
+	 * @var string[]
+	 */
+	const SUPPORTED_ARGS = array(
+		'after',
+		'before',
+		'column',
+		'day',
+		'inclusive',
+		'month',
+		'monthnum',
+		'year',
+	);
+
+	/**
 	 * Resolve a date query into the window an event has to touch.
 	 *
 	 * Reads the range-shaped arguments only: `after`, `before`, `year`,
@@ -75,8 +98,14 @@ final class Date_Query {
 	 * single answer across events in different timezones, so it is left alone
 	 * rather than answered wrongly.
 	 *
-	 * Only the first clause is read. Nested clauses and relations are a
-	 * generality nothing asks for yet.
+	 * A date query is resolved in full or not at all. One clause carrying only
+	 * arguments from `SUPPORTED_ARGS` resolves; anything else, including a
+	 * relation, a second clause, or one unreadable argument alongside readable
+	 * ones, resolves to nothing and is left for WordPress.
+	 *
+	 * Arguments narrow each other the way they do in a WordPress clause, so a
+	 * calendar span alongside an `after` or `before` yields their overlap
+	 * rather than whichever was read last.
 	 *
 	 * Boundaries come back in the clock of the column they will be compared
 	 * against: UTC for the GMT pair, the site's own time for the local pair. A
@@ -92,10 +121,15 @@ final class Date_Query {
 	 *         when nothing could be read.
 	 */
 	public static function resolve( array $date_query ): ?array {
-		$clause = self::first_clause( $date_query );
+		$clause = self::single_clause( $date_query );
+
+		if ( null === $clause ) {
+			return null;
+		}
+
 		$column = (string) ( $clause['column'] ?? self::DEFAULT_COLUMN );
 
-		if ( empty( $clause ) || ! in_array( $column, self::EVENT_COLUMNS, true ) ) {
+		if ( ! in_array( $column, self::EVENT_COLUMNS, true ) ) {
 			return null;
 		}
 
@@ -115,7 +149,8 @@ final class Date_Query {
 			if ( null !== $after ) {
 				// WordPress treats `after` as exclusive unless told otherwise,
 				// and the window is built to the second.
-				$start = $inclusive ? $after : $after->modify( '+1 second' );
+				$after = $inclusive ? $after : $after->modify( '+1 second' );
+				$start = ( null === $start || $after > $start ) ? $after : $start;
 			}
 		}
 
@@ -123,7 +158,8 @@ final class Date_Query {
 			$before = self::to_site_datetime( $clause['before'], true );
 
 			if ( null !== $before ) {
-				$end = $inclusive ? $before : $before->modify( '-1 second' );
+				$before = $inclusive ? $before : $before->modify( '-1 second' );
+				$end    = ( null === $end || $before < $end ) ? $before : $end;
 			}
 		}
 
@@ -141,22 +177,39 @@ final class Date_Query {
 	}
 
 	/**
-	 * Pull the clause to read out of a date query.
+	 * Pull the one clause a date query can be resolved from.
 	 *
-	 * A date query may carry its arguments at the top level or wrap them in a
-	 * list of clauses. Both shapes are common in the wild, so both are read.
+	 * A date query may carry its arguments at the top level or wrap a single
+	 * clause in a list. Both shapes are common in the wild, so both are read.
+	 * Anything else, a relation, a second clause, or an argument outside
+	 * `SUPPORTED_ARGS`, cannot be honored in full and so is not honored at all.
 	 *
 	 * @since 0.36.0
 	 *
 	 * @param array<int|string, mixed> $date_query A WordPress date query.
 	 *
-	 * @return array<string, mixed> The clause to read, or an empty array.
+	 * @return array<string, mixed>|null The clause to read, or null when there is not exactly one to read.
 	 */
-	private static function first_clause( array $date_query ): array {
-		if ( isset( $date_query[0] ) && is_array( $date_query[0] ) ) {
-			return $date_query[0];
+	private static function single_clause( array $date_query ): ?array {
+		$nested = array_filter( $date_query, 'is_int', ARRAY_FILTER_USE_KEY );
+
+		// A wrapped clause is read only when it is the whole date query, which
+		// rules out a relation or a sibling clause sitting beside it.
+		if ( ! empty( $nested ) ) {
+			$date_query = ( 1 === count( $nested ) && count( $nested ) === count( $date_query ) )
+				? (array) reset( $nested )
+				: array();
 		}
 
+		if ( empty( $date_query ) || ! empty( array_diff( array_keys( $date_query ), self::SUPPORTED_ARGS ) ) ) {
+			return null;
+		}
+
+		/**
+		 * The clause, now known to carry supported arguments only.
+		 *
+		 * @var array<string, mixed> $date_query
+		 */
 		return $date_query;
 	}
 
--- a/includes/core/classes/event/class-date-query.php
+++ b/includes/core/classes/event/class-date-query.php
@@ -1,0 +1,294 @@
+<?php
+/**
+ * Resolves a WordPress date query into a window of event time.
+ *
+ * @package GatherPress\Core\Event
+ * @since 0.36.0
+ */
+
+namespace GatherPress\Core\Event;
+
+// Exit if accessed directly.
+defined( 'ABSPATH' ) || exit; // @codeCoverageIgnore
+
+use DateTimeImmutable;
+use DateTimeZone;
+use Exception;
+
+/**
+ * Turns a `date_query` into the window of time an event has to touch.
+ *
+ * WordPress resolves `date_query` against `post_date`, which for an event is
+ * the day someone wrote the post, not the day the event happens. This reads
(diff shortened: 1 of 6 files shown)


includes/core/classes/event/class-date-query.php
```diff
@@ -0,0 +1,294 @@
...
+	 * @since 0.36.0
+	 *
+	 * @param array<int|string, mixed> $date_query A WordPress date query.
+	 *
+	 * @return array{start: ?string, end: ?string, column: string}|null The window and
+	 *         the start column it is for, either bound null when open ended, or null
+	 *         when nothing could be read.
+	 */
+	public static function resolve( array $date_query ): ?array {
+		$clause = self::first_clause( $date_query );
```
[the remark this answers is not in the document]
mauteri: Fixed in 5729c72f3. Confirmed as a real wrong-results bug before fixing, not just a theoretical one.

The reason it bites rather than merely under-filtering is the one you named: `intercept_date_query()` lifts the original `date_query` out of the query vars once a window comes back, so a dropped argument has nothing left to apply it. `array( "year" => 2026, "week" => 24 )` really did return all of 2026.

`first_clause()` is now `single_clause()`, and a date query resolves in full or not at all. A new `SUPPORTED_ARGS` list gates it: `after`, `before`, `column`, `day`, `inclusive`, `month`, `monthnum`, `year`. Any key outside that list refuses the whole clause, a `relation` refuses it, and a wrapped clause is only read when it is the entire date query, which rules out a sibling clause sitting beside it.

Four cases added to the provider, all of which fail against the previous code: `year` + `week`, `year` + `compare`, a relation across two clauses, and two clauses with no relation.

includes/core/classes/event/class-date-query.php
```diff
@@ -0,0 +1,294 @@
...
+			list( $start, $end ) = $calendar;
+		}
+
+		if ( isset( $clause['after'] ) ) {
+			$after = self::to_site_datetime( $clause['after'], false );
+
+			if ( null !== $after ) {
+				// WordPress treats `after` as exclusive unless told otherwise,
+				// and the window is built to the second.
+				$start = $inclusive ? $after : $after->modify( '+1 second' );
```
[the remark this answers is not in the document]
mauteri: Fixed in 5729c72f3. Correct, and it was a last-write-wins bug: the calendar span set both bounds, then `after` and `before` overwrote them outright instead of narrowing.

`array( "year" => 2026, "after" => "2025-12-01" )` opened the window in December 2025, which is wider than either argument asked for. Arguments in a WordPress clause are ANDed, so the resolved window is now their overlap: the later of the two starts, the earlier of the two ends.

Three cases added, covering a boundary outside the span on each side and one inside it:

- `year` 2026 + `after` 2025-12-01, which must not widen the start
- `year` 2026 + `after` 2026-06-15, which must narrow it
- `year` 2026 + `before` 2027-06-30, which must not widen the end

The first and third fail against the previous code. The second passed already, since overwriting happened to give the right answer when the boundary fell inside the span, which is why the bug was easy to miss.

An empty overlap is left to resolve to an empty result rather than being special-cased. `year` 2026 with `after` 2027 yields a start past its end, and the SQL correctly returns nothing.

test/unit/php/includes/tests/core/classes/event/class-test-query.php
```diff
@@ -2545,4 +2551,255 @@ public function test_event_month_filter_matches_an_event_spanning_the_month(): v
...
+	protected function create_june_events(): array {
+		$events = array(
+			'within'   => array( '2026-06-15 10:00:00', '2026-06-15 14:00:00', 'America/New_York' ),
+			'before'   => array( '2026-05-15 10:00:00', '2026-05-15 14:00:00', 'America/New_York' ),
+			'spanning' => array( '2026-05-30 10:00:00', '2026-06-02 14:00:00', 'America/New_York' ),
+		);
+		$ids    = array();
+
+		foreach ( $events as $part => $datetimes ) {
+			$ids[ $part ] = $this->mock->post( array( 'post_type' => Event::POST_TYPE ) )->get()->ID;
```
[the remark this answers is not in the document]
mauteri: Fixed in 5729c72f3. A genuine time bomb, and one I would not have caught until it broke.

`create_june_events()` now sets an explicit `post_date` of `2026-02-01 09:00:00` on all three fixtures, so the assertion means the same thing whenever it runs.

February rather than June is deliberate: the fixtures happen in May and June by event date, so a February publish date keeps the two axes distinct. That let me add a second assertion to the same test, that an event held inside June still comes back from a `post_date` query for 2026, which pins down that publish-date filtering ignores when the event itself happens.

carstingaxion:
@mauteri This is a PR worth using our existing tools ;)

Could you please add this or similar as PR blueprint override?

```json
{
	"$schema": "https://gatherpress.org/playground-preview/pr-override-schema.json",
	"appendSteps": [
		{
			"step": "installPlugin",
			"pluginData": {
				"resource": "url",
				"url": "https://github.com/carstingaxion/gatherpress-calendar/releases/latest/download/gatherpress-calendar.zip"
			},
			"options": {
				"activate": true
			}
		},
		{
			"step": "installPlugin",
			"pluginData": {
				"resource": "wordpress.org/plugins",
				"slug": "advanced-query-loop"
			},
			"options": {
				"activate": true
			}
		},
		{
			"step": "installPlugin",
			"pluginData": {
				"resource": "wordpress.org/plugins",
				"slug": "query-monitor"
			},
			"options": {
				"activate": true
			}
		},
		{
			"step": "writeFile",
			"path": "/wordpress/wp-content/mu-plugins/disable-gatherpress-calendar-filter.php",
			"data": "<?php\n/**\n * Plugin Name: Disables GatherPress Calendar posts_where filter\n * Description: Forces the calendar to use gatherpress core event query capabilities\n */\nadd_action( 'plugins_loaded', function() {\n\tremove_filter( 'posts_where', 'GatherPress\\Calendar\\posts_where', 10 );\n\t} );\n"
		}
	]
}
```

mauteri:
Added in e4a1a25a5. The preview bot will regenerate the Playground link on this PR with gatherpress-calendar, Advanced Query Loop, and Query Monitor installed, and your mu-plugin dropping the calendar's own `posts_where` filter so its queries go through the core date query.

carstingaxion:
Cool thing that works well @mauteri . 

Tested with the updated PR Playground and both plugins and can confirm:

- [x] Advanced Query Loop with "AQL: Date rangequery" lists all events in a given range
- [x] GatherPress Calendar works, renders the calendar well in backend and frontend, with all events of a given month 

_(Tested & tried some time and with different settings and combinations and wasn't able to find any problems)_

I'd be happy to see this come alive in GatherPress core, especially because it would be a great first step towards the inclusion of a Calendar block in GatherPress core and finally solve #369 !

includes/core/classes/event/class-date-query.php
```diff
@@ -0,0 +1,347 @@
...
+	const EVENT_COLUMNS = array( self::DEFAULT_COLUMN, 'datetime_start' );
+
+	/**
+	 * Arguments a clause may carry and still be resolved.
+	 *
+	 * A clause carrying anything else is refused outright rather than
+	 * resolved in part. Honoring `year` while quietly dropping `week` would
+	 * hand back a window a whole year wide, and the caller lifts the original
+	 * `date_query` out of the query once a window comes back, so there is
+	 * nothing left to apply the dropped argument.
```
carstingaxion, line 75: What ...?
I don't get it. Dropped argument? window? 

Please clarify.

includes/core/classes/event/class-date-query.php
```diff
@@ -0,0 +1,347 @@
...
+	 * ones, resolves to nothing and is left for WordPress.
+	 *
+	 * Arguments narrow each other the way they do in a WordPress clause, so a
+	 * calendar span alongside an `after` or `before` yields their overlap
+	 * rather than whichever was read last.
+	 *
+	 * Boundaries come back in the clock of the column they will be compared
+	 * against: UTC for the GMT pair, the site's own time for the local pair. A
+	 * clause naming a column outside `EVENT_COLUMNS` resolves to nothing, so
+	 * the caller leaves it to WordPress.
```
carstingaxion, line 113: This is (for me) also hard to read and understand. Can this be simplified.

carstingaxion (changes requested):

includes/core/classes/event/class-date-query.php
mauteri: Fair, that paragraph was unreadable. It used "window" before anything defined it, said "the caller" without naming it, and referred to a dropped argument it had never introduced.

To answer the question directly: a "window" is the stretch of time this class hands back, a start and an end that an event has to fall inside. A "dropped argument" is one we would have ignored if we resolved a clause only partly.

The rule is that we read every argument in a clause or none of them. `array( "year" => 2026, "week" => 1 )` is the case: `year` is supported and `week` is not. If we read `year` alone we would match all of 2026 instead of one week of it, and `Query` pulls the date query out of the request as soon as this class answers, so nothing downstream would ever apply the `week` part. Wrong results, silently.

Rewritten along those lines in c721ce3c1, with that example in the docblock. I ran it against the code to be sure: it resolves to null.

includes/core/classes/event/class-date-query.php
mauteri: Agreed, and rewritten in c721ce3c1. It was four dense paragraphs describing behavior instead of showing it.

It now says, in order: what it returns, which arguments it reads and why the repeating ones (`week`, `dayofweek`, `hour`) are left alone, that it reads the whole query or none of it, and how the clock of the returned boundaries is chosen.

The intersection rule is the part that most needed an example rather than a sentence, so it now reads: `array( "year" => 2026, "after" => "2026-06-01" )` gives June to December 2026, not the whole year and not everything after June. I verified that against the code rather than asserting it, and it resolves to `2026-06-01 00:00:01` through `2026-12-31 23:59:59`.

Documentation only, no behavior change. Let me know if any of it is still heavy going.

PR #2273, opened 2026-09-07
⛔ DO NOT MERGE: recurring events, all six PRs combined for sizing and testing

mauteri:
# ⛔ DO NOT MERGE

This is a throwaway branch for sizing and testing. It is not a review target, and nothing here should land from this PR.

All of the work is @e1337geek's, from the six stacked drafts in his fork. Commit authorship is preserved. The only additions are conflict resolutions and code drift: his stack is based on develop from 85 commits back, and reconciling it touches 18 files. Those resolutions are mine, not his, and are called out below. Opened here so we can see the whole feature at once, run it in Playground, and try it locally, none of which is possible while the stack lives across a fork.

The plan is unchanged: review and land the six PRs one at a time, folding in edits as they go. This just tells us what we are dealing with first.

## Source

Built from the tip of the stack, `feat/80-recurrence-pr-6`, which carries all six:

| | Draft | Head branch |
| --- | --- | --- |
| 1 | [Storage and expansion engine](https://github.com/e1337geek/gatherpress/pull/6) | `feat/80-recurrence-pr-1` |
| 2 | [Front end and URL routing](https://github.com/e1337geek/gatherpress/pull/7) | `feat/80-recurrence-pr-2` |
| 3 | [Per-occurrence RSVP](https://github.com/e1337geek/gatherpress/pull/8) | `feat/80-recurrence-pr-3` |
| 4 | [Authoring UI and admin list](https://github.com/e1337geek/gatherpress/pull/9) | `feat/80-recurrence-pr-4` |
| 5 | [Calendar export](https://github.com/e1337geek/gatherpress/pull/10) | `feat/80-recurrence-pr-5` |
| 6 | [Forward splits](https://github.com/e1337geek/gatherpress/pull/11) | `feat/80-recurrence-pr-6` |

Refs #80.

## How big it actually is

**146 files, +78,735 / -544.** Three quarters of that is tests:

| Area | Added lines |
| --- | --- |
| `test/unit` | 58,723 |
| `includes/core` | 15,601 |
| `test/e2e` | 2,089 |
| `src/panels` | 2,086 |
| everything else | ~236 |

So the production surface is roughly **15.6k lines**, largely a new `includes/core/classes/event/recurrence/` subsystem. The single largest file is a test at 5,970 lines; the largest production file is `class-occurrences.php` at 2,607.

## Two things to know before testing

**It is based on develop from 85 commits ago.** I have not merged develop in, deliberately: resolving conflicts across 78k lines of a feature I did not write is not something to do silently. If it does not apply cleanly, that is a question for @e1337geek rather than something to paper over here.

**One commit overlaps with work already on develop.** "Raise the PHPStan memory limit from 2G to 4G" duplicates #2267, which landed for the same reason: the 2G ceiling had been outgrown and was failing unrelated PRs. Worth dropping from the stack when PR 1 is restaged, since develop already has it.

There is also a separate follow-up for whoever restages these: #2270 replaced the post type support string literals with class constants (`Event::SUPPORT` and friends). Any use of the literals in this stack should move over at the same time.

---

# Merge notes

`develop` is merged in as of 82289db2, so the branch builds and runs. **The resolutions below are mine, not @e1337geek's.** Where the two sides disagreed I kept the recurrence side, since exercising that feature is the point of this branch. That means develop work was dropped in places, and every instance is listed here so nothing is carried into the real PRs by accident.

18 files conflicted across 37 hunks.

## Kept both sides

- **`rsvp/class-rsvp.php`** — develop's provider guard and null-post return, plus Brady's early cache invalidation. The `no_status` path trashes the stored comment and returned before invalidating, leaving removed attendees in warm counts for `Cache::CACHE_EXPIRATION`.
- **`rsvp/class-cleanup.php`** — Brady's deferred term counting (3n recount queries collapse to one per taxonomy) and develop's `(string) $meta_key` cast.
- **`event/class-query.php`** — Brady's total ordering, rewritten onto develop's `$events_table` / `$posts_table`. Without the ID tiebreaker a tied pair can appear on two pages of a paginated archive and another on none.
- **`event/class-event.php`** — his keyed datetime cache typed as `array<string, EventDatetime>`, keeping develop's phpstan alias.

## Recurrence kept, develop work dropped

**Both of the first two are fine left broken on this branch.** Nothing here is going to production, and the point is to get a feel for how the feature behaves before each PR is fleshed out for review. They are recorded because each carries real work that has to be resolved as the PRs land one at a time, not because either needs fixing today.

### 1. Security: the `rsvp_status_html` route lost its signature check

`event/class-rest-api.php`. develop added `block_signature` verification to that route; the merged file keeps one reference where develop has four. Two develop tests fail proving the hole is open: a wrong signature returns 200 instead of 403, and a missing one returns 200 instead of 400.

This one must be reconciled before PR 3 (per-occurrence RSVP) merges. It is not a question of design, just of carrying develop's verification through his rewrite of the route.

### 2. Conflicting feature: recurrence versus all-day calendar export

`calendar/class-calendar.php`. Brady restructured `get_vevent()` around recurrence while develop separately added all-day handling (`DTSTART;VALUE=DATE:`), `fold_ical_text()` for RFC 5545 line folding, `$post` null guards, and a DTSTAMP fallback. His `datetime_lines()` has no all-day branch at all, so taking his side drops all of it.

Unlike the first, this needs a product decision rather than a merge. His projected-occurrence path emits `DTSTART;TZID=…` from `datetime_start_gmt`, while an all-day occurrence needs `VALUE=DATE`. **Neither branch has ever answered what an all-day recurring event exports as**, so there was no correct resolution to pick and I did not invent one. Worth settling before PR 5 (calendar export) is reviewed.

### 3. His archive refactor replaced develop's guard

`event/class-setup.php`. `is_event_archive_request()` and `first_queried_event_post_type()` supersede develop's inline check. `Test_Setup::test_handle_event_archive_redirect_array_post_type` fails against it, so the two disagree about the widened-array `post_type` case. Minor next to the two above, but the disagreement is real.

### 4. Six test files took the recurrence side

Dropping develop's newer tests there: the two RSVP block `view.test.js` files, `class-test-cache.php`, `class-test-calendar.php`, `class-test-admin-list.php` (a 1,466-line conflict on its own), and `class-test-rsvp.php`.

## Drift fixed to make it run

develop moved under the stack in ways that were fatal rather than conflicting, so these were mechanical repairs:

- **`Event::$event` is now `Event::$post`** — 12 dereferences in `class-calendar.php`.
- **`Event::$rsvp` was removed** — three REST call sites moved to develop's `( new Rsvp( $post_id ) )->…` pattern. Brady's tests still call `$event->rsvp->save()`, which is the one erroring test.
- **Missing `use WP_Comment;`** in `event/class-rest-api.php`.
- **`$this->event->ID` inside `Event` itself** became `$this->post->ID ?? 0`.

Worth knowing that `GatherPress\Core\Rsvp` and `GatherPress\Core\Calendar` resolve through the alias file develop added, which is why the import conflicts took develop's shorter form.

## State

- **Builds.** `npm run build` compiles.
- **Runs.** 2,912 tests, 1 error, 4 failures. The error is test-side (`$event->rsvp`), the four failures are items 1 and 3 above.
- **PHPStan is red, and mostly not from this merge.** develop raised the level from 5 to 8. At level 5, the bar the stack was written and CI-verified against, the merged tree reports **1 error**. At level 8 it reports **109**, of which 93 are missing iterable/generic docblock types.

## Still to do when restaging

- Drop the "Raise the PHPStan memory limit from 2G to 4G" commit; #2267 already did it on develop.
- Move post type support literals to the #2270 constants (`Event::SUPPORT` and friends).
- Bring the docblock types up to PHPStan level 8.


carstingaxion:
Wow! Absolutely cool to play with this idea @e1337geek !!

In general this looks good and feels like a natural part of GatherPress. Until now, everything works as described in the single PR descriptions and I had no problem using the feature.

Some findings for later review, further explanantion or fixing:

|  |  |  |  |
|--------|--------|--------|--------|
| After doing two RSVPs to different occurences of the same event, I couldn't find a way to differentiate them in the backend. Asking myself, which belonged to what date? The linked events in the RSVPs admin column both go to `.../event/critical-mass/`, not the generated date links, that the frontend provides like `.../event/critical-mass/20270625T180000/` | image | image | image | 
| The "Add to calendar" links all point to the "initial" event, not the currently queried occurence. This is OK for the downloads, because the generated ical has RRULE set and working, but the direct calendar links for Google & Yahoo should point to the actual occurence. | image |  |  |
| Querying events by random, works on the frontend, but not in the editor. As you can see, all events show as the same date. Other ORDERBY statements do also not work or at least seem-to-not-work, as the video shows. | image | image |  |

[Bildschirmaufzeichnung vom 13.09.2026, 12:55:07.webm](https://github.com/user-attachments/assets/de6b6b67-ec12-4c35-97b6-23f3e04bce63)

|  |  |  |  |
|--------|--------|--------|--------|
| A selected "Event Date" block, inside a query loop, can have a lot more controls and sidebar panels than regularly. | image |  |  |
| Canceling one occurence works. But I'm wondering, what we are going to do, as soon as GaterPress supports Events Statuses to be set. This has some overlap in naming and maybe even functionality, I think. | image |  |  |
| To make the occurences better visible to search engines, I guess they should be added to the sitemap, but I'm not sure. | image |  |  |
| The startdate of the currently edited event should be updated in GatherPress' data store, while I create my RRULE from the UI. That would instantly update the start date on the Event settings panel, and inside the editor preview for the Event Date block, which otherwise happens earliest after the first safe. The UI in general needs some love. | image |  |  |

PR #2279, opened 2026-09-07
Feature/681 uninstall data removal

### LINKED ISSUE #681: Add uninstall.php or uninstall routine

mattcowan:
Fixes #681

## Proposed changes:

* **An Uninstall settings screen, at the end of the GatherPress settings tabs.** It lists what deleting the plugin can remove. Every checkbox is off when you install the plugin. If you change nothing, deleting GatherPress removes no content.
* **One switch for each kind of data:** events, RSVPs, topics, venues, generated files, scheduled jobs, user preferences, and the plugin settings.
* **Events and venues answer separately**, because someone may want to keep the places and lose the gatherings, or the reverse. Two things travel with their owner rather than sitting on a switch of their own: the event date table goes with events, since every row in it is keyed by `post_id`, and the `_gatherpress_venue` shadow terms go with venues, since each one names a venue post.
* **RSVPs follow events.** An RSVP whose event was deleted is an orphan, so removing events removes them either way, and the RSVP row hides once events are selected. While events are staying, RSVPs remain a choice of their own.
* **The choices are ordinary settings that cannot travel.** This adds `exportable` and `importable` to the Settings API. An option marked `'exportable' => false` is left out of a settings export and refused by an import, and an import now reports what it refused separately from keys it does not recognize. Every uninstall choice is declared that way: every other setting shows its effect as soon as it is wrong, while these do nothing until the plugin is deleted, possibly months later and possibly for somebody who never ran the import.
* **Multisite is decided by the network.** The opt-ins resolve per site inside the uninstall fan-out, so a network administrator who adds them to the inheritance list on the Network tab answers once for every site, and one who leaves them off it lets each site answer for its own data. What no single site owns, the network settings and the shared user table, follows the network's own answer.
* **The consequence is shown where the action is.** A warning on the Uninstall screen says what is and is not reversible, and the GatherPress row on the Plugins screen names the selected data next to the Delete link, in the same shape core uses for an unmet dependency.
* **Things the plugin used to leave behind are now removed:** the `_gatherpress_rsvp_flag` taxonomy, the per-user preferences from the profile and the RSVP screen, `gatherpress_calendar_last_modified`, and `gatherpress_network_settings`.
* **Bookkeeping that SQL deletes do not maintain is now corrected.** `term_taxonomy.count` is decremented for taxonomies that outlive the posts, so a category that had ten events stops reporting ten. `posts.comment_count` is decremented when RSVPs go and their events stay. The object cache is flushed only when a task that writes SQL has actually run.
* **The post relationship delete no longer reaches into RSVP terms.** A comment ID can equal a post ID and nothing in the schema tells the two apart, so the delete is scoped by taxonomy as well.
* **The coverage gate now checks the new settings screen.** `.github/coverage-config.json` skipped it by accident. The pattern `*uninstall.php` is there for the root `uninstall.php`, which no test can load. `should_exclude_file()` turns `*` into `.*`, which also matches `/`, so the pattern matched `includes/core/classes/settings/class-uninstall.php` as well. Both patterns now have no leading `*`. This agrees with `sonar-project.properties`, which already names the two files without one.

This builds on the task shape from #2210. That PR added `Base`, `Setup` and the always-on `Notices` and `Transients` tasks. This PR adds the settings gate and the tasks that remove data.

### Other information:

## Testing instructions:

**Please back up the database first. Steps 4 and after delete content.**

1. Go to **GatherPress > Uninstall**. Confirm that every checkbox is off.
2. Create some events, venues, topics and RSVPs, or use a site that has them.
3. Record what the site holds:

    ```
    wp post list --post_type=gatherpress_event --format=count
    wp post list --post_type=gatherpress_venue --format=count
    wp term list _gatherpress_venue --format=count
    wp eval 'global $wpdb; echo $wpdb->get_var( "SELECT COUNT(*) FROM {$wpdb->comments} WHERE comment_type = \"gatherpress_rsvp\"" );'
    ```

4. Deactivate and uninstall the plugin, but keep the files:

    ```
    wp plugin deactivate gatherpress
    wp plugin uninstall gatherpress --skip-delete
    ```

5. Run the commands from step 3 again. **All the counts must be the same.** Only the transients go, because they are a cache. If any content is missing, a task ran that you did not ask for.
6. Activate the plugin again. Go to **GatherPress > Uninstall**. Turn on every checkbox and save. Confirm that the RSVP row disappears once Events is checked, and that the GatherPress row on the Plugins screen now warns that data will go.
7. Repeat step 4, then step 3. **All the counts must now be zero.** Confirm that these also go:

    ```
    wp eval 'global $wpdb; print_r( $wpdb->get_col( "SHOW TABLES LIKE \"%gatherpress%\"" ) );'
    wp eval 'global $wpdb; print_r( $wpdb->get_col( "SELECT option_name FROM {$wpdb->options} WHERE option_name LIKE \"gatherpress%\"" ) );'
    wp eval 'global $wpdb; print_r( $wpdb->get_col( "SELECT DISTINCT meta_key FROM {$wpdb->usermeta} WHERE meta_key LIKE \"gatherpress%\"" ) );'
    wp cron event list --fields=hook --format=csv
    ls wp-content/uploads/gatherpress
    ```

8. Confirm that the plugin removes nothing that it does not own. Comments that are not RSVPs must stay. An option such as `theme_mods_gatherpress-org-theme` belongs to a theme, and must stay. A category that was applied to a deleted event must report the right count afterwards.
9. Confirm that the choices do not travel. Export the settings from **GatherPress > Tools**, and check the file holds no `uninstall_*` key. Import that file on another site and confirm nothing is armed there.
10. On multisite, set the choices in **Network Admin > Settings > GatherPress > Uninstall**, then add the `uninstall_*` keys to the inheritance list on the **Network** tab. Every subsite should show those fields locked, reading "Inherited from the network". Remove them from the list and each subsite answers for itself again.
11. Restore the database.

To see what will run before you run it:

```
wp eval 'foreach ( GatherPress\Core\Uninstall\Setup::get_instance()->get_tasks() as $t ) { printf( "%-46s %s\n", get_class( $t ), $t->applies() ? "WILL RUN" : "skipped" ); }'
```

My WordPress.org username: matthewneilcowan

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Add an Uninstall settings screen that controls what deleting the plugin removes. Every option is off by default, so deleting GatherPress keeps your data unless you ask for it to go. You can opt in separately to removing events, RSVPs, topics, venues, generated files, scheduled jobs, per-user preferences, and the plugin settings. The choices never travel in a settings export and an import will not set them. The Plugins screen says so in the GatherPress row when anything is selected, so the consequence is beside the Delete link that triggers it. On multisite a network administrator can answer once for every site or leave each site to answer for its own data.

</details>

## Summary by CodeRabbit

- **New Features**
  - Added opt-in uninstall controls for events, venues, RSVPs, topics, generated files, scheduled tasks, user data, and settings.
  - Added multisite-aware uninstall preferences and cleanup behavior.
  - Added Plugins screen warnings showing which data will be removed when GatherPress is deleted.

- **Bug Fixes**
  - Settings imports now preserve non-importable options, report them separately, and display blocked-setting counts.
  - Settings exports omit options marked as non-exportable.
  - Improved checkbox-dependent settings behavior.

- **Documentation**
  - Updated guidance for settings import/export controls and multisite scope.


includes/templates/admin/settings/uninstall.php
```diff
@@ -0,0 +1,113 @@
...
+		<legend class="screen-reader-text">
+			<?php esc_html_e( 'Data to remove when the plugin is deleted', 'gatherpress' ); ?>
+		</legend>
+
+		<?php foreach ( $gatherpress_tasks as $gatherpress_key => $gatherpress_task ) : ?>
+			<?php
+			$gatherpress_field_id    = 'gatherpress-uninstall-' . $gatherpress_key;
+			$gatherpress_describedby = $gatherpress_field_id . '-description';
+			?>
+			<p>
```
mauteri: Styling needs some TLC here. Label and description sit on one line at the same size and color, so "Events and venues Removes every event and venue" reads as a sentence. The `<span class="description">` is inline, and WordPress only greys and separates `.description` when it is a `<p>`.

The other tabs avoid this because they render through `do_settings_sections()`, which is core's `form-table` layout: the option name in a `<th scope="row">`, the checkbox with a short label in the `<td>`, and a grey `<p class="description">` beneath it. Settings > Reading looks the same, so it is what people already know.

Suggest rendering these six the same way, no CSS needed:

```html
<table class="form-table" role="presentation">
	<tr>
		<th scope="row">Events and venues</th>
		<td>
			<label for="gatherpress-uninstall-posts">
				<input type="checkbox" id="gatherpress-uninstall-posts" ...>
				Remove when the plugin is deleted
			</label>
			<p class="description" id="...-description">Removes every event and venue, with their meta, revisions, and the RSVPs recorded against them.</p>
		</td>
	</tr>
</table>
```

Keep the `fieldset`, `legend` and `aria-describedby` around it so nothing accessible is lost. That gives a proper label column, one consistent checkbox phrase, and the explanation in grey underneath, matching the rest of the settings screens.

mattcowan: [f221e67](https://github.com/GatherPress/gatherpress/pull/2279/commits/f221e679a80214cfb55d6818116e740d25631ae4) - The six rows now use `form-table`: the name in `<th scope="row">`, the checkbox in the `<td>`, and the explanation in a `<p class="description">` below it. The `fieldset`, `legend` and `aria-describedby` stay as they were.

One change from your suggestion. Each checkbox keeps its own label, such as "Remove RSVPs when the plugin is deleted", instead of one shared phrase. A checkbox takes its accessible name from its label, and `role="presentation"` also takes the `<th>` out of the accessibility tree. One shared phrase therefore gives all six checkboxes the same name.

Settings > Discussion shows the pattern for this. Under the "Default post settings" heading, each of the three checkboxes has a full label of its own, such as "Allow people to submit comments on new posts". Core also repeats the row title in a `screen-reader-text` legend there, because the `<th>` does not reach the controls.

Tell me if you prefer your wording and I will change it.

uninstall-settings-after

includes/core/classes/uninstall/class-terms.php
```diff
@@ -0,0 +1,159 @@
...
+				. ' INNER JOIN %i AS tt ON tt.term_id = tm.term_id'
+				. ' WHERE tt.taxonomy IN ( %s, %s, %s, %s )',
+				$wpdb->termmeta,
+				$wpdb->term_taxonomy,
+				Topic::TAXONOMY,
+				Venue::TAXONOMY,
+				Status::TAXONOMY,
+				Provider::TAXONOMY
+			)
+		);
```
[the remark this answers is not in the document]
mattcowan: [0c3440e](https://github.com/GatherPress/gatherpress/pull/2279/commits/0c3440e6785a21b1ea495f4e9432d13f6ef0a708) - The termmeta delete now has the same `NOT EXISTS` condition as the term row delete below it.

`test_keeps_term_row_shared_with_another_taxonomy` now adds term meta to the shared term and asserts that the meta row survives.

I checked the test against the bug. With the condition removed, the new assertion fails with "Failed asserting that 0 is identical to 1". `test_removes_plugin_terms_and_relationships` continues to pass, because its term is in one taxonomy only. The condition therefore keeps the meta of a shared term and still removes the meta of a term that only this plugin uses.

test/unit/php/includes/tests/core/classes/settings/class-test-uninstall.php
```diff
@@ -0,0 +1,347 @@
...
+		$_POST[ Uninstall::NONCE_NAME ] = 'not-a-valid-nonce';
+
+		$this->expectException( \WPDieException::class );
+
+		try {
+			Uninstall::get_instance()->handle_save();
+		} finally {
+			unset( $_POST[ Uninstall::NONCE_NAME ] );
+		}
+	}
```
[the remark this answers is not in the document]
mattcowan: The observation is correct, but the suggested code cannot run.

`phpunit.xml.dist` excludes the `multisite` group and `phpunit-multisite.xml.dist` includes only that group. A test without `@group multisite` therefore runs on a single site only, where `is_multisite()` is always false. The `grant_super_admin()` branch never executes, and the coverage does not change.

A correct version is a second test with `@group multisite` that grants super admin before the call. I have not added one. The nonce check already has single-site coverage, and both environments run the same code in `handle_save()`. The network-settings test in this pull request is different, because `delete_site_option()` and `delete_option()` behave differently on multisite.

mattcowan:
Hey @mauteri, in the meeting today I made a note about doing this with wp-cron. The shape in #2210 and your note on #2041 was synchronous, so could you give some direction where the cron job would get scheduled? I would want to be sure I have your intent before I change the design. The approach I have submitted here would delete the data on uninstall.

If a cron job is scheduled from `uninstall.php`, it cannot run. `delete_plugins()` calls `uninstall_plugin()` and then deletes the plugin directory in the same request — `wp-admin/includes/plugin.php` lines 970 and 989 on WP 7.1. The callback file is gone before the event is due, so WP-Cron fires the hook, finds no listener, and unschedules it. Nothing is removed, and nothing reports an error.

Deactivation can schedule it, because the files stay in place, but I do not think it holds up either. WordPress offers Delete only for an inactive plugin (`class-wp-plugins-list-table.php` line 946), so every delete follows a deactivate. The sweep would race the delete, and whether the data goes would depend on whether WP-Cron fired in the gap. It would also need unscheduling on activation, or a site that deactivates to test a conflict loses its data.

Did you have something different in mind for this? I would be happy to change direction here if so.

mauteri:
You read the constraints right, and I hadn't followed them through in the meeting. Anything scheduled from `uninstall.php` fires into nothing, and anything scheduled from deactivation races the delete. So no change of direction: keep the synchronous sweep exactly as you have it.

The case I was reaching for is a large site or network, where one request of bulk deletes can outlive a web server timeout. That doesn't change this PR. It's a gate in front of it, and it'll be a follow-up: scope what the checked tasks would remove, show the counts on the Uninstall screen, and above a threshold refuse the Plugins-screen delete with a message pointing at `wp plugin uninstall gatherpress`, since WP-CLI has no timeout and is what operators of big sites use anyway. `uninstall.php` can tell it's under WP-CLI in-process, and it runs before the files are removed, so a refusal leaves the plugin installed and the data untouched. The idempotent tasks are what make that safe.

Two small things for this PR, both cheap:

- `set_time_limit( 0 )` and `ignore_user_abort( true )` at the top of `uninstall.php`, so PHP's own limit doesn't cut a sweep short where the host allows it.
- A line on the Uninstall screen recommending WP-CLI for large sites, so the advice exists before the gate does.

mattcowan:
[48066c8](https://github.com/GatherPress/gatherpress/pull/2279/commits/48066c80de622968bd4ec4a889514bcbc3e3e0c6) should have both above bullet points addressed. One note here: the recommended command needs `--deactivate`. `wp plugin uninstall` warns and skips while the plugin is active, which it always is for someone reading that screen, the plain command answers `Error: No plugins uninstalled.`

That impacts the gate you described. If the refusal message points at `wp plugin uninstall gatherpress`, it sends the operator to a command that refuses.

I also updated three docblocks that still described #681 as a coming follow-up, which is why `class-base.php` and `class-setup.php` are in the diff.

includes/core/classes/settings/class-uninstall.php
```diff
@@ -0,0 +1,249 @@
...
+	 *
+	 * @since 0.36.0
+	 *
+	 * @return void
+	 */
+	protected function setup_hooks(): void {
+		parent::setup_hooks();
+
+		add_action( 'gatherpress_settings_section', array( $this, 'settings_section' ), 9 );
+		add_action( 'admin_post_' . self::SAVE_ACTION, array( $this, 'handle_save' ) );
```
carstingaxion: Use sprintf to combine the action name.

includes/templates/admin/settings/uninstall.php
```diff
@@ -0,0 +1,141 @@
...
+$gatherpress_scope       = isset( $scope ) && 'network' === $scope ? 'network' : 'blog';
+
+/*
+ * `label` names the row, the way core's settings screens do. `checkbox` is the
+ * control's own label, and says what it does without the row heading: the table
+ * is `role="presentation"`, so the heading is not read out with the control.
+ */
+$gatherpress_tasks = array(
+	Preferences::TASK_POSTS    => array(
+		'label'       => __( 'Events and venues', 'gatherpress' ),
```
carstingaxion: We tried to avoid hardcoded post type labels in the past. Use the post type labels with sprintf and proper translator comments or use with _x for context.

Have a look in the codebase, I guess we already have some helper functions that you could leverage.

includes/templates/admin/settings/uninstall.php
```diff
@@ -0,0 +1,141 @@
...
+		'checkbox'    => __( 'Remove events and venues', 'gatherpress' ),
+		'description' => __( 'Removes every event and venue, with their meta, revisions, and the RSVPs recorded against them.', 'gatherpress' ),
+	),
+	Preferences::TASK_COMMENTS => array(
+		'label'       => __( 'RSVPs', 'gatherpress' ),
+		'checkbox'    => __( 'Remove RSVPs', 'gatherpress' ),
+		'description' => __( 'Removes every RSVP record and the answers people gave to custom RSVP fields.', 'gatherpress' ),
+	),
+	Preferences::TASK_TERMS    => array(
+		'label'       => __( 'Topics and venue terms', 'gatherpress' ),
```
carstingaxion: Even this might be technically correct, that the step will delete the venue terms. 

1. they might be already deleted by step one (events and venues) powered through the normal hook that deletes a venue term, when it's post is deleted.
2. It might confuse normal users to read a second time about venue-removing on this same screen. Maybe we just drop it here.

includes/core/classes/settings/class-uninstall.php
mattcowan: Fixed in [93214d5](https://github.com/GatherPress/gatherpress/pull/2279/commits/93214d5afb4d00951b4b5d1913c915149adb3ee8)

includes/templates/admin/settings/uninstall.php
mattcowan: Made changes to use helper functions in 
93214d5, `Utility::post_type_label()` and `Utility::taxonomy_label()`, with translator comments on the format strings.

includes/templates/admin/settings/uninstall.php
mattcowan: Kept the deletion, but changed the label, in [93214d5](https://github.com/GatherPress/gatherpress/pull/2279/commits/93214d5afb4d00951b4b5d1913c915149adb3ee8) - please let me know if you disagree here though.

The posts step does not remove the venue terms here. That step is one SQL statement, so `delete_post_{$type}` never fires, and `uninstall.php` runs with the plugin unloaded, so `Shadow_Source::delete_term()` is not hooked at all. Dropping the venue terms from this task would leave orphaned rows in `term_taxonomy` and `termmeta`.

The row now reads as just "Topics", and the description names the internal venue and RSVP terms that go with it, so venues appear once on the screen.

includes/core/classes/settings/class-uninstall.php
```diff
@@ -0,0 +1,249 @@
...
+	 * Set up hooks for various purposes.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @return void
+	 */
+	protected function setup_hooks(): void {
+		parent::setup_hooks();
+
+		add_action( 'gatherpress_settings_section', array( $this, 'settings_section' ), 9 );
```
carstingaxion, line 72: Add a comment why prio 9 is needed here, please.

includes/core/classes/settings/class-uninstall.php
```diff
@@ -0,0 +1,249 @@
...
+
+	/**
+	 * Get the name for the uninstall settings page.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @return string The localized name for the uninstall settings page.
+	 */
+	protected function get_name(): string {
+		return __( 'Uninstall', 'gatherpress' );
```
carstingaxion, line 95: I guess (but am not sure) that it would be ok to just use the "default" textdomain here, to keep the GatherPress translations free from simple defaults. 

I know we discussed this in the past at some point...

includes/templates/admin/settings/uninstall.php
```diff
@@ -0,0 +1,169 @@
...
+	Preferences::TASK_TERMS    => array(
+		'label'       => Utility::taxonomy_label( 'name', Topic::TAXONOMY ),
+		'checkbox'    => sprintf(
+			/* translators: %s: Plural taxonomy label (e.g. "Topics"). */
+			__( 'Remove %s', 'gatherpress' ),
+			Utility::taxonomy_label( 'name', Topic::TAXONOMY )
+		),
+		'description' => sprintf(
+			/* translators: 1: Plural taxonomy label (e.g. "Topics"), 2: Plural post type label (e.g. "Venues"). */
+			__( 'Removes all %1$s and the internal terms GatherPress keeps for %2$s and RSVP records. Categories and tags are not touched.', 'gatherpress' ),
```
carstingaxion: I would leave "Categories and tags" out here, because GatherPress doesn't reference them elsewhere (and don't use them).


BUT, from my work on carstingaxion/gatherpress-export-import, I know that a lot of other event plugins do use the defaults taxonomies also for their events post type. So there could be a real chance that someone who migrates to GatherPress from another plugin, or someone just liking the idea, might register the core categories and tags with gatherpress_event posts.

Maybe it could be from interest to ask (programmatically) for all taxonomies registered with our post types, and delete all relations in the next step. @mauteri what do you mean?

includes/templates/admin/settings/uninstall.php
```diff
@@ -0,0 +1,169 @@
...
+		'checkbox'    => __( 'Remove settings', 'gatherpress' ),
+		'description' => __( 'Removes the GatherPress settings, including the choices on this screen.', 'gatherpress' ),
+	),
+);
+?>
+
+<h2><?php esc_html_e( 'Uninstall', 'gatherpress' ); ?></h2>
+
+<p class="description">
+	<?php esc_html_e( 'Choose what GatherPress removes when the plugin is deleted from the Plugins screen. Everything is off by default, so deleting the plugin keeps your data unless you ask for it to go.', 'gatherpress' ); ?>
```
carstingaxion: In WordPress settings I usually "select" or "choose", but do not "ask". Maybe only me with bad English skills, but don't know.

includes/templates/admin/settings/uninstall.php
```diff
@@ -0,0 +1,169 @@
...
+</p>
+
+<div class="notice notice-warning inline">
+	<p>
+		<strong><?php esc_html_e( 'These choices cannot be undone.', 'gatherpress' ); ?></strong>
+		<?php esc_html_e( 'Deactivating the plugin removes nothing. The data goes only when the plugin is deleted, and it cannot be recovered without a backup.', 'gatherpress' ); ?>
```
carstingaxion: This would be a nice last stop to hint users to the settings export and the regular wordpress core export as a last chance to backup later.

carstingaxion (dismissed):

includes/templates/admin/settings/uninstall.php
mauteri: I agree. Any core taxonomy we should leave. I would only remove things that GatherPress registers explicitly. Like Topics, but not Categories or Tags which are default taxonomies and can be used along with posts or other post types.

includes/core/classes/settings/class-uninstall.php
mattcowan: Added comment in [c110a81](https://github.com/GatherPress/gatherpress/pull/2279/commits/c110a81765921e3edc702f89bd3294f8160e7ee5)

includes/core/classes/settings/class-uninstall.php
mattcowan: I need to push back on this one, the ruleset does not allow it. `phpcs.ruleset.xml` line 82 enables `WordPress.WP.I18n` with `text_domain` set to `gatherpress`, and a call without the domain fails with `WordPress.WP.I18n.MissingArgDomainDefault`. I checked by changing this one and running PHPCS. No string in `includes/` uses the default domain today, so I have left it as it is.

includes/templates/admin/settings/uninstall.php
mattcowan: Dropped "Categories and tags are not touched" from the description in commit [ae473a7](https://github.com/GatherPress/gatherpress/pull/2279/commits/ae473a74de8b8d96663bc4acadcf8788150587c5)

On the migration case: the posts task already deletes `term_relationships` for every post it removes (`class-posts.php` line 118), so an event that carries core categories loses those relationships with the post. The category terms themselves stay, because other content uses them. That case is covered now without discovering the taxonomies. I can still change this if you or @mauteri want the terms of any taxonomy registered on our post types handled differently.

includes/templates/admin/settings/uninstall.php
mattcowan: Changed to "unless you select it here" in [ae473a7](https://github.com/GatherPress/gatherpress/pull/2279/commits/ae473a74de8b8d96663bc4acadcf8788150587c5)

includes/templates/admin/settings/uninstall.php
mattcowan: Added a sentence that points to the Tools tab for a settings export and to Tools > Export for events and venues.

mauteri:
@coderabbitai review

mauteri:
@coderabbitai review

includes/core/classes/class-settings.php
```diff
@@ -1481,16 +1594,18 @@ public function import_settings( array $data, string $mode = 'merge', string $sc
...
 
-		// Filter to only known keys.
-		$to_import = array_intersect_key(
-			$data['settings'],
-			$field_type_map
+		// Filter to only known keys, then drop the ones declared out of the
+		// importer's reach.
+		$to_import = array_diff_key(
+			array_intersect_key( $data['settings'], $field_type_map ),
+			array_flip( $this->get_non_importable_keys() )
```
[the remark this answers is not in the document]
mauteri: Agreed, and fixed in 25da8c6a1. Replace mode deleted the stored options before writing, so a refused key was reset to its default while the result told the administrator it was left alone. The blocked keys are now read before the delete and merged back after sanitizing. `test_replace_import_preserves_a_refused_key` covers it: the file asks to turn the key off, in replace mode, and the stored value survives.

includes/core/classes/settings/class-uninstall.php
```diff
@@ -0,0 +1,382 @@
...
+	 * @return string The admin URL of the uninstall settings page.
+	 */
+	public static function get_page_url(): string {
+		return add_query_arg(
+			array(
+				'post_type' => Event::POST_TYPE,
+				'page'      => Utility::prefix_key( self::get_instance()->slug ),
+			),
+			admin_url( 'edit.php' )
+		);
```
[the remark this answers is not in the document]
mauteri: Agreed, and fixed in 25da8c6a1. This matters more than minor, since the network Plugins screen is where a plugin is actually deleted on multisite. `armed_labels()` and `get_page_url()` take a `$network` flag, `Plugin_Row` passes `is_network_admin()`, and the link goes to the `uninstall_settings` tab of the network settings page. Site-admin behavior is unchanged, and both branches are covered.

includes/core/classes/uninstall/class-setup.php
```diff
@@ -99,8 +111,22 @@ public function get_tasks(): array {
 	 * @return void
 	 */
 	public function run(): void {
+		$flush = false;
+
 		foreach ( $this->tasks as $task ) {
+			$flush = $flush || ( $task->applies() && $task->invalidates_cache() );
```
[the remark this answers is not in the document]
mauteri: Agreed, and fixed in 25da8c6a1. Both cases you name were reachable. `Base::run()` now returns whether any cleanup ran, per site or network, and the registry decides the flush from that instead of from `applies()`. `test_run_flushes_for_a_network_only_task` covers the `Users` shape: `applies()` false, `applies_to_network()` true, cache flushed.

test/unit/php/includes/tests/core/classes/class-test-settings.php
```diff
@@ -3588,4 +3588,238 @@ public function test_get_map_tile_url_custom_setting_on_a_carto_host_is_not_keye
...
+	 * A flag left out of a declaration is on.
+	 *
+	 * @covers ::get_keys_opted_out_of
+	 *
+	 * @return void
+	 */
+	public function test_keys_opted_out_of_reads_an_absent_flag_as_on(): void {
+		$this->declare_travel_flag_fixture();
+
+		$keys = \PMC\Unit_Test\Utility::invoke_hidden_method(
```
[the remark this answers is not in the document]
mauteri: Agreed, and fixed in 25da8c6a1. The file already imports `PMC\Unit_Test\Utility` as `Utility`, with GatherPress's own aliased to `GatherPress_Utility`, so the fully qualified call was just inconsistent.

uninstall.php
```diff
@@ -23,6 +23,20 @@
...
+ * does not stop part-way and leave half the data behind. Neither call is
+ * guaranteed: a host can disable `set_time_limit()`, and a hard web server
+ * timeout is outside PHP's control. The Uninstall screen recommends WP-CLI
+ * for sites large enough for that to matter.
+ */
+if ( function_exists( 'set_time_limit' ) ) {
+	set_time_limit( 0 );
+}
+
+ignore_user_abort( true );
```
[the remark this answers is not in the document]
mauteri: Agreed, and fixed in 25da8c6a1. Guarded with `function_exists()` to match `set_time_limit()` directly above it. A host that disables it would have thrown before `Setup::run()`, so no cleanup would have run at all.

mauteri:
@coderabbitai review

All five findings from the last pass are addressed in 25da8c6a1, with a reply on each thread. Please confirm.

mauteri:
Thanks @mattcowan and @carstingaxion. The task shape and the opt-in premise here are Matt's and have not changed. A fair amount moved after the review though, so rather than make anyone reconstruct it from the commit list, here is what changed and why. The PR description has been rewritten to match.

## The choices are split up

Events and venues each answer for themselves now, instead of one combined switch. Someone may want to keep the places and lose the gatherings, or the reverse.

Two things travel with their owner rather than sitting on a switch of their own, because leaving either behind is corruption rather than conservation:

- The event date table goes with events. Every row in it is keyed by `post_id`.
- The `_gatherpress_venue` shadow terms go with venues. Each one names a venue post.

RSVPs follow events for the same reason, and the RSVP row hides once events are selected. While events are staying, RSVPs remain a choice of their own, because they can live on any post type that declares `gatherpress-rsvp`.

Two options are new: the files the plugin generates under `uploads/gatherpress`, and the preferences people set for themselves on their profile and the RSVP screen. Neither was removed before.

## The screen uses the Settings API

The bespoke form, save handler, nonce and separate option are gone. The reason they existed was real: `Settings::import_settings()` writes any key it finds in the field-type map, so a settings file could have armed data deletion.

That is now answered where the problem is. The Settings API takes two new flags on an option declaration:

- `'exportable' => false` keeps the value out of an export, and implies the second.
- `'importable' => false` refuses it on the way in.

An import reports what it refused separately from keys it does not recognize. Every uninstall choice is declared `exportable => false`. Documented in `docs/developer/settings/extending-settings.md`.

Deleting the bespoke path also means the screen gets the things every other settings page already had: it appears on the network settings screen, and an inherited field renders locked with "Inherited from the network" instead of silently disagreeing with the network.

## Multisite is decided by the network

`Base::run()` asks each site whether a task applies while switched to it, rather than once before the fan-out. That lets the existing inheritance machinery decide: add the `uninstall_*` keys to the inheritance list on the Network tab and one answer governs every site; leave them off and each site answers for its own data. The network settings and the shared user table belong to no site, so those follow the network's own answer.

## Correctness fixes found along the way

- The `_gatherpress_rsvp_flag` taxonomy arrived on develop after this branch opened and nothing removed it.
- `term_taxonomy.count` was left reading high for taxonomies that outlive the posts, so a category that had ten events still said ten on a site where GatherPress no longer exists.
- `posts.comment_count` was left high when RSVPs went and their events stayed.
- The post relationship delete could take an RSVP's own terms with it. A comment ID can equal a post ID and nothing in the schema tells the two apart, so it is scoped by taxonomy as well now.
- The object cache was flushed on every uninstall, including ones that removed nothing.
- `gatherpress_topic_children`, the hierarchy cache core keeps for a hierarchical taxonomy, survived because it is an option rather than a term row.

## The consequence is shown where the action is

A warning on the Uninstall screen, and a warning in the GatherPress row on the Plugins screen when anything is selected, naming what will go and linking back to the screen where it was chosen. The second one uses the same markup core uses for an unmet dependency, and answers for the network when it renders in network admin.

## Verification

Beyond the suite, the testing steps in the description were run end to end on a three site multisite:

- With nothing selected, the uninstall removed nothing. Every count matched the baseline on all three sites.
- With everything selected network wide, all three sites went to zero, while `theme_mods_gatherpress-org-theme` and a comment on a non event post survived, and a core category that had been applied to an event had its count corrected.

That run is what caught the `gatherpress_topic_children` leftover.

CodeRabbit raised five findings on the final state. All five were real and are fixed in 25da8c6a1, with a reply on each thread and its confirmation on four of them. The notable one was that a replace-mode import deleted the stored options before writing, which reset an uninstall choice the file was never allowed to make while telling the administrator it had been left alone.

## Scope

This got large. Anything anyone wants changed from here is welcome as a follow up rather than another round on this branch. Two are already known and not addressed here: where the capability constants live (#2306) and whether venue should have its own capability type (#2307).

PR #2280, opened 2026-09-07
Use Rsvp::is_comment_type() for hand-rolled RSVP comment checks

### LINKED ISSUE #2240: Use Rsvp::is_rsvp() everywhere the check is hand-rolled
Follow-up to the review on #2064.

"Is this comment an RSVP" is spelled out by hand in a dozen places, each a slightly different shape: some compare `$comment->comment_type`, some call `get_comment_type( $id )`, `Token` did both. #2064 adds `Rsvp::is_rsvp( int|WP_Comment $comment ): bool` and uses it in `Check_In` and `Token`. The remaining sites should call it too, so there is one definition of the check.

Sites on `develop` as of today:

- `includes/core/classes/blocks/class-rsvp-form.php:708`
- `includes/core/classes/blocks/class-rsvp-response.php:293`
- `includes/core/classes/rsvp/class-form.php:252`
- `includes/core/classes/rsvp/class-form.php:304`
- `includes/core/classes/rsvp/class-form.php:597`
- `includes/core/classes/rsvp/class-query.php:331`
- `includes/core/classes/rsvp/class-query.php:406`
- `includes/core/classes/rsvp/class-setup.php:506`
- `includes/core/classes/rsvp/class-setup.php:639`
- `includes/core/classes/rsvp/class-storage.php:250`

Not in scope: `class-query.php:286` and `:305` compare a query var, not a comment, and `class-setup.php:530` and `:590` compare a screen slug. Those are a different check that happens to reuse the constant.

Mechanical change, one file at a time is fine. Each swap should keep the surrounding behavior identical, including where a site only had one half of the check (existence or type) and gains the other by calling the helper. Worth a look at each to confirm that is harmless, which it should be.

faisalahammad:
Fixes #2240

## Proposed changes:

* Uses the `Rsvp::is_comment_type( int|WP_Comment $comment )` helper from #2064 in the ten hand-rolled "is this comment an RSVP" checks listed in the issue. This PR no longer adds a helper of its own.
* Fixes an infinite recursion in that helper. It called `get_comment()` even when handed a comment object, and one of its callers is a `get_comment` filter callback, so the round trip re-fired the filter and exhausted the stack on every comment read. The helper now reads the type off an object directly and only calls `get_comment()` for an int.

Site notes:

* `Rsvp_Form::process_custom_fields_for_form` and `Form::process_custom_fields` pass the comment ID to the helper, per review feedback. The two `get_comment()` call sites carry a multi-line `/** @var WP_Comment $comment */` annotation after the guard so PHPStan level 8 still knows the property reads are safe. Behavior is unchanged.
* `Rsvp_Response::modify_avatar_for_gatherpress_rsvp` (`get_avatar_data`) keeps its `instanceof WP_Comment` guard, agreed with @mauteri. That filter can pass a user ID, an email, or a `WP_User`, and the helper reads an int as a comment ID, so a user ID colliding with an RSVP comment ID would match. The typed parameter would also `TypeError` on an email or `WP_User`. `Query::mask_anonymous_rsvp_rest_author` (`rest_prepare_comment`) calls the helper directly because that filter always receives a comment object.
* `Form::handle_rsvp_comment_post` and `Setup::remove_rsvp_notification_emails` gain the existence half of the check. Harmless: a missing comment never passed the type comparison.
* Left alone, as the issue asked: query-var compares in `Query` and screen-slug compares in `Setup`. Token stays with #2064.

### Other information:

No test changes remain, since the helper and its coverage landed with #2064. `test_is_comment_type` covers the object and int paths, and it runs with the `get_comment` filter live, which is what guards the recursion from coming back. The ten swapped sites were already covered on both branches. Full suite: 2062 tests, 7831 assertions. PHPCS clean on touched files. PHPStan clean.

## Testing instructions:

This is an internal refactor. There is no UI change. Confirm RSVP still behaves the same:

1. Create an event with RSVP enabled and the RSVP Form block.
2. Submit an RSVP as a logged-in user. Confirm the RSVP is stored and the confirmation email is sent (or queued).
3. Submit an anonymous RSVP. Confirm the attendee list still masks the author for users who cannot moderate RSVPs.
4. Open the event as a visitor who cannot moderate comments. Confirm RSVP comment content is not shown on the frontend.
5. RSVP from the REST path (or the block editor attendee manager). Confirm custom fields still save.

Expected: same behavior as before this PR. No new screens, no copy changes.

Automated: `npm run lint:php`, `npm run lint:phpstan`, `npm run test:unit:php`.

### Changelog entry

Internal refactor, no user-visible change. Since the Skip Changelog label needs a maintainer and the checkbox workflow cannot push to a fork, the changelog entry file (`.github/changelog/refactor-use-rsvp-is-rsvp`) was pushed to the branch per the template's fork fallback. A maintainer can still swap it for the Skip Changelog label.

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>


faisalahammad:
Internal refactor with no user-visible change, so this PR needs the `Skip Changelog` label. I don't have write access to add it myself; could a maintainer apply it?

mauteri:
@faisalahammad note this relies on https://github.com/GatherPress/gatherpress/pull/2064 merging first.

includes/core/classes/blocks/class-rsvp-form.php
```diff
@@ -705,7 +706,7 @@ private function generate_form_id(): string {
 	 */
 	public function process_custom_fields_for_form( int $comment_id ): void {
 		$comment = get_comment( $comment_id );
-		if ( ! $comment || Rsvp::COMMENT_TYPE !== $comment->comment_type ) {
+		if ( ! $comment instanceof WP_Comment || ! Rsvp::is_rsvp( $comment ) ) {
```
mauteri: This should become a lot simpler here and every where. `$comment` is mixed, so you don't need to check it prior to passing `$comment` to the method.

```
if ( ! Rsvp::is_rsvp( $comment ) ) {
```

faisalahammad: Done. This site and the other get_comment() call site in Form::process_custom_fields now use `if ( ! Rsvp::is_rsvp( $comment ) ) {` directly.

One exception: `Rsvp_Response::modify_avatar_for_gatherpress_rsvp` keeps its `instanceof WP_Comment` guard. That filter (get_avatar_data) can also receive a user ID or an email address. is_rsvp() reads a numeric value as a comment ID, so a user ID that collides with an RSVP comment ID would match. The old code never treated those values as comment IDs, so I kept the guard there and left a comment explaining why. Full suite passes (2002 tests, 7700 assertions), PHPStan and PHPCS clean.

includes/core/classes/blocks/class-rsvp-response.php
```diff
@@ -285,12 +285,13 @@ public function attach_dropdown_interactivity( string $block_content ): string {
...
-		// doesn't have to nest under the positive guard.
+		// get_avatar_data also passes user IDs and email addresses here, and
+		// is_rsvp() would read a numeric value as a comment ID. The instanceof
+		// guard keeps those callers from ever matching an RSVP comment, so
+		// unlike the get_comment()-backed call sites this one keeps its guard.
 		if (
-			! $comment
-			|| ! is_a( $comment, 'WP_Comment' )
-			|| Rsvp::COMMENT_TYPE !== $comment->comment_type
+			! $comment instanceof WP_Comment
```
mauteri, line 293: This should change too

faisalahammad: Keeping this one, with a repro of why the guard has to stay.

get_avatar_data() also passes user IDs and email addresses here, not only comments. Without the instanceof check, is_rsvp() reads a numeric value as a comment ID. Example: user 55 gets an avatar built somewhere on the site, and RSVP comment 55 exists. The filter now matches comment 55, so user 55 gets that attendee's avatar URL everywhere the user ID path renders an avatar. The old code compared the type only on a comment object, so it never had this false positive.

A union of both checks (instanceof first, then the helper) keeps one definition of the type check and keeps the user ID path safe. If you still want this site changed, tell me and I will comply.

faisalahammad:
Summary of this round:

- Fixed the PHP Coding Standards failure: the two single-line @var annotations are now multi-line doc comments.
- Added the changelog entry file (.github/changelog/refactor-use-rsvp-is-rsvp) per the template's fork fallback, so the changelog check can pass without the Skip Changelog label. A maintainer can still swap the label in.
- Replied on the avatar guard thread and resolved it. The instanceof check stays because get_avatar_data also passes user IDs, and is_rsvp would read those as comment IDs.

Tests: PHPCS clean on touched files, PHPStan clean, PHPUnit 2002 tests and 7700 assertions pass.

@mauteri ready for re-review when you have time.

carstingaxion (dismissed):
Looks good, given that 50% will be already in after #2064 is merged.

includes/core/classes/blocks/class-rsvp-response.php
mauteri: Keep the guard here, agreed. `get_avatar_data` hands this filter `$id_or_email`, which can be a user ID, a hash, an email, a `WP_User`, a `WP_Post` or a `WP_Comment`, so an int at this site is a user ID and the helper would read it as a comment ID. The other sites in this PR are fed a comment or a comment ID, which is why they need no guard of their own.

The bigger change: #2064 merged today with the helper already in it, so this PR needs to drop the `Rsvp::is_rsvp( mixed $comment )` it adds, along with its tests, and call the merged one instead: `Rsvp::is_comment_type( int|WP_Comment $comment )`. It has a typed parameter and a `<= 0` guard for `get_comment()`'s global-comment fallback, per the review on #2064. The branch is conflicting against develop and calls `is_rsvp()` in 24 places, so please rebase onto develop and rename the call sites. The typed parameter is also the second reason this site keeps its `instanceof`: an email or a `WP_User` would otherwise be a `TypeError`.

mauteri (changes requested):

includes/core/classes/blocks/class-rsvp-response.php
faisalahammad: Done. The branch is rebased onto develop and this PR no longer adds its own helper.

What changed:

- Dropped `Rsvp::is_rsvp( mixed $comment )` and its `test_is_rsvp` test. The PR now uses the merged `Rsvp::is_comment_type( int|WP_Comment $comment )` from #2064.
- Renamed all ten call sites to `is_comment_type()`.
- The guard here stays, now with both reasons: `get_avatar_data` can pass a user ID or a `WP_User`, and the helper's typed parameter would `TypeError` on those.

One fix rode along, and it was needed to get the suite to run at all. The merged helper called `get_comment()` even when it was already handed a comment object. `Query::prepare_rsvp_comment` is a `get_comment` filter callback and calls the helper with the comment it was handed, so the round trip re-fired the filter and recursed until the stack ran out. That hit every comment read on the site, not only RSVPs, so it was not fixable at the call sites. The helper now reads the type straight off the object and only hits `get_comment()` for an int. Suite is green: 2062 tests, 7831 assertions. PHPCS and PHPStan clean.

Your `is_int()` guard in the helper became redundant after that early return, so it reads `$comment <= 0` now, which still rejects `0` and negatives.

faisalahammad:
Rebased onto develop and the review feedback is addressed. Summary of where the PR stands now:

- Dropped this PR's own `Rsvp::is_rsvp()` helper and its test, and switched the ten call sites to the merged `Rsvp::is_comment_type()` from #2064.
- Fixed a stack overflow in that merged helper: it resolved an already-passed comment object through `get_comment()`, which re-fired the `get_comment` filter that the helper is called from. Every comment read on the site was affected, so it had to be fixed in the helper rather than at the call sites.
- The avatar site keeps its `instanceof WP_Comment` guard, as agreed, now with the added reason that the typed parameter would `TypeError` on the email or `WP_User` values that filter can pass.

Gates: PHPCS clean, PHPStan clean, 2062 tests / 7831 assertions passing.

@mauteri this is ready for a re-review when you have a moment. I could not set you as a reviewer from the fork (the API returns 404 for that call), so apologies for tagging you here instead.

includes/core/classes/rsvp/class-rsvp.php
```diff
@@ -180,9 +180,17 @@ public function __construct( int $post_id ) {
 	 * @return bool True if the comment exists and has the RSVP comment type, false otherwise.
 	 */
 	public static function is_comment_type( int|WP_Comment $comment ): bool {
```
mauteri, line 182: Let's cleanup and simplify this method.

```
public static function is_comment_type( int|WP_Comment $comment ): bool {
	// Only resolve IDs: passing an object back through get_comment() would
	// re-fire the get_comment filter these callers run from and recurse.
	// get_comment( 0 ) also falls back to the global comment, so guard that.
	if ( ! $comment instanceof WP_Comment ) {
		$comment = $comment > 0 ? get_comment( $comment ) : null;
	}

	return $comment instanceof WP_Comment && self::COMMENT_TYPE === $comment->comment_type;
}
```

mauteri (changes requested):

includes/core/classes/rsvp/class-form.php
```diff
@@ -584,11 +584,16 @@ private function process_custom_fields( int $comment_id, array $data ): void {
...
+		if ( ! Rsvp::is_comment_type( $comment_id ) ) {
 			return;
 		}
 
+		$comment = get_comment( $comment_id );
+		/**
+		 * Comment resolution returns a union, so reassert the type for the reads below.
+		 *
+		 * @var WP_Comment $comment
+		 */
```
mauteri: Remove this comment block.

mauteri (changes requested):

includes/core/classes/rsvp/class-rsvp.php
faisalahammad: Done. The method is your version now, single return and no extra branches.

includes/core/classes/rsvp/class-form.php
faisalahammad: Done. The comment is fetched first and narrowed with instanceof now, same shape as in class-query.php, so the docblock is gone and the check still goes through the helper. The same docblock in Rsvp_Form::process_custom_fields_for_form is removed the same way, this PR added it there too.

faisalahammad:
Both review points are done in 973249f4.

- `Rsvp::is_comment_type()` is now the single-return version you posted.
- The `@var WP_Comment` docblock is gone from `Form::process_custom_fields()`. The comment is fetched first and narrowed with `instanceof`, the same shape already used in `class-query.php`, so the type check still runs through the helper and the second `get_comment()` call is gone too.
- The same docblock in `Rsvp_Form::process_custom_fields_for_form()` got the same treatment. This PR added it there as well, so I did not want to leave a twin of the block you asked me to remove.

Checks: `npm run lint:php` clean, `npm run lint:phpstan` clean, `npm run test:unit:php` green at 2062 tests and 7831 assertions.

@mauteri ready for another look when you have time.

mauteri (approved):

PR #2288, opened 2026-09-11
fix(ci): resolve @since TBD tags automatically

### LINKED ISSUE #2222: Write @since TBD, and resolve it automatically on develop

Every PR that adds a symbol has to guess the version it will ship in. Today AGENTS.md tells contributors to write the next planned release, which means:

- The guess is wrong whenever a PR slips a cycle, and nothing catches it.
- Long-lived branches go stale. A PR opened during `0.35.0` and merged after the bump carries `@since 0.35.0` into `0.36.0`.
- Reviewers have to know the current development line to check it, and the value looks correct either way.
- The current rule explicitly forbids the placeholder that would solve it: AGENTS.md line 218 lists `@since TBD` as bad.

The version is only knowable at release time, so writing it at authoring time is the wrong moment.

**1. `@since TBD` becomes the convention for new symbols.**

Contributors write `@since TBD`. AGENTS.md reverses the current rule, which today reads:

> ❌ Bad: `@since 1.0.0`, `@since unreleased`, `@since TBD`.

The rest of that section still applies unchanged — deriving `@since` from git history when correcting an existing symbol, the `0.27.0` floor, and signature changes not moving `@since`.

**2. A workflow on `develop` resolves them.**

Finds every `@since TBD` and replaces it with the current version, stripped of any pre-release suffix, so `0.36.0-alpha.2` resolves to `0.36.0`. It opens a PR and auto-merges it once required checks pass.

`extract-wp-hooks-as-docs.yml` is the working pattern for this: it runs on pushes to `develop`, regenerates, opens a PR with `peter-evans/create-pull-request`, and enables auto-merge with `gh pr merge --auto --squash --delete-branch`. This can follow the same shape with `paths: [ '**.php', '**.js' ]`.

**3. A PR lint that fails on a new `@since` that isn't `TBD`.**

Looks at added lines only. An `@since 0.36.0` on a line the PR adds fails; an untouched one in context does not. This is what keeps the convention from decaying back into guesses.

### Things worth deciding while implementing

- **When does it run?** On every push to `develop`, or once at version-bump time. Per-push keeps `develop` free of `TBD`, but means most merges are followed by a second automated PR. Bump-time batches them, at the cost of `TBD` sitting in `develop` between releases.
- **What resolves the version?** `package.json` is the obvious source, and matches what `.github/scripts/release/generate-version.php` already patches.
- **Alpha cycle markers.** #2166 puts `0.36.0-alpha.0` on `develop` right after a release. Stripping the suffix gives `0.36.0`, which is the right answer — worth an explicit test, since that marker exists precisely so nothing treats it as a release.
- **JS as well as PHP.** The docblock conventions in AGENTS.md apply to JSDoc too, so the scan and the lint should cover `src/` as well as `includes/`.
- **Patch branches.** A fix cherry-picked to a `version-X.Y.N` branch off `main` would want `X.Y.N`, not develop's version. Simplest answer is that the workflow only runs on `develop` and patch branches carry already-resolved values, but it should be stated.
- **Does the lint need an escape hatch?** Restoring a `@since` while correcting a wrong one is a legitimate case for adding a non-`TBD` line.

**Leave it as is.** Costs nothing to build, and stays wrong whenever a PR slips a release.

**Resolve at release time only, by hand.** No workflow, but it becomes another step in a runbook that already has several, and it is exactly the kind of step that gets skipped.

**Reviewers catch it.** They cannot, reliably. A stale `@since` is indistinguishable from a correct one without knowing when the branch was opened.

faisalahammad:
Fixes #2222

## Proposed changes:

* Makes `@since TBD` the required value for new symbols on `develop`, and adds automation that resolves it to the stable version after the change lands:
    * `.github/scripts/resolve-since.php` replaces `@since TBD` with the stable version read from `package.json` (suffixes like `-alpha.2` are stripped) across `includes/` and `src/`.
    * `.github/workflows/resolve-since.yml` runs it on every push to `develop` that touches PHP or JS source, then opens a signed PR (`Skip Changelog` label, auto-merge enabled) following the existing `extract-wp-hooks-as-docs` pattern. The workflow includes a PHP smoke test against a temporary fixture.
    * `.github/scripts/check-since.php` + `.github/workflows/check-since.yml` lint PRs targeting `develop`: an added line may only introduce `@since TBD` or replace an existing tag in the same hunk, so corrections and the resolver's own PR both pass. The workflow runs with `contents: read` and does not execute PR code, and includes a PHP smoke test for both allowed and flagged tags.
* Updates `AGENTS.md` docblock conventions: `@since TBD` for new symbols on `develop`, existing historical rules unchanged (git-history lookup, `0.27.0` floor, signature-change rule). Documents that patch branches are not processed by the workflow and carry resolved values.

### Other information:

### Testing instructions:

* Check out this branch and run `npm run lint:php` and `npm run lint:phpstan` (both pass, 0 errors, 0 warnings).
* Run the checker workflow smoke command locally from the repository root; it must accept a new `@since TBD` line and reject a new resolved tag.
* Run the resolver workflow smoke command locally against a temporary fixture; a package version `0.36.0-alpha.1` and a `@since TBD` in `includes/x.php` must resolve to `@since 0.36.0`.
* Run the PR lint locally: `GITHUB_BASE_REF=develop php .github/scripts/check-since.php` from this branch (passes, since the only added `@since` lines are TBD or test fixtures).
* Simulate the PR check failing: edit a file in `includes/` to add `@since 0.99.0`, run `git diff develop...HEAD -- includes src` and feed it through `gatherpress_since_diff_violations()` (returns one violation). Adding `@since TBD` returns none.
* Simulate the correction escape hatch: replace an existing `@since 0.36.0` line with `@since 0.37.0` in the same diff hunk; the checker allows it.
* After merge, verify on `develop`: a PR that introduces a new symbol with `@since TBD` gets resolved by a signed automated PR titled "Resolve @since tags automatically" with the `Skip Changelog` label, auto-merge on, and the tag replaced by the current stable version from `package.json`.

### Changelog entry


test/unit/php/class-test-since-scripts.php
```diff
@@ -0,0 +1,239 @@
+<?php
```
mauteri: Let's drop this test file. None of the other `.github/scripts` have tests, coverage already excludes `.github/**`, and pulling CI tooling into the plugin suite is what's turning the PHP jobs red: `@runInSeparateProcess` inside a WP test case makes PHPUnit serialize WP globals (which hold closures), so the whole run aborts with `Serialization of 'Closure' is not allowed` and exit code 2.

A smoke step in each workflow is more useful here. Something like this, after Setup PHP and before the real step:

`check-since.yml`
```yaml
    - name: Smoke test the checker
      run: |
        php -r '
        require ".github/scripts/check-since.php";
        $flagged = gatherpress_since_diff_violations( "+++ b/includes/x.php\n@@ -1 +1 @@\n+ * @since 0.36.0\n" );
        $clean   = gatherpress_since_diff_violations( "+++ b/includes/x.php\n@@ -1 +1 @@\n+ * @since TBD\n" );
        exit( ( 1 === count( $flagged ) && array() === $clean ) ? 0 : 1 );
        '
```

`resolve-since.yml`
```yaml
    - name: Smoke test the resolver
      run: |
        root="$(mktemp -d)"
        mkdir -p "$root/includes" "$root/src"
        echo '{"version":"0.36.0-alpha.1"}' > "$root/package.json"
        printf '<?php\n/**\n * @since TBD\n */\n' > "$root/includes/x.php"
        php -r 'require ".github/scripts/resolve-since.php"; exit( gatherpress_run_since_resolver( $argv[1] ) );' "$root"
        grep -q '@since 0.36.0' "$root/includes/x.php"
```

The resolver smoke fails on this branch, and that's the real bug the two `F`s in the log are pointing at. In `resolve-since.php` the replacement is `'$1' . $version`, which becomes `$10.36.0`, and PCRE reads that as group 10. So `@since TBD` resolves to `.36.0`, and the resolve workflow would open a PR with that in every docblock. `'${1}' . $version` fixes it and the smoke step goes green with that change.

mauteri (changes requested):

test/unit/php/class-test-since-scripts.php
faisalahammad: Thank you for the review. All three points are fixed in ae7037b8.

- Deleted test/unit/php/class-test-since-scripts.php. You are right, @runInSeparateProcess serializes the WP globals and kills the whole suite. No other .github/scripts have tests, so this file is removed.
- Added the smoke steps you provided to both check-since.yml and resolve-since.yml.
- Fixed the real bug in resolve-since.php: '$1' . $version was read as capture group 10, so '@since TBD' resolved to '.36.0'. Changed to '${1}' . $version.

Verified locally: both smoke steps pass, the resolver now writes @since 0.36.0, lint:php and phpstan pass.

faisalahammad:
@mauteri The review feedback is addressed in ae7037b8: deleted the PHPUnit test file, added your smoke steps to both workflows, and fixed the `${1}` capture-group bug. The thread is resolved. Could you re-review when you have a chance?

mauteri:
Thank you for this @faisalahammad, and sorry to close it after the work you put in.

The `@since TBD` idea is right and we are taking it forward. What we cannot take is resolving the tag on every push to `develop`, and the reason is specific to how we cut releases.

A fix written on `develop` does not necessarily ship from `develop`. We regularly cherry-pick one onto a `version-X.Y.N` branch off `main` for a patch release, and that release goes out before the next minor. With resolution happening on merge, the tag is stamped with `develop`'s version within minutes, so the cherry-picked copy claims a version it did not ship in. That is not hypothetical: 0.35.4 went out with docblocks saying `0.36.0` until we corrected them by hand, which is the exact chore this was meant to remove. Picking before the resolver runs is no better, since the commit then carries a literal `TBD` onto a patch branch that no workflow processes, and `TBD` ships to wp.org.

The fix is to resolve at release time instead, inside `generate-version.php`, which already patches every other version string and already runs on whichever branch is being versioned. The patch case then comes out right on its own.

That turned out to pull in a second piece of the release flow, the post-release cleanup PRs in #2311, so the two are going together in #2326. Your checker is the half that survives close to intact, and the `Skip Since Check` escape hatch came from review feedback on this thread.

You are credited in #2326 and will be in the release notes for it. For next time: #2222 was assigned, and it is worth a comment on a ticket before starting so nobody duplicates your effort or, as happened here, so we can tell you the release constraints up front.

mauteri:
Superseded by #2326. See the comment above for the release-flow reason.

Issue #1611, opened 2026-05-04
Default post type archive shows all posts regardless of upcoming/past filter on custom event-supporting post types


mauteri:
### Describe the bug

The default WordPress post type archive (no custom template) for a custom event-supporting post type (e.g. `production`) renders all posts of that type, with no honoring of the upcoming/past temporal filter that the standard event archive applies. Past productions and future productions appear together in the same list with no separation.

Spotted by @carstingaxion in https://github.com/GatherPress/gatherpress/issues/1591#issuecomment-4370868742 (screenshot included). Carsten suspects this overlaps with #1379 ("Fix inherited queries using event-query-block").

### Suspected root cause

`Event\Query::prepare_event_query_before_execution()` only swaps the temporal filter into play when the queried object matches the configured `past_events` / `upcoming_events` archive page. A bare `/production/` post type archive URL does not match that condition, so the query runs as a vanilla `WP_Query` over the post type with no datetime ordering or upcoming/past gating.

This is distinct from #1609 (which is about the Query Loop block in a custom archive template ignoring the post type narrowing) — that one is the inverse symptom (block respects "Past" filter but leaks across post types), this one is no temporal filter at all on the bare archive.

May share root cause / fix surface with [#1379](https://github.com/GatherPress/gatherpress/issues/1379), which tracks inherited queries using the event-query-block.

### Steps to reproduce

1. Open the demo Playground (seeds 5 demo productions, mix of past and future): [![Playground link](https://img.shields.io/badge/GatherPress_+_GatherPress_Productions-blue?logo=wordpress&logoColor=%23fff&labelColor=%233858e9&color=%233858e9)](https://playground.wordpress.net/?blueprint-url=https://raw.githubusercontent.com/carstingaxion/gatherpress-productions/main/.wordpress-org/blueprints/blueprint.json)
2. Visit `/production/` (the default post type archive URL).
3. **Bug:** all productions render together with no upcoming/past separation or filtering.

### Expected behavior

Custom event-supporting post types should get the same default temporal handling as the standard event archive — either an upcoming/past split, or at minimum the upcoming/past query parameters should narrow the result set when present in the URL.

### Related

- #1591 (parent thread)
- #1379 (inherited queries using event-query-block)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

PR #2290, opened 2026-09-12
fix(events): resolve event posts by post type support instead of the gatherpress_event slug

### LINKED ISSUE #2289: Event helpers hardcode gatherpress_event instead of resolving by post type support
### Describe the bug

Two helpers in `src/helpers/event.js` fetch a post by a hardcoded `gatherpress_event` post type instead of resolving the type through `gatherpress-event-date` support. Both are on `develop`; #2205 only shifts their line numbers.

- `getEventMeta()`, in the explicit-override branch: `getEntityRecord( 'postType', 'gatherpress_event', postId )` (develop line 545).
- `hasOnlineEventTerm()`: `getEntityRecord( 'postType', currentPostType || 'gatherpress_event', postId )` (develop line 445). This one is wrong twice over: it assumes the given `postId` belongs to the post type currently open in the editor, then falls back to the hardcoded slug.

A block whose `postId` override points at a companion plugin's event type, for example a `production` that declares `gatherpress-event-date`, sends the request to the events endpoint with a foreign ID. It 404s, the record is undefined, and the block quietly reports defaults: no guest limit, RSVP enabled, anonymous RSVP off, no online-event term. Nothing surfaces the failure.

### Fix

The support-based resolver already exists in the same file. `findEventPostById( selectFunc, postId )` walks every post type declaring `gatherpress-event-date` and looks the ID up in each, returning the full post with `meta`. The event-date block already uses it for this exact override case. Both call sites above should go through it rather than naming a post type.

Two things to account for while doing it:

- `findEventPostById()` returns only published posts, where `getEntityRecord()` returns anything the user can edit. An override pointing at a draft event would go from "meta read" to "defaults". That is likely the right behavior for a front-end block, but it is a change, so it should be deliberate and covered.
- The three `getEventMeta` tests mock `getEntityRecord` directly. They will need their mocks moved to `getPostTypes` and `getEntityRecords`, and a case added for a supported non-`gatherpress_event` post type so the regression cannot come back.

Per AGENTS.md, checks against `Event::POST_TYPE` or the slug literal are to be replaced with support checks. These are the two remaining in this helper.

### Steps to Reproduce

1. Register a post type with `gatherpress-event-date` support, or use one from a companion plugin.
2. Create a published event of that type with a guest limit and RSVP enabled.
3. On another post, add an RSVP block and set its post ID override to that event.
4. The block shows RSVP defaults rather than the event's settings, and the network tab shows a 404 against `/wp/v2/gatherpress_events/{id}`.

mauteri:
Closes #2289

## What

`src/helpers/event.js` fetched override posts with a hardcoded `gatherpress_event` post type in two helpers. A block whose Post ID override points at a companion plugin's event type (a `production` that declares `gatherpress-event-date`) sent the request to the events endpoint with a foreign ID, got a 404, and quietly reported RSVP defaults.

Both helpers now go through the existing support-based resolver, `findEventPostById()`:

- `getEventMeta()` resolves the override across every event-supporting post type. The live-editor branch also uses the module's `selectFunc`-based support check so the read subscribes through the caller's `useSelect`.
- `hasOnlineEventTerm()` resolves the override the same way, and derives the venue taxonomy from the post type that owns the found post instead of whatever type the editor has open.

Scanning for the same pattern turned up three PHP call sites comparing `Event::POST_TYPE` where a support check belongs:

- `Rest_Api::send_emails()` and `Rest_Api::rsvp_responses()` now gate on `Rsvp::SUPPORT`. The routes already validate `post_id` through `Validate::event_post_id()`, which is support-based, so the handlers were the only thing refusing companion post types.
- `Import::validate()` now gates on `Event::SUPPORT`, matching the export side, which marks posts through `Validate::event_post_id()`. Companion event types round-trip their datetimes through WXR.

## Behavior change

`findEventPostById()` returns only published posts, where `getEntityRecord()` returned anything the user could edit. An override pointing at a draft event now reads defaults instead of the draft's meta. Each helper has a test for it.

## Left alone, on purpose

- The archive-page branches in `Event\Setup` and `Event\Setup::get_event_archive_mode()` compare `Event::POST_TYPE` because the archive settings only exist for the standard type (#1611).
- `Event\Meta::register_event_only_meta()` registers the RSVP and online-link meta on `Event::POST_TYPE` only, and its docblock says that is deliberate. A companion type with `gatherpress-rsvp` support therefore has no `gatherpress_enable_rsvp` meta in REST unless it registers its own. Flagging rather than changing here.

## Tests

- JS: the `getEventMeta` override tests moved from `getEntityRecord` mocks to `getPostTypes` and `getEntityRecords`, plus a companion-type case and a draft-override case. The `hasOnlineEventTerm` override tests likewise, plus a case asserting the taxonomy comes from the override's post type (`_location`) and an unpublished-override case. `src/helpers/event.js` sits at 100% statements, branches, functions and lines. Full JS suite passes.
- PHP: `test_rsvp_responses_custom_rsvp_post_type`, `test_rsvp_responses_event_date_only_post_type`, `test_send_emails_custom_rsvp_post_type`, and `test_validate` extended with a supported custom type and a plain `post`. `Test_Rest_Api` and `Test_Import` pass. phpcs, PHPStan, and PHPMD are clean.

## Summary by CodeRabbit

- **New Features**
  - Event imports now support compatible custom content types that provide event-date support.
  - RSVP responses and event email notifications now work with compatible custom content types that provide RSVP support.
  - Event venue and metadata overrides now resolve correctly across supported custom event types.

- **Bug Fixes**
  - Improved handling of unpublished, missing, or incomplete event overrides.
  - Online event taxonomy and metadata lookups now use the correct content type.
  - Improved security for nonce requests and RSVP verification.

### DIFF
--- a/.github/changelog/2289-event-helpers-post-type-support
+++ b/.github/changelog/2289-event-helpers-post-type-support
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Resolve event posts by post type support instead of the hardcoded `gatherpress_event` slug in the editor event helpers, the RSVP REST handlers, and the WXR importer, so blocks and endpoints pointed at a companion plugin's event post type read its settings instead of quietly falling back to defaults.
--- a/includes/core/classes/class-import.php
+++ b/includes/core/classes/class-import.php
@@ -97,15 +97,21 @@ final class Import extends Migrate {
 	}
 
 	/**
-	 * Checks if the currently imported post is of type 'gatherpress_event'.
+	 * Checks if the currently imported post belongs to an event-supporting post type.
+	 *
+	 * Mirrors the export side, which marks posts through `Validate::event_post_id()`,
+	 * so a companion plugin's event post type round-trips its datetimes too.
 	 *
 	 * @param  array<string, mixed> $post_data_raw The result of 'wp_import_post_data_raw'.
 	 *
-	 * @return bool                                True, when the currently imported post is of type
-	 *                                             'gatherpress_event', false otherwise.
+	 * @return bool                                True, when the currently imported post's type declares
+	 *                                             `gatherpress-event-date` support, false otherwise.
 	 */
 	protected function validate( array $post_data_raw ): bool {
-		return ( isset( $post_data_raw['post_type'] ) && Event::POST_TYPE === $post_data_raw['post_type'] );
+		return (
+			isset( $post_data_raw['post_type'] ) &&
+			post_type_supports( (string) $post_data_raw['post_type'], Event::SUPPORT )
+		);
 	}
 
 	/**
--- a/includes/core/classes/event/class-rest-api.php
+++ b/includes/core/classes/event/class-rest-api.php
@@ -482,9 +482,9 @@ final class Rest_Api {
 	/**
 	 * Send emails to selected members.
 	 *
-	 * This method is responsible for sending emails to specific members. It checks if the given
-	 * `$post_id` corresponds to a specific post type, retrieves the list of members to email, and sends the email with
-	 * the appropriate subject, body, and headers.
+	 * This method is responsible for sending emails to specific members. It checks that the given
+	 * `$post_id` belongs to a post type with RSVP support, retrieves the list of members to email, and sends the
+	 * email with the appropriate subject, body, and headers.
 	 *
 	 * @since 0.34.0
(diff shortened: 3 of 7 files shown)


carstingaxion (dismissed):

mauteri:
@coderabbitai review

PR #2292, opened 2026-09-12
feat(rsvp): add no-JS RSVP fallback

### LINKED ISSUE #2039: RSVP trigger is inert when JavaScript is unavailable

The RSVP block renders fine without JavaScript. Its markup is produced server-side by `Blocks\Rsvp::transform_block_content()`, and the Interactivity API is layered on afterward. What doesn't survive is the *action*: the trigger opens a modal, and the modal needs JS. So with JS unavailable the user sees a normal-looking RSVP button that does nothing when clicked.

A dead control is a worse failure than an absent one, because there's no signal that anything went wrong.

This is worth framing as resilience rather than progressive enhancement. Deliberately disabling JS is rare; JS failing to arrive is less rare and less predictable (CSP, a blocked or slow CDN, an extension conflict, an uncaught error earlier on the page). See #1121 for the longer discussion.

### Designs

Give the RSVP trigger a no-JS path that covers the single most common action and nothing else.

**Scope, deliberately narrow:**

- **Logged-in users only.** Anonymous RSVP needs an email field, and once you're collecting input you've rebuilt the modal. A logged-in toggle needs no inputs at all.
- **No guests, no anonymous option, no custom fields.** Those stay modal-only.

**Mechanism:**

1. A marker class flags which trigger opts into the no-JS treatment, the same way `gatherpress-rsvp--trigger-update` already flags the interactivity target for `apply_rsvp_button_interactivity()`. The class is a flag for the server, not the mechanism itself.
2. For a flagged trigger, emit a `<noscript>` containing its own `<form method="post">` with a real submit control. `<noscript>` renders only when JS is off, so this cannot affect the JS path, and it avoids rewriting the existing button. That matters because the trigger may be an `<a>` (the core Button block renders one), which a form can't submit, and `WP_HTML_Tag_Processor` can't change tag names or wrap elements.

**It must be POST.** A link with a nonce in the href is tempting and one line of work, but prefetchers, email scanners, and browser preloading will fire it and RSVP people who never clicked.

### Things to decide before building

**Toggle semantics.** RSVP is not binary. The states are no-status, attending, not-attending, and waiting-list.

- no-status → attending, and attending → not-attending, are both obvious.
- not-attending → attending is presumably fine.
- waiting-list → ? Needs a decision.

**Full events.** When the event is at max attendance, choosing "attending" lands the user on the waiting list instead. They have to be told, which means the post-submit redirect carries a status message. `Rsvp\Form::handle_rsvp_comment_redirect()` is the existing precedent for that pattern.

**Where to POST.** Three candidates:

1. `admin-post.php` with a registered action. Most boring, most reliable.
2. The endpoint infrastructure from #955, which would give real URLs like `/events/my-event/rsvp`. This is @carstingaxion's own suggestion from the #1121 thread.
3. The comment submission flow that `Rsvp\Form` already rides (`preprocess_comment` / `comment_post`), for consistency with how the RSVP Form block posts without JS today.

**Nonce and page caching.** A nonce baked into server-rendered HTML goes stale behind a full-page cache. This is logged-in-only and logged-in users typically bypass page caching, so it's likely a non-issue, but it should be a conscious decision rather than something discovered in a bug report later.

Leaving it. The audience is small and nobody expects status switching, guest counts, and waiting lists to work without JS. The counterargument is that a button which silently does nothing is a bad enough failure mode to be worth a bounded fix, and this one is bounded.

faisalahammad:
Fixes #2039

## Proposed changes:

* Adds a no-JavaScript fallback for the RSVP trigger. Logged-in users get a `<noscript>` form with a real submit button, rendered at the RSVP block wrapper level so it stays visible even though the status triggers live inside modals that are hidden without JS.
* The form posts to `admin-post.php` through a new handler in `Rsvp\Form`. It verifies the nonce, requires a logged-in user, whitelists the status to attending/not_attending, and checks event visibility, RSVP enabled state, and past events before saving through `Rsvp::save()`.
* The redirect back to the event reports the outcome: waiting-list placement when a full event converts the RSVP, a generic success message otherwise, and an error message when the save fails.
* Triggers opt in via a `gatherpress-rsvp--trigger-no-script` marker class, added to the four default RSVP templates. Existing saved blocks keep working unchanged since unmarked content is left alone.
* Toggle semantics mirror the JS path: no status or not attending submits attending; attending and waiting list submit not attending.

### Other information:

## Testing instructions:

* Build the plugin or run the development server, and activate GatherPress.
* Log in and open an upcoming event page.
* Disable JavaScript in your browser (or use a browser profile with JS blocked).
* Result: the RSVP block shows an Attend or Not Attending button that works without JS. Submitting it reloads the page and shows a status message confirming the RSVP.
* Submit again to toggle the RSVP back, and check that the button and message reflect the change.
* Set the event's max attendance limit low, fill it with another RSVP, then submit Attend without JS. Result: the message says you are on the waiting list.
* Re-enable JavaScript and confirm the normal modal RSVP flow still works.
* Run `npm run test:unit:php` to see the new unit tests pass.

### Changelog entry

User-visible feature. The changelog entry file (`.github/changelog/2039-rsvp-no-js-fallback`) was pushed to the branch per the template's fork fallback, since the checkbox workflow cannot push to a fork.

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Add a no-JavaScript RSVP fallback form for logged-in users so the RSVP trigger still works when JavaScript fails to load.

</details>

### DIFF
--- a/.github/changelog/2039-rsvp-no-js-fallback
+++ b/.github/changelog/2039-rsvp-no-js-fallback
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Add a no-JavaScript RSVP fallback form for logged-in users so the RSVP trigger still works when JavaScript fails to load.
--- a/includes/core/classes/blocks/class-rsvp.php
+++ b/includes/core/classes/blocks/class-rsvp.php
@@ -16,6 +16,8 @@ use GatherPress\Core\Blocks\Form_Field;
 use GatherPress\Core\Blocks\General_Block;
 use GatherPress\Core\Event;
 use GatherPress\Core\Rsvp as Core_Rsvp;
+use GatherPress\Core\Rsvp\Form;
+use GatherPress\Core\Rsvp\Response\Status;
 use GatherPress\Core\Rsvp\Setup as Rsvp_Setup;
 use GatherPress\Core\Traits\Singleton;
 use GatherPress\Core\Utility;
@@ -41,6 +43,15 @@ final class Rsvp {
 	 * @var string
 	 */
 	const BLOCK_NAME = 'gatherpress/rsvp';
+
+	/**
+	 * Constant representing the class that flags a trigger for the no-JS RSVP form.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @var string
+	 */
+	const NO_SCRIPT_TRIGGER_CLASS = 'gatherpress-rsvp--trigger-no-script';
 
 	/**
 	 * Class constructor.
@@ -69,6 +80,7 @@ final class Rsvp {
 		add_filter( $render_block_hook, array( $this, 'apply_rsvp_button_interactivity' ) );
 		// Priority 11 ensures this runs after transform_block_content which modifies the block structure.
 		add_filter( $render_block_hook, array( $this, 'apply_guest_count_watch' ), 11 );
+		add_filter( $render_block_hook, array( $this, 'render_no_script_fallback' ), 12, 2 );
 		// Priority 9 ensures this runs before transform_block_content to properly register form field hooks.
 		add_filter( $render_block_hook, array( $this, 'apply_guests_input_interactivity' ), 9 );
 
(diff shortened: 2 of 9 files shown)


mauteri:
@faisalahammad be sure to be mindful of milestones. This issue is slated for `0.37.0` not `0.36.0`.

PR #2301, opened 2026-09-14
feat(blocks): add Subscribe to Events block

### LINKED ISSUE #1686: Add FSE/editor block for ICS subscribe links

The calendar subsystem (PR #955 and follow-ups) ships subscribable ICS endpoints at predictable URLs — `/event/<slug>/ical`, `/feed/ical`, `/event/feed/ical`, `/venue/<slug>/feed/ical`, `/topic/<slug>/feed/ical`. They're reachable via direct URL or via the existing `add-to-calendar` block on event singulars, but site editors building landing pages, sidebars, or sitewide subscribe surfaces have no first-class block to insert these links.

A "Subscribe to events" block (or family of blocks / variations) that:

- Lets the editor pick the feed scope: sitewide, per-post-type archive, per-venue, per-topic / per-term.
- Renders an `<a>` (or family of `<a>`s — iCal, Google Calendar, Outlook, etc.) pointing at the resolved endpoint URL.
- For per-term scopes, exposes the term selector inline.
- Optionally exposes the `webcal://` variant for one-click subscription on supporting platforms.

Two paths to get there (per Carsten in #603):

1. Pull [carstingaxion/gatherpress-feed-buttons](https://github.com/carstingaxion/gatherpress-feed-buttons) into the GatherPress umbrella — re-namespace, port to GatherPress conventions (AGENTS.md docblock format, hook naming), add tests.
2. Adopt the upstream pattern from [WordPress/gutenberg#65351](https://github.com/WordPress/gutenberg/issues/65351) once that lands.

Path 1 is the faster shipping option; path 2 leans on core to do the right thing eventually.

- Document the URL scheme and let editors hand-author `<a href>` links. Works for technical site builders, doesn't scale to less-technical editors.
- Build a per-block solution for each scope (one block for sitewide, one for venue, one for topic). Fewer attributes per block, more blocks to maintain. Variations on a single block is the more consistent shape.

### Context

Surfaced in https://github.com/GatherPress/gatherpress/issues/603#issuecomment-4377157417 (Carsten) and acknowledged as a follow-up in https://github.com/GatherPress/gatherpress/issues/603#issuecomment-4364756998 (mauteri).

### Code of Conduct

faisalahammad:
Fixes #1686

## Proposed changes:

* Adds a Subscribe to Events block. Editors pick a scope (all events on the site, an events archive, one venue, or one topic) and the block renders an iCal link, a subscribe (webcal) link, or both, so visitors can subscribe instead of downloading a single event.
* Adds `Calendar\Feed_Url`, a static URL resolver with the four feed scopes, plus a new `gatherpress_calendar_feed_url` filter so companion plugins can supply feed shapes core does not know about.
* Refactors `Calendar\Setup::get_calendar_url_for_request()` so its four feed scopes delegate to the new resolver. The single-event branch stays in `Setup`. One source of truth, and the existing `test_get_calendar_url_for_request*` tests still pass untouched.
* Adds a `TopicSelect` component for the topic scope, following the existing `EventSelect` / `useVenueOptions` pattern.
* Adds a user docs page for the block and cross-references it from the calendar feeds and features pages.

### Other information:

* PHP: `class-test-feed-url.php` covers every scope plus the invalid-scope, unknown-post-type, missing-post, missing-term, unpublished-venue, and non-event-post-type branches, the filter override on both known and unknown scopes, and `to_webcal()` across `http`, `https`, and already-`webcal` input. `class-test-subscribe-to-events.php` renders through `do_blocks()` for each scope and each link format, plus the early-return branches.
* JS: `edit.test.js` and `TopicSelect.test.js` cover scope switching, the conditional pickers, the link format and label controls, and the store query shape. Both suites are at full statement, branch, function, and line coverage.

## Testing instructions:

* Build the plugin or run the development server, and activate GatherPress.
* Create an event, a venue, and a topic term, and assign the venue and topic to the event.
* Edit a page and insert the Subscribe to Events block.
* Leave the scope on "All events on the site" and confirm the preview shows iCal and Subscribe links. Click the iCal link and confirm it returns a calendar file.
* Switch the scope to "An events archive" and confirm the archive picker appears with the configured event post types.
* Switch the scope to "Events at one venue", pick your venue, and confirm the preview links resolve to that venue's feed.
* Switch the scope to "Events in one topic", pick your topic, and confirm the preview links resolve to that topic's feed.
* Open the Links panel and switch between iCal and subscribe links, then set both label fields. Result: the preview shows only the chosen links, with the labels you typed.
* Set the venue scope to a venue that is still a draft. Result: the block renders nothing on the front end.
* Add a `gatherpress_calendar_feed_url` filter that returns a custom URL and pass an unlisted scope. Result: the filter receives the call and the returned URL is used.
* Run `npm run test:unit:php` and `npm run test:unit:js` and confirm the new tests pass.

### Changelog entry

User-visible feature. The changelog entry file (`.github/changelog/1686-subscribe-to-events-block`) was pushed to the branch per the template's fork fallback, since the checkbox workflow cannot push to a fork.

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Added a Subscribe to Events block. Editors can place it to render an iCal or webcal link for the whole site, the events archive, one venue, or one topic, so visitors can subscribe instead of downloading a single event.

</details>

## Summary by CodeRabbit

* **New Features**
  * Added a Subscribe to Events block for live iCal and Webcal subscriptions.
  * Configure feeds for the whole site, event archives, venues, or topics.
  * Customize link formats and labels, with searchable venue and topic selection.

* **Documentation**
  * Added user guidance for the Subscribe to Events block.
  * Clarified the difference between live subscriptions and one-time calendar downloads.
  * Documented calendar feed options and usage across supported scopes.


src/blocks/subscribe-to-events/edit.js
```diff
@@ -0,0 +1,244 @@
...
+		value: 'venue',
+	},
+	{
+		label: __( 'Events in one topic', 'gatherpress' ),
+		value: 'topic',
+	},
+];
+
+/**
+ * The link flavours the block can render.
```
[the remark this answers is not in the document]
faisalahammad: Fixed. The comment now reads "flavors".

src/blocks/subscribe-to-events/render.php
```diff
@@ -0,0 +1,77 @@
...
+use GatherPress\Core\Calendar\Feed_Url;
+
+if ( ! isset( $attributes ) || ! is_array( $attributes ) ) {
+	return;
+}
+
+// An unknown scope would otherwise fall through to the sitewide feed.
+$gatherpress_scope = $attributes['scope'] ?? '';
+if ( ! in_array( $gatherpress_scope, Feed_Url::SCOPES, true ) ) {
+	return;
```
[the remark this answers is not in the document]
faisalahammad: Fixed. The guard and its comment are gone, so the filter now gets called for any scope. The renderer still bails on an empty or false URL, so an unknown scope with no filter renders nothing. Added `test_render_honors_the_feed_url_filter_for_unknown_scope` to cover the filter path with a scope core does not know. It fails against the old code and passes now.

test/unit/php/includes/tests/core/classes/blocks/class-test-subscribe-to-events.php
```diff
@@ -0,0 +1,246 @@
...
+			'An unknown scope should render no wrapper.'
+		);
+	}
+
+	/**
+	 * Coverage for the filter reaching the block output.
+	 *
+	 * @return void
+	 */
+	public function test_render_honours_the_feed_url_filter(): void {
```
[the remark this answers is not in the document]
faisalahammad: Fixed. Renamed the test method to `test_render_honors_the_feed_url_filter` and changed "flavours" to "flavors" in the docs. The typos command from CI exits clean now.

faisalahammad:
Pushed 39a06c1b with the review fixes.

- `render.php`: dropped the scope guard, so `gatherpress_calendar_feed_url` is now called for any scope. An unknown scope with no filter still renders nothing, since `Feed_Url::get()` returns `false`.
- Added `test_render_honors_the_feed_url_filter_for_unknown_scope`. It fails against the old guard and passes now.
- US spelling in the `edit.js` comment, the block docs, and the filter test method name. The typos command from CI exits clean.

All three threads are resolved. PHP suite is green (2091 tests) and the local CodeRabbit pass reports no findings. Ready for another look.

src/components/TopicSelect.js
```diff
@@ -0,0 +1,156 @@
...
+export function useTopicOptions( search, topicId ) {
+	const { topics, selected, isResolving } = useSelect(
+		( wpSelect ) => {
+			const { getEntityRecord, getEntityRecords, isResolving: resolving } =
+				wpSelect( coreStore );
+			const query = {
+				context: 'view',
+				per_page: PER_PAGE,
+				search,
+				orderby: search ? 'relevance' : 'name',
```
mattcowan, line 86: `relevance` is not in the REST terms `orderby` enum (`id, include, name, slug, include_slugs, term_group, description, count`), so a search returns HTTP 400 and the picker shows no matches while the user types. Use `name` for both cases. The test at `TopicSelect.test.js` line 154 asserts `relevance` and needs the same change.

```suggestion
				orderby: 'name',
```

includes/core/classes/calendar/class-feed-url.php
```diff
@@ -0,0 +1,291 @@
...
+	 * to resolve to a published tax-like shadow source.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @param int $venue_id Venue post ID.
+	 *
+	 * @return string|false Feed URL, or false when the ID is not a venue.
+	 */
+	protected static function venue_url( int $venue_id ): string|false {
+		$venue = get_post( $venue_id );
```
mattcowan, line 240: `get_post( 0 )` returns the global post, so a venue scope with no venue picked renders the current venue's feed on a venue page or inside a venue Query Loop. The docs (line 23 of the user page) say it renders nothing. Return false for a non-positive ID before the lookup.

```suggestion
		$venue = 0 < $venue_id ? get_post( $venue_id ) : null;
```

src/blocks/subscribe-to-events/edit.js
```diff
@@ -0,0 +1,244 @@
...
+						/>
+					</VStack>
+				</PanelBody>
+			</InspectorControls>
+
+			<div { ...blockProps }>
+				<ServerSideRender
+					block="gatherpress/subscribe-to-events"
+					attributes={ attributes }
+				/>
```
mattcowan, line 240: Add `skipBlockSupportAttributes`. `useBlockProps()` already applies the padding, color, and font-size supports to the wrapper, and the server applies them again to the `<ul>`, so the editor preview shows them twice and the front end once.

```suggestion
				<ServerSideRender
					block="gatherpress/subscribe-to-events"
					attributes={ attributes }
					skipBlockSupportAttributes
				/>
```

includes/core/classes/calendar/class-feed-url.php
```diff
@@ -0,0 +1,291 @@
...
+ *
+ * The calendar subsystem serves subscribable iCalendar feeds at a fixed set of
+ * URL shapes: `/feed/ical`, `/event/feed/ical`, `/venue/<slug>/feed/ical`, and
+ * `/topic/<slug>/feed/ical`. Those URLs are built by WordPress core's feed-link
+ * functions, which each take a different object — a post type, a post ID, or a
+ * term — so resolving one from a scope plus an identifier needs the dispatch
+ * this file provides.
+ *
+ * The block that renders subscribe links and the `<link rel="alternate">` tags
+ * in `<head>` both need that dispatch, which is why it lives here rather than
```
mattcowan, line 13: Is it intended that the `<head>` alternate links do not go through this resolver? `Calendar\Setup` still calls the WP feed-link functions directly at lines 381, 413, 537, 560, and 650, so a `gatherpress_calendar_feed_url` filter changes the block and the `Accept: text/calendar` redirect but not auto-discovery. I think either route those through `Feed_Url::get()` or trim this paragraph and the matching one in `render.php` line 7.

mattcowan (commented):
Reviewed at `39a06c1b`. PHPUnit (2091) passes, and I rendered each scope and both link formats at runtime. The resolver and the `Setup` refactor look clean. Four inline comments, three are code that should be addressed and one is a docblock wording question.

src/components/TopicSelect.js
faisalahammad: Fixed. `orderby` is now `name` in both cases. You are right, `relevance` is not in the terms enum and searching gave a 400. The code no longer branches on the search term, so both paths send `name`. Updated the test in `TopicSelect.test.js` to assert `name` for the search and the non-search case.

includes/core/classes/calendar/class-feed-url.php
faisalahammad: Fixed. `venue_url()` now checks the ID before the lookup, so a non-positive ID returns false and no longer picks up the global post. Added a test that puts a venue in the global post context and passes venue ID 0, it asserts false. The topic path already failed safe because `get_term( 0 )` returns a `WP_Error`.

src/blocks/subscribe-to-events/edit.js
faisalahammad: Fixed. Added `skipBlockSupportAttributes` to `ServerSideRender`. The supports now come from `useBlockProps()` on the wrapper only, so the preview and the front end match. Updated the mock in `edit.test.js` to pass the prop through and added an assertion for it.

includes/core/classes/calendar/class-feed-url.php
faisalahammad: Good catch. The `<head>` links are built by `Calendar\Setup` with the core feed-link functions, so they do not go through this resolver. I trimmed the claims instead of routing `Setup` through it, since that is a bigger change and does not belong in this PR.

The class docblock now says the block needs the dispatch, and that the alternate tags in `<head>` are built separately and do not pass through the resolver. Dropped the matching claim in `render.php` line 7. No behavior change.

faisalahammad:
All four review points are fixed and the threads are resolved.

- `orderby` is `name` for topics now, the terms endpoint does not accept `relevance`.
- `venue_url()` guards the ID first, so a venue scope with no venue picked renders nothing instead of the global post's feed.
- `ServerSideRender` now uses `skipBlockSupportAttributes`, so the supports are not applied twice in the editor.
- Trimmed the docblock claims. The `<head>` alternate links are still built by `Calendar\Setup` and do not go through the resolver.

Tests added for the venue guard and the SSR prop. All headless checks pass.

@mattcowan ready for another look when you have time.

mattcowan (dismissed):
Re-reviewed at `d4e8b2f2`. All 4 of my orgiinal findings are fixed so I'll approve, but Github shows a merge conflict. @faisalahammad can you pull develop into this branch to resolve that? It would be good to have that cleared before anyone else looks. 

Additionally I think you'll want to look at `package-lock.json`, it has ~60 lines of npm-version noise around the one real change, `@wordpress/server-side-render`. If you restore develop's copy of that file and run `npm install` on the `.nvmrc` Node (24), the diff comes down to just that dependency, and it should stop flipping back and forth between contributors on different npm versions.

faisalahammad:
Updated the branch after review.

- Merged the current upstream `develop` branch and resolved the documentation conflicts.
- Restored `package-lock.json` from `develop` and kept only the `@wordpress/server-side-render` dependency change.
- Fixed the requested feed resolver, spelling, topic/venue ID normalization, and block preview support handling.
- All review threads are replied to and resolved.
- Local checks passed before the merge update: focused JS tests 28/28 with 100% coverage, focused PHP tests 31/31, PHPCS, CSS, Markdown, package JSON lint, production build, plugin zip, and scoped CodeRabbit review.
- Manual zip testing was confirmed before commit and push.

GitHub Actions checks for the new head are queued now.

PR #2305, opened 2026-09-15
Tighten origin, signature and visibility checks

mauteri:
## Proposed changes:

* The `/event/nonce` route answers requests from the site's own origin (no `Origin` header, or one of the site's allowed origins) and sends no CORS headers. `Utility::ensure_user_authentication()` applies the same check before restoring the cookie user, via a new `Utility::is_same_origin_request()`.
* `Rsvp_Template::sign_template()` and `verify_template()` take the event ID, and the signature covers the block name, the site and the event along with the template. The status endpoint verifies against the `post_id` it was asked to render.
* The venue block resolves its source post through one check, `get_viewable_source_post()`, whether the post comes from `selectedPostId`, a `postId` override, or the event's venue term. Looking up an event's venue also requires that the viewer could open the event.
* The RSVP count block renders for post types that support RSVPs and only for events the viewer could open, matching the other RSVP blocks.
* An event's online link is carried in `meta.gatherpress_online_event_link` only for a viewer who could edit the event. The key is registered so the editor can write it, and it was read back with the post; the `meta.online_event_link` key the block reads is unchanged and still answers on the event's terms.

### Other information:

Pages cached before this change carry the previous template signature, so the RSVP list's live refresh on those pages is refused until the page cache is purged. The server-rendered list still shows.

## Testing instructions:

* Log in, open a published event that has a few RSVPs, and change your response. The RSVP buttons work and the response list updates.
* Assign a venue to an event and view it. The venue block renders. Open the venue itself and its venue block renders too.
* Add an RSVP count block to a published event. It shows the count.
* As a logged-out visitor, request an event through `wp/v2/gatherpress_events`. The online link is empty in both meta keys. RSVP as attending and the block still shows the link.
* Run the PHP suites. New cases cover requests from another origin, a signature used for a different event or site, and venue and count blocks pointed at draft, private, trashed, or password-protected posts.

### Changelog entry

Added manually at `.github/changelog/rest-context-checks`.

## Summary by CodeRabbit

- **Security**
  - Event template signatures are now tied to their specific event and site, preventing reuse elsewhere.
  - Nonce requests are restricted to same-origin requests, with cross-origin access denied.
  - Online event links are hidden from viewers without editing permissions.

- **Bug Fixes**
  - Venue and RSVP Count blocks now consistently respect event visibility and content access rules.
  - REST event responses now provide accurate online-event links for authorized editors.

- **Tests**
  - Added coverage for signature isolation, visibility rules, origin validation, and protected event links.


includes/core/classes/class-utility.php
```diff
@@ -789,6 +796,23 @@ public static function ensure_user_authentication(): int|false {
...
+	 * site's own. Only the `allowed_http_origins` filter widens the list.
+	 *
+	 * @since 0.35.4
+	 *
+	 * @return bool True when the request has no origin or an allowed one.
+	 */
+	public static function is_same_origin_request(): bool {
+		$origin = get_http_origin();
+
+		return '' === $origin || '' !== is_allowed_http_origin( $origin );
```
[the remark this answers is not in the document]
mauteri: @coderabbitai Fixed in 0eb1c9109. `Utility::is_same_origin_request()` no longer uses `is_allowed_http_origin()`. It normalizes the request origin with a new private `get_url_origin()`: lowercase scheme and host, port kept only when it is not the scheme's default. It then compares that exactly against the origins of `home_url()`, `site_url()` and `admin_url()`. An empty origin is still treated as same-origin, and `ensure_user_authentication()` and the nonce route both go through this one predicate.

New tests in `class-test-utility.php`:
- `test_is_same_origin_request_on_a_non_default_port` and `test_ensure_user_authentication_on_a_non_default_port`: a site at `https://example.test:8443` accepts that origin and refuses `https://example.test` and `https://example.test:9443`.
- `test_is_same_origin_request`: now also checks case-insensitive matching and refuses the same host on the other scheme.
- `test_get_url_origin_*`: direct tests of default-port removal, non-default ports, and values with no host.

Can you confirm this resolves the finding?

includes/core/classes/blocks/class-venue.php
```diff
@@ -148,49 +148,59 @@ private function get_source_post( array $block ): ?WP_Post {
...
+	 * @return WP_Post|null The post when the viewer may see it, otherwise null.
+	 */
+	private function get_viewable_source_post( ?WP_Post $candidate, string $source_post_type ): ?WP_Post {
+		$viewable = $candidate instanceof WP_Post
+			&& $source_post_type === $candidate->post_type
+			&& post_type_supports( $source_post_type, Shadow_Source::SUPPORT )
+			&& ! post_password_required( $candidate )
+			&& (
+				is_post_publicly_viewable( $candidate )
+				|| current_user_can( Event::READ_CAPABILITY, $candidate->ID )
```
carstingaxion, line 201: This is the one I meant on slack.

Shouldn't this READ_CAPABILITY not come from the post type object of the candidate?

mauteri: It does, it's just not obvious at that line. `READ_CAPABILITY` is `read_post`, which is a meta capability, so `current_user_can()` sends it through `map_meta_cap()`, and that resolves it against the candidate's own post type object: `get_post_type_object( $post->post_type )` at `capabilities.php:322`, then `cap->read` for a public status (`:370`), `cap->read_private_posts` for private (`:377`), and for draft, pending or scheduled it falls through to `map_meta_cap( 'edit_post', ... )` for that same type (`:379`). A post type registered with `map_meta_cap => false` gets its own `read_post` string instead (`:341`).

So a companion venue type with custom capabilities is checked with its own, and the edit check you're asking about is already there for the non-public statuses.

What we shouldn't do is read the cap off the post type object directly here, like `->cap->read`. That's the primitive capability and it drops the per-post part: status, authorship, password. The meta cap is the one that accounts for those.

Where you're right is readability. `Event::READ_CAPABILITY` while checking a venue is weird. The constant predates this PR and the selected-post branch used it the same way, so rather than churn this one, it should move somewhere neutral. Opened #2306 for it.

PR #2322, opened 2026-09-20
Expose venue geo data on events via lazy Geodata-standard sync

### LINKED ISSUE #1478: Expose venue geo data on related events
## Overview

Now that #1471 exposes the [WordPress Geodata standard](https://codex.wordpress.org/Geodata) meta (`geo_latitude`, `geo_longitude`, `geo_address`, `geo_public`) on `gatherpress_venue`, we should decide how to surface that same geo data on the events associated with a venue so Geodata-aware plugins (e.g. [Simple Location](https://wordpress.org/plugins/simple-location/)) can consume it at the event level too.

Opening this as a follow-up to [#1471 (comment)](https://github.com/GatherPress/gatherpress/pull/1471#issuecomment-4269608429).

## Approaches to consider

A few directions worth evaluating — not mutually exclusive:

1. **Mirror venue geo meta onto the event post.** On venue selection (or on event save), copy `geo_latitude`, `geo_longitude`, `geo_address`, `geo_public` from the selected venue to the event as read-only meta, following the same `auth_callback => '__return_false'` + `rest_pre_insert_{post_type}` filter pattern used for the venue keys and read-only datetime meta in `Event_Setup`. Simple, no schema changes, Geodata-aware plugins work out of the box.
2. **Custom table for geo data.** Introduce a dedicated table keyed by `post_id` + `post_type` (and `site_id` for multisite) to centralize geo storage across event and venue post types. Heavier lift, but scales better for queries like "events within X miles" and avoids duplicating meta.
3. **Hybrid.** Keep the venue post meta as the source of truth, derive event meta on save (option 1), and add a custom table later if/when we need spatial querying.

## Open questions

- Should `geo_public` on the event reflect the event's post status, the venue's, or both (e.g. `1` only if both are published)?
- How should we handle venue changes after an event is published — resync on venue save via `wp_after_insert_post`?
- Do we want to support multiple venues per event down the road, and if so does that push us toward option 2?
- Should this land before or after #1263 (events API filters incl. geolocation)?

## Related

- #1471 (venue Geodata standard — merged)
- #1263 (events API filters incl. geolocation)
- #1264 (richer schema.org PostalAddress + OSM autofill)

### Designs

_No response_

_No response_

### Code of Conduct

jmarx:
Fixes #1478

## Proposed changes:

* Registers `geo_latitude` / `geo_longitude` / `geo_address` / `geo_public` — the WordPress Geodata micro-standard, deliberately unprefixed so third-party plugins (e.g. Simple Location) recognize them without any GatherPress-specific integration — as read-only meta on both venues and events.
* Venue side derives its own `geo_*` from its own `gatherpress_latitude` / `gatherpress_longitude` / `gatherpress_address` fields on save.
* Event side is lazy rather than eager: `geo_*` is refreshed against the linked venue's current data only when that specific event is actually requested (a front-end single view, or a REST read) — never pushed to every linked event when the venue saves. A venue can plausibly have ~1,000 linked events, and eager fan-out on every venue save doesn't scale, regardless of whether anyone ever views those events again.
* Skipped entirely for events that have already happened — a past event's recorded location is a historical fact, not something that should retroactively change because the venue's address was later corrected or edited.

### Other information:

## Testing instructions:

This was verified against a real, unmodified install of [Simple Location](https://wordpress.org/plugins/simple-location/) — not just unit tests.

1. Install and activate Simple Location. No configuration needed — no map provider, no API key.
2. Create a venue with a real address, publish it.
3. Create an event, assign it to that venue, publish it.
4. Check the event's REST response (`/wp-json/wp/v2/gatherpress_event/<ID>`) — `meta.geo_latitude` etc. should still be empty at this point. This confirms nothing writes it on save.
5. Load the event's real front-end permalink once.
6. Re-check the REST response (or `wp post meta list <ID>`) — `geo_*` now matches the venue. View source on the event page — a `.sloc-display` block should be present, rendered entirely by Simple Location's own code, showing the address.
7. Edit the venue's address and save it. Without touching the event, reload its front-end page — the address updates, both in meta and in what Simple Location renders. This is the actual scaling-friendly behavior this PR is about: no batch update ever runs, only the one event actually being viewed gets refreshed.
8. Create a second event dated in the past, assign the same venue, load its page — `geo_*` stays empty, confirming past events are never retroactively updated.

My WordPress.org username: jmarx75

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Read-only venue and event geo data (`geo_latitude`, `geo_longitude`, `geo_address`, `geo_public`) for Geodata-aware plugins like Simple Location, kept current without eager updates when a venue's address changes.

</details>

---

Authored by jmarx, assisted by Claude Code.

## Summary by CodeRabbit

* **New Features**
  * Added automatic synchronization of venue and event location metadata.
  * Exposed standardized geolocation fields through REST responses for supported venues and events.
  * Kept location metadata read-only while ensuring it reflects the assigned venue.
  * Updated public visibility metadata based on publication status.

* **Tests**
  * Added comprehensive coverage for synchronization, REST responses, supported post types, drafts, past events, and venue changes.


jmarx:
@coderabbitai review

includes/core/classes/event/class-geo-sync.php
```diff
@@ -0,0 +1,173 @@
...
+			'geo_latitude'  => $information['latitude'],
+			'geo_longitude' => $information['longitude'],
+			'geo_address'   => $information['address'],
+			'geo_public'    => 'publish' === get_post_status( $event_id ) ? 1 : 0,
+		);
+
+		$changed = false;
+
+		foreach ( $desired as $key => $value ) {
+			$current = get_post_meta( $event_id, $key, true );
```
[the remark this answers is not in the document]
jmarx: Confirmed and fixed in be6b0ece. `maybe_refresh()` now clears `geo_*` and forces `geo_public` to `0` when the linked venue isn't published; an event with no venue linked keeps its prior behavior. Added `test_maybe_refresh_hides_location_for_unpublished_venue` to cover it.

test/unit/php/includes/tests/core/classes/venue/class-test-geo-sync.php
```diff
@@ -0,0 +1,142 @@
...
+				'post_status' => 'publish',
+			)
+		)->get();
+
+		$revision_id = wp_save_post_revision( $venue->ID );
+		$this->assertIsInt( $revision_id, 'A revision should have been created.' );
+
+		$revision = get_post( $revision_id );
+
+		Geo_Sync::get_instance()->on_post_saved( $revision_id, $revision );
```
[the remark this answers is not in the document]
jmarx: Fixed in be6b0ece — split into test_on_post_saved_skips_revisions (unchanged) and a new test_on_post_saved_skips_autosaves, which builds an actual autosave-shaped post (post_type=revision, post_name={parent}-autosave-v1) and asserts wp_is_post_autosave() recognizes it before checking on_post_saved().

includes/core/classes/venue/class-meta.php
```diff
@@ -294,6 +296,43 @@ protected function register_venue_information_meta( string $post_type ): void {
...
+	/**
+	 * Registers the Geodata standard meta (`geo_*`) as read-only.
+	 *
+	 * @since 0.36.0
+	 *
+	 * @param string $post_type The post type to register against.
+	 *
+	 * @return void
+	 */
+	protected function register_geo_meta( string $post_type ): void {
```
mauteri, line 308: Method seems pretty much the same as `register_venue_geo_meta`. Let's refactor and abstract logic. Maybe put this in a generalized `class-geo-sync.php` and `class-meta.php` rather than repeating the same meta/logic in event and venue.

includes/core/classes/event/class-geo-sync.php
```diff
@@ -0,0 +1,179 @@
+<?php
+/**
+ * Keeps an event's Geodata standard meta (geo_latitude/longitude/address/
+ * public) in sync with its venue, checked lazily on view rather than
+ * pushed on every venue save — a venue can have ~1,000 linked events.
+ *
+ * @package GatherPress\Core\Event
+ * @since 0.36.0
```
mauteri, line 8: Sync `develop` and change all to `@since TBD`. New workflow that will change TBD to version on release.

mauteri (changes requested):

includes/core/classes/venue/class-meta.php
mauteri: Actually, keep the `class-meta.php` in each but generalize `class-geo-sync.php` so both can leverage it. Keep logic in geo sync, but hookups in meta. If something is to repeat, add it as a constant or method in geo sync. I think that's the shape for this.

PR #2328, opened 2026-09-20
Resolve event overrides by post type support

### LINKED ISSUE #2289: Event helpers hardcode gatherpress_event instead of resolving by post type support

mauteri:
Fixes #2289

## Proposed changes:

Two helpers in `src/helpers/event.js` fetched a post by naming `gatherpress_event` instead of resolving the type through `gatherpress-event-date` support.

* `getEventMeta()` asked for the override by name.
* `hasOnlineEventTerm()` assumed the ID belonged to whatever post type was open in the editor, then fell back to the same hardcoded slug.

An override pointing at a companion plugin's event type, say a `production` declaring `gatherpress-event-date`, sent the request to the events endpoint with a foreign ID. It 404s, the record comes back undefined, and the block reports defaults with nothing surfacing the failure: no guest limit, RSVP enabled, anonymous RSVP off, no online-event term.

Both go through `findEventPostById()` now, the support-based resolver already in that file, which the event date block already uses for this exact case.

`hasOnlineEventTerm()` needed more than the swap. It derived the venue taxonomy from the editor's post type, which is only right when the override happens to be the same type, so it read the wrong taxonomy as well as the wrong endpoint. The post is resolved first and the taxonomy now comes from what came back.

### One deliberate behavior change

`findEventPostById()` returns published posts only, where `getEntityRecord()` returned anything the user could edit. An override pointing at a **draft** now reads as defaults rather than reading the draft's values. That is the right answer for a block rendering on the front end, it was called out in the issue, and it is documented in both docblocks and covered by tests.

### Other information:

The three override tests mocked `getEntityRecord` directly, so they move onto `getPostTypes` and `getEntityRecords`. Worth knowing: **two of them had started passing for the wrong reason**, returning false because the post could not be resolved rather than because the term did not match.

New cases cover an override on a companion post type, an override pointing at a draft, and a resolved post carrying no venue terms. `src/helpers/event.js` is at 100% statements, branches, functions and lines.

## Testing instructions:

1. Register a post type with `gatherpress-event-date` support, or use one from a companion plugin.
2. Create a published event of that type with a guest limit and RSVP enabled.
3. On another post, add an RSVP block and set its post ID override to that event.
4. The block shows that event's settings rather than defaults, and there is no 404 against `/wp/v2/gatherpress_events/{id}` in the network tab.
5. Repeat with an online-event on a companion type and confirm the online-event term is detected.
6. Point an override at a draft event and confirm the block falls back to defaults.


mauteri:
Closing as a duplicate. #2290 already covers this, opened Sept 12, and reaches the same conclusion including deriving the venue taxonomy from the resolved post rather than the editor's post type. It also fixes the same class of bug in `class-import.php` and `event/class-rest-api.php`, which this branch did not touch.

My mistake for not checking for an existing PR on #2289 before starting. Bringing #2290 up to date with develop instead.

PR #2347, opened 2026-09-23
fix: stop telling CodeRabbit to reject @since TBD

mauteri:
## The problem

CodeRabbit has flagged `@since TBD` as an error on two PRs now, #2340 and #2346, both times citing *"Source: Path instructions"*. It was not misreading `AGENTS.md`. It was following `.coderabbit.yaml`, which told it the opposite:

> New functions, classes, and hooks need complete docblocks with an `@since` tag that **resolves to a released tag or the next planned release, never 1.0.0 or 'unreleased'**.

`AGENTS.md` line 218 says the opposite: a new symbol gets `@since TBD`, because the release it ships in is not knowable while the work is in flight, and `npm run version:bump` resolves every `TBD` at release time on the branch doing the releasing.

Taking CodeRabbit's advice would have broken CI, since the `@since tags` workflow fails a PR into `develop` that adds a literal version.

The config header says `AGENTS.md` is picked up automatically and these instructions "only add per-path emphasis". This one was not emphasis, it was a contradiction.

## The change

The `includes/**` instruction now states the actual convention and tells CodeRabbit not to ask for a version, so the config agrees with `AGENTS.md` and with the workflow.

CodeRabbit has recorded a learning from each of those threads, but learnings were working against a path instruction that kept asserting the opposite, so this is the deterministic fix.

## Notes

- `includes/**` was the only instruction mentioning `@since`; nothing else in the file conflicts.
- Config only, so `Skip Changelog`.


mauteri:
Folding this into #2346 instead, where the friction came up. Same commit, no separate PR to track.

